PR #137, opened 2015-01-22
Modifying the docs for Stormpath Servlet Plugin to explicitly state it is the Sto...

omgitstom:
...rmpath Java Servlet Plugin

### DIFF
--- a/docs/source/about.rst
+++ b/docs/source/about.rst
@@ -3,17 +3,17 @@
 About
 =====
 
-Not sure if the Stormpath Servlet Plugin is for you?  This page will help you decide if is a good fit!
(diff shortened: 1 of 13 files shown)


docs/source/about.rst
```diff
@@ -3,17 +3,17 @@
 About
 =====
 
-Not sure if the Stormpath Servlet Plugin is for you?  This page will help you decide if is a good fit!
+Building a web app, but not sure if the Stormpath Java Servlet Plugin is right for you?  This page will help you decide if is a good fit!
```
lhazlewood: Let's make this a valid sentence please :)  

"Are you building a web app, but not sure if the Stormpath Java Servlet Plugin is right for you?"

PR #142, opened 2015-02-10
Added PasswordPolicy Resource for directories

antollinim:


impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultPasswordPolicy.java
```diff
@@ -0,0 +1,119 @@
...
+        String value = getStringProperty(RESET_EMAIL_STATUS.getName());
+        if (value == null) {
+            return null;
+        }
+        return EmailStatus.valueOf(value.toUpperCase());
+    }
+
+    @Override
+    public PasswordPolicy setResetEmailStatus(EmailStatus status) {
+        setProperty(RESET_EMAIL_STATUS, status.name());
```
maranai: Seems like this method is missing an assert for the case when the "status" object is null

impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultPasswordPolicy.java
```diff
@@ -0,0 +1,119 @@
...
+        String value = getStringProperty(RESET_SUCCESS_EMAIL_STATUS.getName());
+        if (value == null) {
+            return null;
+        }
+        return EmailStatus.valueOf(value.toUpperCase());
+    }
+
+    @Override
+    public PasswordPolicy setResetSuccessEmailStatus(EmailStatus status) {
+        setProperty(RESET_SUCCESS_EMAIL_STATUS, status.name());
```
maranai: Seems like this method is missing an assert for the case when the "status" object is null

impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultStrength.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public int getMinSymbol() {
+        return getInt(MIN_SYMBOL);
+    }
+
+    @Override
+    public Strength setMinSymbol(int minSymbol) {
+        isTrue(minSymbol >= 0, "minSymbol cannot be a negative number.");
```
maranai: Quick question, aren't the assertions in this file missing the preceding "Assert."?

antollinim: I have added the `Assert` class as a static import:
```
import static com.stormpath.sdk.lang.Assert.*;
```

impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultPasswordPolicy.java
antollinim: Done. Thx

impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultPasswordPolicy.java
antollinim: Done. Thx

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+     * Return the time-to-live for the reset password token.
+     *
+     * @return the time-to-live for the reset password token
+     */
+    int getResetTokenTtl();
+
+    /**
+     * Specifies the amount of time (in hours) that a reset token will remain valid. Once that time has passed, the token
+     * gets invalidated.
+     *
```
josebarrueta: It has upper limit, so it might be useful to state that in the javadoc.

josebarrueta: Must be a positive integer, less than 169 (0 < i < 169)

api/src/main/java/com/stormpath/sdk/directory/Strength.java
```diff
@@ -0,0 +1,128 @@
...
+import com.stormpath.sdk.resource.Saveable;
+
+/**
+ * {@link Strength} is the resource used to configure the password strength policy. This policy defines requirements
+ * like minimum uppercase chars or maximum password length (among others) and will enforced by Stormpath when setting
+ * or resetting accounts' passwords.
+ *
+ * @since 1.0.0
+ */
+public interface Strength extends Resource, Saveable {
```
josebarrueta: For what is worth and to align the expectations to what actually happens in the ServerSide can we do a better javadoc of each character definition:

```
char[] symbolCharsArray =
            { '\u0020', '\u0021', '\u0022', '\u0023', '\u0024', '\u0025', '\u0026', 0x0027, '\u0028', '\u0029',
                '\u002A', '\u002B', '\u002C', '\u002D', '\u002E', '\u002F', '\u003A', '\u003B', '\u003C', '\u003D',
                '\u003E', '\u0040', '\u005B', 0x005C, '\u005D', '\u005E', '\u005F', '\u007B', '\u007C', '\u007D',
                '\u007E', '\u00A1', '\u00A7', '\u00A9', '\u00AB', '\u00AC', '\u00AE', '\u00B1', '\u00B5', '\u00B6',
                '\u00B7', '\u00BB', '\u00BD', '\u00BF', '\u00D7', '\u00F7' };

        char[] diacriticCharsArray =
            { '\u00C0', '\u00C1', '\u00C2', '\u00C3', '\u00C4', '\u00C5', '\u00C6', '\u00C7', '\u00C8', '\u00C9',
                '\u00CA', '\u00CB', '\u00CC', '\u00CD', '\u00CE', '\u00CF', '\u00D0', '\u00D1', '\u00D2', '\u00D3',
                '\u00D4', '\u00D5', '\u00D6', '\u00D8', '\u00D9', '\u00DA', '\u00DB', '\u00DC', '\u00DD', '\u00DE',
                '\u00DF', '\u00E0', '\u00E1', '\u00E2', '\u00E3', '\u00E4', '\u00E5', '\u00E6', '\u00E7', '\u00E8',
                '\u00E9', '\u00EA', '\u00EB', '\u00EC', '\u00ED', '\u00EE', '\u00EF', '\u00F0', '\u00F1', '\u00F2',
                '\u00F3', '\u00F4', '\u00F5', '\u00F6', '\u00F8', '\u00F9', '\u00FA', '\u00FB', '\u00FC', '\u00FD',
                '\u00FE', '\u00FF' };
```

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ApplicationCustomDataIT.groovy
```diff
@@ -200,7 +199,11 @@ class ApplicationCustomDataIT extends AbstractCustomDataIT {
     }
 
     def cleanTenantCustomData() {
-        client.currentTenant.customData.delete()
```
josebarrueta: Doesn't seem related to this, but also delete should be idempotent, so shouldn't matter wether is was deleted before, should be safe to call it again.

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
```diff
@@ -0,0 +1,53 @@
...
+public interface ResetEmailTemplate extends EmailTemplate<ResetEmailTemplate> {
+
+    /**
+     * An {@link java.util.Collections#unmodifiableMap(java.util.Map) UnmodifiableMap} where Stormpath-reserved
+     * properties (like <code>linkBaseUrl</code>) are stored.
+     *
+     * @return the map where Stormpath-reserved properties (like linkBaseUrl) are stored.
+     * @see #setLinkBaseUrl(String)
+     */
+    Map<String, String> getDefaultModel();
```
josebarrueta: Wonder if that statement of this map being an unmodifiable map is correct? because the linkBaseUrl is part of the defaultModel and can be modified to what customers want.

Having a convienience method to set/get the linkBaseUrl might be good, but doesn't reflect the REST API, also when we add more properties to define and/or customer can define their own, we won't be able to keep adding the convenience methods every time.

I think (maybe @lhazlewood  can weigh in too) that the way it should work is that the DefaultModel is a extension of map that define convenience methods in its interface, but also let the users add their own (with the drawback that at this time other than linkBaseURL will be ignored). 

In a sense will be kind of customData for email templates
```
getDefaultModel().setLinkBaseUrl(String);
getDefaultModel().getLinkBaseUrl();

but also can be added as:

getDefaultModel.put("linkBaseUrl", String);
getDafaultModel.get('linkBaseUrl");
```

This can help users to use this same version DefaultModel later, when we support new properties (if that time comes).

Thoughts?

api/src/main/java/com/stormpath/sdk/directory/Strength.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return the maximum quantity of total characters allowed in a password by this policy.
+     */
+    int getMaxLength();
+
+    /**
+     * Specifies the maximum quantity of total characters allowed in a password by this policy.
+     *
+     * @return this instance for method chaining.
+     */
+    Strength setMaxLength(int maxLength);
```
josebarrueta: There is an upper limit for the strength:

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
antollinim: @josebarrueta. This is the rationale of why I thought it would be a good idea to allow the user to see the whole map but to restrict the addition of data via pre-defined methods (currently just one: `setLinkBaseUrl()`):

1. I wanted to prevent users to remove the `linkBaseUrl` key altogether.
2. The UI only allows users to modify the value of the `linkBaseUrl` but does not allow them to add any other value
3. I did not want them to use this map as a `CustomData` because if we actually decide to treat this as a `CustomData` this should be called `CustomData` just for the sake of homogeneity and re-use already available code for the sake of maintainability.
4. If we later decide to add some reserved world to this map (like `linkBaseUrl` currently is) it could collide with custom keys already being used by clients.

I am completely open to allowing the map to be writable, I just thought it would do more harm to leave it writable  than keeping it closed.

Maybe @lhazlewood can help us define what we should do based on his future plans for the `defaultModel` concept.

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
antollinim: Done. Thx

api/src/main/java/com/stormpath/sdk/directory/Strength.java
antollinim: Done. Thx

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
antollinim: @lhazlewood here is where I need some guidance from you.

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Specifies the amount of time (in hours) that a reset token will remain valid. Once that time has passed, the token
+     * gets invalidated.
+     * <p/>
+     * Defaults to 24 hrs. Must be a positive integer, less than 169 (i.e. 0 < resetTokenTtl < 169). 168 hours equals 7 days.
+     *
+     * @param resetTokenTtl how long (in hours) a reset token can remain active until it is used.
+     * @return this instance for method chaining.
+     */
+    PasswordPolicy setResetTokenTtl(int resetTokenTtl);
```
lhazlewood: I think this should be explicit in the method name, for example, `setResetTokenTtlHours`.  This is pretty common when a TimeUnit isn't an argument to the method so that the time unit is explicit.

api/src/main/java/com/stormpath/sdk/directory/Strength.java
lhazlewood: Agreed - this should be explicit.  I see that some characters are listed below - did @antollinim do this already?

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
lhazlewood: Glad to clarify:

1.  If the SDK user attempts to remove `linkBaseUrl`, we can throw an exception locally (and should).  But they should be able to change the value.
2.  The UI is not fully current with what the API can/should support.  The 'source of truth' for what we need to support in the SDKs are the REST API, not the UI.  Users should be able to customize the default model if we support that in the API.
3.  Jose was just stating that it should operate like CustomData.  We can't  `CustomData` however because, IIRC, it is not an independent resource with its own href - i.e. it cannot be a `Resource` implementation.   A new `com.stormpath.sdk.mail.Model` interface (that is a sub-interface of `Map<String,Object>`, but not of `Resource`) would allow convenience methods as well as expected Map/JSON functionality.

Basically, *if* it is writable via the REST API, it should be writable via the SDK.  If it is not writable by the REST API, it should not be writable by the SDK.  Our representation in the SDK should be an *exact* reflection of what is possible with the API.  Once that is the case, we can add convenience methods on top of that to make things easier to use, but the raw API functionality should always be reflected in the SDK.

impl/src/main/java/com/stormpath/sdk/impl/directory/DefaultStrength.java
lhazlewood: I think, in this particular case, not using `Assert` in the method itself is confusing:  `isTrue`, without a prefix, seems like it would evaluate to a boolean, and then some logic should happen after that case.

The reason why the Assert class is named as it is is for readability:

`Assert.isTrue(condition, messageIfNotTrue)` is more readable than something that appears to return a boolean (which it does not).

lhazlewood: So my request is to not use static imports for the Assert class.

api/src/main/java/com/stormpath/sdk/directory/Strength.java
antollinim: Yes, I have added every single supported character in their corresponding type.

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.Map;
+
+/**
+ * The ResetEmailTemplate is a {@link EmailTemplate} resource which also provides the ability to configure the url where
+ * the user will be redirected once he clicks the link received in the Reset Password Email.
+ *
+ * @since 1.0.0
+ */
+public interface ResetEmailTemplate extends EmailTemplate<ResetEmailTemplate> {
```
lhazlewood: There doesn't appear to be anything 'Reset' specific about this.  I wonder if this should be called `ModelEmailTemplate` ?

api/src/main/java/com/stormpath/sdk/directory/Strength.java
lhazlewood: Sorry I didn't catch this earlier - this should be named `PasswordStrength`

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplateList.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.stormpath.sdk.resource.CollectionResource;
+
+/**
+ *
+ * A {@link CollectionResource} containing {@link ResetEmailTemplate} instances.
+ *
+ *  @since 1.0.0
+*/
+public interface ResetEmailTemplateList extends CollectionResource<ResetEmailTemplate> {
```
lhazlewood: `ModeledEmailTemplateList`

api/src/main/java/com/stormpath/sdk/mail/ResetSuccessEmailTemplate.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.stormpath.sdk.mail;
+
+/**
+ *
+ * A special kind of {@link EmailTemplate} to customize properties for Reset Success Emails.
+ *
+ * @since 1.0.0
+ */
+public interface ResetSuccessEmailTemplate extends EmailTemplate<ResetSuccessEmailTemplate> {
```
lhazlewood: Any reason why this exists considering it does not have any methods?

impl/src/main/java/com/stormpath/sdk/impl/resource/AbstractResource.java
```diff
@@ -428,6 +428,26 @@ protected boolean getBooleanProperty(String key) {
...
+    /**
+     * @since 1.0.0
+     */
+    protected Map getMapProperty(String key) {
+        Object value = getProperty(key);
+        if (value != null) {
+            if (value instanceof Map) {
+                return (Map) value;
+            }
+        }
```
lhazlewood: I can't remember our usage paradigm here, but I believe if the value is not null, we should use an assertion that the value is indeed a Map.  Otherwise the caller of this method is probably trying to do something wrong (i.e. expecting a map when the property is not a map), which would be a bug and should be fixed.

api/src/main/java/com/stormpath/sdk/mail/ResetEmailTemplate.java
antollinim: Makes total sense. Thx

api/src/main/java/com/stormpath/sdk/mail/ResetSuccessEmailTemplate.java
antollinim: @lhazlewood The only reason I created this was to prevent the public API to be affected if for some reason in the future the Email Templates need to diverge. The way I modeled it the `EmailTemplate` is not exposed in the API, it only provides common methods. Only `ModeledEmailTemplate` and `ResetSuccessEmailTemplate` are exposed. No other reason than that. If you feel this is not necessary I am totally fine with having `EmailTemplate` and `ModelEmailTemplate` (extending `EmailTemplate`) exposed and deleting `ResetSuccessEmailTemplate`.

lhazlewood: Is `ResetSuccessEmailTemplate` a ModeledEmailTemplate`?  If so, shouldn't it be a sub-interface of the latter?

antollinim: Nop. The `ResetSuccessEmailTemplate` does not have a `defaultModel`:

```json
{
   "href":"https://api.stormpath.com/v1/passwordPolicies/4dRaROoWwRcZG7fsP1k4AN/resetSuccessEmailTemplates",
   "offset":0,
   "limit":25,
   "size":1,
   "items":[
      {
         "href":"https://api.stormpath.com/v1/emailTemplates/3HEMyedoGyLyCpedrjlrpM",
         "name":"Default Password Reset Success Email Template",
         "description":"This is the reset success email template that is associated with the directory",
         "fromName":"Test Name",
         "fromEmailAddress":"test@stormpath.com",
         "subject":"Your password has been changed",
         "textBody":"Your password has been successfully changed.\n\nIf your password was changed without your notice please contact test@stormpath.com\n\n--------------------\nFor general inquiries or to request support with your account, please email test@stormpath.com",
         "htmlBody":"<p>Your password has been successfully changed.</p><p>If your password was changed without your notice please contact test@stormpath.com</p><p>--------------------<br/>For general inquiries or to request support with your account, please email test@stormpath.com</p>",
         "mimeType":"text/plain"
      }
   ]
}
```
While the `ResetEmailTemplate` (or Java-SDK's `ModeledEmailTemplate`) has a `defaultModel`:
```json
{  
   "href":"https://api.stormpath.com/v1/passwordPolicies/4dRaROoWwRcZG7fsP1k4AN/resetEmailTemplates",
   "offset":0,
   "limit":25,
   "size":1,
   "items":[  
      {  
         "href":"https://api.stormpath.com/v1/emailTemplates/3HEMybJjMO2za0YmfTubDI",
         "name":"Default Password Reset Email Template",
         "description":"This is the password reset email template that is associated with the directory",
         "fromName":"Test Name",
         "fromEmailAddress":"test@stormpath.com",
         "subject":"Reset your Password",
         "textBody":"Forgot your password?\n\nWe've received a request to reset the password for this email address.\n\nTo reset your password please click on this link or cut and paste this URL into your browser (link expires in ${expirationWindow} hours):\n${url}\n\nThis link takes you to a secure page where you can change your password.\n\nIf you don't want to reset your password, please ignore this message. Your password will not be reset.\n---------------------\nFor general inquiries or to request support with your account, please email test@stormpath.com",
         "htmlBody":"<p>Forgot your password?</p><br/><br/><p>We've received a request to reset the password for this email address.</p><p>To reset your password please click on this link (link expires ${expirationWindow} hours):<br/><a href=\"${url}\">Link</a></p><p>This link takes you to a secure page where you can change your password. If you don't want to reset your password, please ignore this message. Your password will not be reset.</p><p>----------------------<br/>For general inquiries or to request support with your account, please email test@stormpath.com</p>",
         "mimeType":"text/plain",
         "defaultModel":{  
            "linkBaseUrl":"https://api.stormpath.com/passwordReset"
         }
      }
   ]
}
```

To summarize, in the Java SDK the hierarchy is: 
```java
public interface EmailTemplate<T extends EmailTemplate> extends Resource, Saveable {
```
```java
public interface ModeledEmailTemplate extends EmailTemplate<ModeledEmailTemplate> {
```
```java
public interface ResetSuccessEmailTemplate extends 
EmailTemplate<ResetSuccessEmailTemplate> {
```
Meaning the parent interface is `EmailTemplate`. And its children are `ResetSuccessEmailTemplate` and `ModeledEmailTemplate`. There is no relationship between `ResetSuccessEmailTemplate` and `ModeledEmailTemplate` other than being siblings. For simplicity, think of `ResetSuccessEmailTemplate` as a non-modeled email template.

If you think the non-modeled email will never evolve independently of the modeled email template we could just have:

```java
public interface EmailTemplate<T extends EmailTemplate> extends Resource, Saveable {
```
```java
public interface ModeledEmailTemplate extends EmailTemplate<ModeledEmailTemplate> {
```
This will cause the `EmailTemplate` to be exposed in the API.

lhazlewood: Just a note per our Google Hangout: `ResetSuccessEmailTemplate` should be renamed to `PasswordResetSuccessEmailTemplate`

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return this instance for method chaining.
+     */
+    PasswordPolicy setResetSuccessEmailStatus(EmailStatus status);
+
+    /**
+     * Return the {@link PasswordStrength} resource to configure the password strength policy.
+     *
+     * @return the {@link PasswordStrength} resource to configure the password strength policy.
+     */
+    PasswordStrength getStrength();
```
josebarrueta: Should be renamed to ```getPasswordStrength()```?

api/src/main/java/com/stormpath/sdk/mail/PasswordResetSuccessEmailTemplate.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.stormpath.sdk.mail;
+
+/**
+ *
+ * A special kind of {@link EmailTemplate} to customize properties for Password Reset Success Emails.
+ *
+ * @since 1.0.RC4
+ */
+public interface PasswordResetSuccessEmailTemplate extends EmailTemplate<PasswordResetSuccessEmailTemplate> {
```
josebarrueta: Why is it ```PasswordResetSuccessEmailTemplate``` the only template that is a subclass of ```EmailTemplate```, doesn't add anything new.

api/src/main/java/com/stormpath/sdk/mail/PasswordResetSuccessEmailTemplateList.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.stormpath.sdk.resource.CollectionResource;
+
+/**
+ *
+ * A {@link CollectionResource} containing {@link PasswordResetSuccessEmailTemplate} instances.
+ *
+ * @since 1.0.RC4
+ */
+public interface PasswordResetSuccessEmailTemplateList extends CollectionResource<PasswordResetSuccessEmailTemplate> {
```
josebarrueta: Same comment as above.

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
antollinim: Les wanted the Resource to be scoped to the Password in case a new kind of
strength resource is required in the future. That is why the class was
renamed here in the SDK to PasswordStrength.

On the other hand since the resource is already a password (PasswordPolicy)
and the Rest API endpoint is '/strength'. I thought it was more accurate to
just keep the method as getStrength().

So between passwordPolicy.getStrength() and
passwordPolicy.getPasswordStrength() I think there is no more clarity in
the latter than in the former. Besides the latter seems redundant while the
former mimics the actual REST API endpoint.

api/src/main/java/com/stormpath/sdk/mail/PasswordResetSuccessEmailTemplate.java
antollinim: Les, Tom and I had a meeting around Email Templates today. Email Template
is exposed in the SDK API. We decided to expose an specific
PasswordResetSuccessEmailTemplate kindoif resource in order to be able to
modify it in the future without impacting other templates. Think of this as
a concrete empty class created only to provide flexibility of change in the
future. Makes sense?

api/src/main/java/com/stormpath/sdk/directory/PasswordPolicy.java
josebarrueta: Yeah, it does make sense, and I don't have an strong opinion to be one or the other. I think my comment is more related to Java Bean conventions, but I don't see any problem to be just ```getStrength()```, I like the fact that is shorter.

api/src/main/java/com/stormpath/sdk/mail/PasswordResetSuccessEmailTemplate.java
josebarrueta: It also makes sense, I was wondering why only this template and not the others, and making sure this wasn't left by mistake.

antollinim: @lhazlewood hopefully this is the last comment about Email Templates. The way it is modeled now, in the `PasswordPolicy` the API looks like this:

```java
    ModeledEmailTemplateList getResetEmailTemplates();
    PasswordResetSuccessEmailTemplateList getResetSuccessEmailTemplates();
```
Don't you think it looks inhomogeneous?
Just double checking with you before I consider the feature ready.

lhazlewood: The only reason why we might want a `PasswordResetEmailTemplate` (and not just `ModeledEmailTemplate` as it is now) is if we think we'll add Password Reset-specific behavior to the email template itself in the future.  I can't foresee a need to to do this, so I'm ok with the way the API is now.

The *only* reason for the `PasswordResetSuccessEmailTemplate` interface (with no methods) is so that we can make its parent be `ModeledEmailTemplate` in the future if/when a `defaultModel` is supported for that type of template.  This allows us to inject the `ModeledEmailTemplate` interface into that type hierarchy and does not require any of the SDK users to change their code.

If the API supported a `defaultModel` for password reset success email templates, we'd make that a `ModeledEmailTemplate` as well and then everything would be congruent.  Since the API does not support that today, this is a future-proof workaround that is good for now.

antollinim: Perfectly understood. I was only bringing up the point of the Templates not looking "homogenous" in their names.

If you think it is OK to be approved. Could you please do it? Thanks!

lhazlewood: Approved :)  Would you like me to merge the PR as well?

antollinim: If you like :)

PR #143, opened 2015-02-12
Avoid retry when server error

antollinim:


extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -338,7 +336,8 @@ private boolean shouldRetry(HttpRequestBase method, Throwable t, int retries) {
 
         if (t instanceof NoHttpResponseException ||
                 t instanceof SocketException ||
-                t instanceof SocketTimeoutException) {
+                t instanceof SocketTimeoutException ||
+                t instanceof ConnectTimeoutException) {
```
antollinim: @lhazlewood We were not previously retrying on ConnectionTimeout. Was it unintentionally left out? I added that feature now.  Please let me know if I should keep the previous behavior (i.e. not retrying on ConnectionTimeout). Thanks!

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -246,10 +247,7 @@ public Response executeRequest(Request request) throws RestException {
                         throw new RestException("HTTP 429: Too Many Requests.  Exceeded request rate limit in the allotted amount of time.");
                     }
 
-                    if (!response.isServerError() || retryCount > this.numRetries) {
-                        return response;
-                    }
-                    //otherwise allow the loop to continue to execute a retry request
```
antollinim: Retries do not happen in the code block any longer. They take place on `Exceptions`, not on proper server error responses.

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
lhazlewood: We should retry on ConnectionTimeout.

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
lhazlewood: We *should* retry on server error - if it is a 503 for example, we should definitely retry.

lhazlewood:
Why was this issue created? What is wrong with retrying on server error?

If it is a 503 for example, there should absolutely be a retry because the server will likely be available soon enough.

antollinim:
The issue was created because a few days ago there was a bug in the Backend (500 error) and the backend was getting some spikes because of the error+continual retries.

I agree that not all the server errors are the same, I just did it to broad, my bad. I should only avoid retrying on 500 specifically. Do you agree?

lhazlewood:
I looked at the status codes we are likely to return and we should retry on 503 or 504 - all others should not be retried.

antollinim:
@lhazlewood Thanks for the clarification. Code has been updated accordingly.

lhazlewood:
Let's get the build to succeed and I'll merge.

PR #144, opened 2015-02-13
Sporadic custom data errors fix

antollinim:
Added s short delay to avoid sporadic CustomData inconsistencies between backend and sdk


antollinim:
Fix does not work. The Custom Data IT error keeps happening from time to time

PR #148, opened 2015-02-18
New EmailVerificationToken#getToken() method

antollinim:
Additioanlly, account removed from cache after verification

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/account/EmailVerificationToken.java
+++ b/api/src/main/java/com/stormpath/sdk/account/EmailVerificationToken.java
@@ -23,4 +23,18 @@ import com.stormpath.sdk.resource.Saveable;
  */
 public interface EmailVerificationToken extends Resource, Saveable {
 
+    /**
+     * Returns the actual token required to verify the account. For example:
+     * <pre>
+     *     EmailVerificationToken emailVerificationToken = account.getEmailVerificationToken();
+     *     account = client.verifyAccountEmail(emailVerificationToken.getToken());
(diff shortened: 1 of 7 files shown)


api/src/main/java/com/stormpath/sdk/account/EmailVerificationToken.java
```diff
@@ -23,4 +23,18 @@
...
+     *     EmailVerificationToken emailVerificationToken = account.getEmailVerificationToken();
+     *     account = client.verifyAccountEmail(emailVerificationToken.getToken());
+     *     //at this point the account has been successfully verified.
+     * </code>
+     * </pre>
+     *
+     * @return the actual token required to verify the account.
+     * @since 1.0.RC4
+     */
+    String getToken();
```
lhazlewood: Should be called `getTokenValue()` ?  i.e. token.getToken() seems redundant.  What is the name of the `json` field?

antollinim: @lhazlewood I like "getValue()" better. I'll change it now...

BTW, there is no JSON field. In the email we get:
`https://api.stormpath.com/emailVerificationTokens?sptoken=5kVV3GEwDw9VyFSoM1gfMQ` 
while the resource href is
`https://api.stormpath.com/v1/accounts/emailVerificationTokens/5kVV3GEwDw9VyFSoM1gfMQ`

lhazlewood: Nice, I like `getValue()` too.

PR #149, opened 2015-02-18
Prevent PasswordResetEmail to be cached

antollinim:


jarias:
@antollinim Looks good, might need to check the issues in CI, but they don't seem related to this.

PR #155, opened 2015-03-02
Added the ability to retrieve all the applications for an account

antollinim:


api/src/main/java/com/stormpath/sdk/account/Account.java
```diff
@@ -447,4 +448,52 @@
      * @since 1.0.RC
      */
     ApiKey createApiKey(ApiKeyOptions options);
+
+    /**
+     * Returns a paginated list of all of the account's {@link com.stormpath.sdk.application.Application Application} resources.
```
lhazlewood: Let's have the JavaDoc reflect how an account is mapped to an application since it may not be obvious by looking at only the method signature:

    Returns a paginated list of the applications that the account may login to.  You may control which accounts may log in to specific applications by managing an application's mapped {@link AccountStore}s.
    
    @see Application#getAccountStoreMappings()

api/src/main/java/com/stormpath/sdk/account/Account.java
```diff
@@ -447,4 +448,52 @@
...
+     * the resource URL as query parameters, for example:
+     * <pre>
+     * .../accounts/accountId/applications?param1=value1&param2=value2&...
+     * </pre>
+     *
+     * @param queryParams the query parameters to use when performing a request to the collection.
+     * @return a paginated list of the Account's applications that match the specified query criteria.
+     * @since 1.0.RC4
+     */
+    ApplicationList getApplications(Map<String, Object> queryParams);
```
lhazlewood: Similar wording here, just indicating the ability to specify query params.

api/src/main/java/com/stormpath/sdk/account/Account.java
```diff
@@ -447,4 +448,52 @@
...
+     */
+    ApplicationList getApplications();
+
+    /**
+     * Returns a paginated list of the account's applications that match the specified query criteria.
+     * <p/>
+     * Each {@code queryParams} key/value pair will be converted to String name to String value pairs and appended to
+     * the resource URL as query parameters, for example:
+     * <pre>
+     * .../accounts/accountId/applications?param1=value1&param2=value2&...
```
lhazlewood: Need to use `&amp;` for &

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/ImmutableAccount.java
```diff
@@ -222,6 +224,24 @@ public ApiKey createApiKey(ApiKeyOptions options) {
         return null;
     }
 
+    /** @version 1.0.RC4 */
```
lhazlewood: This is a good example of why we may want the sub projects to be modules in the same Maven project space: these kinds of modifications are caught immediately, rather than in the build of the other project later on.

api/src/main/java/com/stormpath/sdk/account/Account.java
```diff
@@ -447,4 +448,52 @@
...
+     *     .withAccounts(10, 10)
+     *     .offsetBy(20)
+     *     .limitTo(25));
+     * </pre>
+     *
+     * @param criteria the  the query parameters to use when performing a request to the collection.
+     * @return a paginated list of the Account's applications that match the specified query criteria.
+     * @since 1.0.RC4
+     */
+    ApplicationList getApplications(ApplicationCriteria criteria);
```
lhazlewood: Same - wording, etc.

lhazlewood:
No Integration Tests for this?

antollinim:
I did not consider necessary to add ITs since all the new code is been tested with Unit Tests and it mainly relies in Collections that are already been ITed in different resources. However, given the fact that we sometime use the SDK as a tool to thoroughly test the backend I just added some ITs

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/account/AccountIT.groovy
```diff
@@ -755,6 +758,101 @@ class AccountIT extends ClientIT {
...
+
+        app1.delete() //now, account should belong to a single app
+
+        count = 0
+        for(Application application : account.getApplications()) {
+            count++
+        }
+        assertEquals(count, 1)
+
+//        def dir = client.instantiate(Directory)
```
maranai: I think these commented lines can just be removed.

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ApplicationCustomDataIT.groovy
```diff
@@ -116,8 +116,6 @@ class ApplicationCustomDataIT extends AbstractCustomDataIT {
         updateTenant(tenant, postedCustomData, [:])
         updateTenant(tenant, postedCustomData, createDataForUpdate())
 
-        cleanTenantCustomData()
```
maranai: Were these lines removed to prevent the tests that used to fail from failing? I'm just curious about the solution and why this tests fail randomly.

antollinim:
@lhazlewood @maranai Is this approved?

lhazlewood:
@antollinim let's get the build passing first please.

Issue #35, opened 2014-03-03
Support options on single object retrieval.


hereen:
The documentation(1) currently describes a way to retrieve customData using Link expansion:
```
AccountList accounts = application.getAccounts(
            Accounts.where(Accounts.email().eqIgnoreCase("some@email.com"))
                    .withCustomData()
); 
```
Which requires the user to handle a list even when using a unique value (like email) to return only 1 object. 

A circuitous solution would be to modify the url used in retrieval:
```
Account account = client.getResource("https://api.stormpath.com/v1/accounts/{acctId}*?expand=customData*", Account.class);
```
But this is not very "SDK-like".

Please provide official SDK retrieval support for a single object with expansion options.

--------- End Notes ---------------
(1) http://docs.stormpath.com/java/product-guide/#retrieve-custom-data

mrioan:
Released

PR #158, opened 2015-03-17
Add a way to retrieve a single Resource with Expansion

maranai:


api/src/main/java/com/stormpath/sdk/ds/DataStore.java
```diff
@@ -54,6 +55,22 @@
...
+     * of {@link Resource}, for example {@link Account Account}, {@link Directory Directory}, etc.
+     *
+     * @param href  the URL of the resource to retrieve
+     * @param clazz the {@link Resource} sub-interface to instantiate
+     * @param <T>   type parameter indicating the returned value is a {@link Resource} instance.
+     * @param options the {@link Options} sub-interface with the properties to expand
+     * @return an instance of the specified class based on the data returned from the specified {@code href} URL.
+     * @since 1.0.RC4
+     */
+    <T extends Resource> T getResourceExpanded(String href, Class<T> clazz, Options options);
```
antollinim: Please move this new method declaration to the `InternalDataSource` class

antollinim: Please move this new method declaration to the `InternalDataStore` class

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+    /**
+     * Ensures that when retrieving a Tenant, the associated {@link Tenant#getGroups() groups} are also retrieved in the same request.
+     *
+     * @param limit the number of results in the groups collection's first page.  Min: 1, Max: 100.
+     * @param offset the starting index of the first {@link Group} to retrieve in the overall groups collection's result set.
+     * @return this instance for method chaining.
+     * @since 1.0.RC4
+     */
+    T withGroups(int limit, int offset);
+
```
antollinim: Nice catch, these options were missing before!

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+     *
+     * @return this instance for method chaining.
+     * @since 1.0.RC4
+     */
+    T withAccounts();
+
+    /**
+     * Ensures that when retrieving a Tenant, the associated {@link Tenant#getAccounts() accounts} are also retrieved in the same request.
+     *
+     * @param limit the number of results in the accounts collection's first page.  Min: 1, Max: 100.
```
antollinim: @param limit defines the maximum quantity of items to be retrieved when expanding this Collection Resource. Min:1, Max: 100. The default value is 25.

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+     * @return this instance for method chaining.
+     * @since 1.0.RC4
+     */
+    T withAccounts(int limit);
+
+    /**
+     * Ensures that when retrieving a Tenant, the associated {@link Tenant#getAccounts() accounts} are also retrieved in the same request.
+     *
+     * @param limit the number of results in the accounts collection's first page.  Min: 1, Max: 100.
```
antollinim: Idem

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+     * @param offset the starting index of the first Account to retrieve in the overall accounts collection's result set.
```
antollinim: @param offset the zero-based starting index in the entire collection of the first item to return. Default is 0

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+     *
+     * @return this instance for method chaining.
+     * @since 1.0.RC4
+     */
+    T withGroups();
+
+    /**
+     * Ensures that when retrieving a Tenant, the associated {@link Tenant#getGroups() groups} are also retrieved in the same request.
+     *
+     * @param limit the number of results in the groups collection's first page.  Min: 1, Max: 100.
```
antollinim: Idem

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
```diff
@@ -43,4 +43,67 @@
...
+     * @return this instance for method chaining.
+     * @since 1.0.RC4
+     */
+    T withGroups(int limit);
+
+    /**
+     * Ensures that when retrieving a Tenant, the associated {@link Tenant#getGroups() groups} are also retrieved in the same request.
+     *
+     * @param limit the number of results in the groups collection's first page.  Min: 1, Max: 100.
+     * @param offset the starting index of the first {@link Group} to retrieve in the overall groups collection's result set.
```
antollinim: Idem for both limit and offset

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/DirectoryIT.groovy
```diff
@@ -215,6 +216,38 @@ class DirectoryIT extends ClientIT {
...
+
+        String href = dir.href
+
+        DirectoryOptions options = Directories.options().withAccounts()
 
+        // test options created successfully
+        assertNotNull options
+        assertEquals options.expansions.size(), 1
+
+        Directory retrieved = client.getResourceExpanded(href, Directory.class, options).iterator().next()
```
antollinim: Why does this line need `iterator().next()` in order to retrieve the expanded resource? Isn't the signature of the method `<T extends Resource> T getResourceExpanded(String href, Class<T> clazz, Options options);` ? This means the returned object should be a single `DIrectory` Resource, then why is `iterator().next()` needed here?

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/DirectoryIT.groovy
```diff
@@ -215,6 +216,38 @@ class DirectoryIT extends ClientIT {
...
+        Account account = client.instantiate(Account)
+        account = account.setGivenName('John')
+                .setSurname('Doe')
+                .setEmail('johndoe@email.com')
+                .setPassword('Changeme1!')
+        dir.createAccount(account)
+
+        retrieved = client.getResourceExpanded(href, Directory.class, options).iterator().next()
+        def retrievedAccount = retrieved.accounts.iterator().next()
+        assertNotNull retrievedAccount.username
```
antollinim: Am I wrong or this will always assert not null with or without expansion. If the account is no already espanded the username will be retrieved from the backend, right? The proper way to assert this is to get the internal property map and check its current state. I have done that in the past. Please take a look at `AccountIT#testInternalProperties`, here: `Map properties01 = getValue(AbstractResource, account01, "properties")`

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/tenant/TenantIT.groovy
```diff
@@ -384,4 +386,51 @@ class TenantIT extends ClientIT {
...
+        group2.name = uniquify("Java SDK: TenantIT.testTenantExpansion_group2")
+        group2 = app.createGroup(group2)
+        deleteOnTeardown(group2)
+
+        retrieved = client.getResourceExpanded(tenant.href, Tenant.class, options).iterator().next()
+        assertEquals(tenant.href, retrieved.href)
+
+        def finalGroups = 0
+        iterator = retrieved.groups.iterator()
+        while(iterator.hasNext()){
```
antollinim: As previously explained, Expansion is not being properly tested. This will go to the backend if the expansion is not working, so this test will always succeed.

impl/src/main/java/com/stormpath/sdk/impl/client/DefaultClient.java
```diff
@@ -391,4 +409,11 @@ public GroupList getGroups(Map<String, Object> queryParams) {
         return getCurrentTenant().getGroups(queryParams);
     }
 
+//    /**
+//     * @since 1.0.RC4
+//     */
+//    @Override
+//    public <T extends Resource> T getResourceWithExpansionOptions(String href, Class<T> clazz, Class<Options> options) {
+//        return this.dataStore.getResourceWithExpansionOptions(href, clazz, options);
+//    }
```
antollinim: Please remove this

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -192,11 +174,24 @@ public CacheManager getCacheManager() {
 
     @Override
     public <T extends Resource> T getResource(String href, Class<T> clazz, Map<String, Object> queryParameters) {
+        Assert.hasText(href, "href argument cannot be null or empty.");
+        Assert.notNull(clazz, "Resource class argument cannot be null.");
```
antollinim: Assert.notNull(clazz, "clazz argument cannot be null.");

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -192,11 +174,24 @@ public CacheManager getCacheManager() {
...
         SanitizedQuery sanitized = QuerySanitizer.sanitize(href, queryParameters);
         return getResource(sanitized.getHrefWithoutQuery(), clazz, sanitized.getQuery());
     }
 
     @Override
+    public <T extends Resource> T getResourceExpanded(String href, Class<T> clazz, Options options) {
+        Assert.hasText(href, "href argument cannot be null or empty.");
+        Assert.notNull(clazz, "Resource class argument cannot be null.");
```
antollinim: Idem

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -289,26 +284,26 @@ public CacheManager getCacheManager() {
...
-                            .setHref(href)
-                            .setOffset(filteredQs.containsKey(OFFSET.getName()) ? Integer.valueOf(filteredQs.get(OFFSET.getName())) : 0)
-                            .setLimit(filteredQs.containsKey(LIMIT.getName()) ? Integer.valueOf(filteredQs.get(LIMIT.getName())) : 25)
-                            .setItemsMap(apiKeyData);
-                    data = builder.build();
-                }
-            } else {
-                data = getCachedValue(href, clazz);
-            }
-        }
```
antollinim: Why is this removed? This change is very likely to cause some error because it seems to be a necessary hack for ApiKey data.

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -316,7 +311,7 @@ public CacheManager getCacheManager() {
             Request request = createRequest(HttpMethod.GET, href, filteredQs);
             data = executeRequest(request);
 
-            if (!Collections.isEmpty(data) && isCacheUpdateEnabled(clazz)) {
+            if (!Collections.isEmpty(data) && isCacheUpdateEnabled(clazz) && !qs.containsKey("expand")) {
```
antollinim: Just curious about the addition of `&& !qs.containsKey("expand")`. Does this prevent the error that we have discussed to take place? Does this also mean that resources with expansion are not cached?

impl/src/main/java/com/stormpath/sdk/impl/ds/InternalDataStore.java
```diff
@@ -54,8 +54,6 @@
 
     <T extends Resource> T getResource(String href, Class<T> clazz, Criteria criteria);
 
-    //<T extends Resource> T getResource(String href, Class<T> clazz, Options options);
-
```
antollinim: As mentioned before, the declaration of this method should be here and not in `DataStore.java`

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
maranai: If I moved this method declaration to the InternalDataStore, how could I call it using client.getResourceExpanded()? This because the DefaultClient has a DataStore instance, not an InternalDataStore one.
I like getResource better but we have a conflict with Options and Criteria because Criteria extends Options and then the sign is invalid (duplicate) as we already have a getResource(href param, class param, Criteria param) method .

api/src/main/java/com/stormpath/sdk/tenant/TenantOptions.java
maranai: Much better comment. Change applied.

impl/src/main/java/com/stormpath/sdk/impl/client/DefaultClient.java
maranai: Removed

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
maranai: Changed

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
maranai: Restored. Not intended to remove this.

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
maranai: Yes, this is a temporarily fix to prevent the expanded resources from being cached and later being wrongly returned. This should be removed when a cache fix is implemented.

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
antollinim: 1. WRT `InternalDataStore`, you are right. I did not realize that operation would not be exposed by the `Client` instance. You can leave it in `DataStore`
2. The `Criteria` and `Options` are not related to each other at the top of the hierarchy. Unless I am overlooking something, having `<T extends Resource> T getResource(String href, Class<T> clazz, Options options);` should not collide with `<T extends Resource> T getResource(String href, Class<T> clazz, Criteria criteria);`. Please clearly clarify in its JavaDoc that the new method is intended for retrieving a single resource with expansion for internal properties and provide a few examples of how this method is supposed to be used.

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/DirectoryIT.groovy
antollinim: You are obtaining the `properties01` map but you are not asserting its internal state. You are still manipulating the `retrieved` Directory resource: `def retrievedAccount = retrieved.accounts.iterator().next()`. If you do this the internal state will be modified by internally doing a backend call and thus you are not properly testing that this expansion really worked with the single execution of `client.getResourceExpanded`. You need to assert the status of `properties01` checking that the expanded account is indeed there without doing `retrieved.accounts.iterator()....`

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/tenant/TenantIT.groovy
antollinim: Idem about not invoking `iterator = retrieved.groups.iterator()`. You need to validate the internal state of the `retrieved` `Tenant` after the single execution of `client.getResourceExpanded(...)`

impl/src/main/java/com/stormpath/sdk/impl/client/DefaultClient.java
```diff
@@ -222,6 +223,23 @@ private DataStore createDataStore(Object requestExecutor, Object secondCtorArg,
...
+     * @param options the {@link Options} sub-interface with the properties to expand
+     * @param <T>   type parameter indicating the returned value is a {@link Resource} instance.
+     * @return an instance of the specified {@code Class} based on the data returned from the specified {@code href} URL.
+     *
+     * @since 1.0.RC4
+     */
+    @Override
+    public <T extends Resource> T getResourceExpanded(String href, Class<T> clazz, Options options) {
+        return this.dataStore.getResourceExpanded(href, clazz, options);
+    }
```
antollinim: Please rename to `getResource(...)`

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
antollinim: Please add an inline comment explaining the reason why this `!qs.containsKey("expand")` was added. Also add a Github Issue describing the problem (will be nice to have a test that demonstrate the issue as well) and add the link to the issue here so we can get back to this piece of code when we work on the Github issue.

impl/src/main/java/com/stormpath/sdk/impl/ds/InternalDataStore.java
antollinim: Please remove this line

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
lhazlewood: Please rename this method to be named `getResource` only (not 'Expanded' suffix).  The fact that it is overloaded with an Options argument implies expansion, but this may not always be the case - there might be an option in the future that is something other than just expansion, at which point the method name wouldn't make sense anymore.

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
```diff
@@ -54,6 +55,22 @@
     <T extends Resource> T getResource(String href, Class<T> clazz);
 
     /**
+     * Looks up (retrieves) the resource at the specified {@code href} URL with the properties specified in {@code options} expanded
```
lhazlewood: Let's write this as:

Retrieves the resource at the specified {@code href} according to the specified {@code Options} and returns the resource as an instance of the specified {@code clazz}.

Also note that immediately closed paragraph markers (<p/>) are not valid JavaDoc syntax for the Java 8 JavaDoc parser.  Paragraphs after the opening paragraph should use the full notation, i.e. <p>text here</p>.

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
lhazlewood: Also, since Options is not a valid argument (the caller *must* specify a particular sub-interface of Options), the signature should be:

    <T extends Resource, O extends Options> T getResource(String href, O options);

lhazlewood: I'm also not aware of any conflict with getResource having an `Options` vs a `Criteria` argument - that's just basic method overloading:

```java
<T extends Resource, O extends Options> T getResource(String href, Class<T> clazz, O options);

<T extends Resource, C extends Criteria> T getResource(String href, Class<T> clazz, C criteria);
```

Where is the conflict?

impl/src/main/java/com/stormpath/sdk/impl/client/DefaultClient.java
lhazlewood: :+1:

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
lhazlewood: :+1: on the //TODO: comment + GitHub issue.  We'll totally forget to to do this without the issue.

api/src/main/java/com/stormpath/sdk/ds/DataStore.java
maranai: @lhazlewood the problem I found is that the implementation of Criteria interface is:
public class DefaultCriteria<T extends Criteria<T>, O extends Options> implements Criteria<T>, Expandable 
so any class extending it will extend Criteria and Options and a call to getResource results in a "reference to getResource is ambiguous" error

maranai: @lhazlewood this is the problem I found after renaming the getResourceExpanded method to getResource. Do you have a recommendation to avoid it? Tnks

api/src/main/java/com/stormpath/sdk/directory/Directory.java
```diff
@@ -365,7 +365,7 @@
      * It allows to configure the emails that will be delivered for accounts created under the parent {@link Directory}
      *
      * @return the {@link AccountCreationPolicy} for this Directory.
-     * @since 1.0-SNAPSHOT
+     * @since 1.0.RC4.5
```
maranai: Version needs to be 1.0.RC4.6

mrioan:
@maranai I approve this PR. You are free to merge it @maranai . Thanks!

maranai:
PR has been merged.

mrioan:
resolves #35

PR #162, opened 2015-03-24
Add code to expose and query by creation and modification timestamps (JSDK-181)

maranai:
(JSDK-181)


api/src/main/java/com/stormpath/sdk/account/Accounts.java
```diff
@@ -30,7 +31,7 @@
  * <b>Accounts.where(Accounts.surname()</b>.containsIgnoreCase("Smith")<b>)</b>
  *     .and(<b>Accounts.givenName()</b>.eqIgnoreCase("John"))
  *     .orderBySurname().descending()
- *     .withGroups(10, 10)
+ *     .withAccounts(10, 10)
```
lhazlewood: I don't think this is accurate - an account does not have have a paged collection of accounts.

api/src/main/java/com/stormpath/sdk/account/Accounts.java
```diff
@@ -43,7 +44,7 @@
  * <b>where(surname()</b>.containsIgnoreCase("Smith")<b>)</b>
  *     .and(<b>givenName()</b>.eqIgnoreCase("John"))
  *     .orderBySurname().descending()
- *     .withGroups(10, 10)
+ *     .withAccounts(10, 10)
```
lhazlewood: Same.

api/src/main/java/com/stormpath/sdk/query/DateExpressionFactory.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package com.stormpath.sdk.query;
+
+/**
+ * An {@code DateExpressionFactory} creates a single condition (matches) for resource properties of Date type.
+ *
+ * @since 1.0.RC4
+ */
+public interface DateExpressionFactory {
```
lhazlewood: This should have various type-safe methods so the user is not forced to do string manipulation in code to construct String expressions.

For example:

    DateCriterion gt(Date date);
    DateCriterion gte(Date date)
    DateCriterion lt(Date date);
    DateCriterion lte(Date date)
    DateCriterion equals(Date date)
    DateCriterion in(Date begin, Date end);
    DateCriterion in(Date begin, Duration duration);

- Using both `equals` and any of the other methods should throw an exception: they are mutually exclusive.  
- `in` is always equivalent to `gte(begin).lt(end)` (that is, begin time is inclusive, the end time is exclusive).
- `in(begin, duration)` is a convenience: the `Duration` argument is used to calculate the end timestamp and then normal `within` semantics apply.

We'll have to move the Duration class from the `com.stormpath.sdk.impl.util` package in the `impl` module to `com.stormpath.sdk.lang` package in the `api` module (which is fine - we haven't had the need to expose it in the public API, but now seems like a good time to do that).

Finally, there should probably be a `com.stormpath.sdk.lang.Instants` utility class that allows people to construct UTC-based dates easily:

    of(year); // implementation calls `of(year, 0)`
    of(year, month); //implementation calls `of(year, month, 0)`
    of(year, month, day) //implementation calls of(year, month, day, 0)`
    // ... etc ...
    of(year, month, day, hour, minute, second, millisecond);

The final method would use the GregorianCalendar class to construct a Date that reflects that timestamp instant in the UTC timezone.

Then anyone could construct type-safe queries:

    Groups.where(Goups.modifiedAt().in(begin, new Duration(1, TimeUnit.HOURS)));

lhazlewood: In other words, ISO 8601 timestamps are necessary for the REST API because that is the general-purpose representation for strings.  But because Java is a type-safe language, we shouldn't force this lower-level concept on users when we can use nice things like dates, durations and intervals.

api/src/main/java/com/stormpath/sdk/account/Accounts.java
maranai: You're right. Just fixed it.

api/src/main/java/com/stormpath/sdk/account/Accounts.java
maranai: Idem

api/src/main/java/com/stormpath/sdk/query/DateExpressionFactory.java
maranai: Sounds good, working on this now.

lhazlewood: Thanks!  Looking forward to it.  Great job so far!

api/src/main/java/com/stormpath/sdk/account/Accounts.java
```diff
@@ -243,6 +244,58 @@ public static EqualsExpressionFactory status() {
...
+     * <pre>
+     * AccountCriteria criteria = Accounts.criteria();
+     * DateExpressionFactory createdAt = Accounts.createdAt();
+     * Criterion createdAtMatches = createdAt.matches("[,2014-04-05T12:00:00]");
+     * criteria.add(createdAtMatches);
+     * </pre>
+     *
+     * @return a new {@link Account#getCreatedAt() createdAt}-specific {@link DateExpressionFactory} instance, to be
+     *         used to construct a criterion when building an {@link AccountCriteria} query.
+     * @since 1.0-SNAPSHOT
```
mrioan: You will need to do a pull from master in order to get the correct pom versions. They have been modified to `1.0.RC4.3-SNAPSHOT`. So please replace ever `@since 1.0-SNAPSHOT` with `@since 1.0.RC4.3` everywhere you see it in these commits. Please note that the version in `@since` must not have the `-SNAPSHOT` suffix.

api/src/main/java/com/stormpath/sdk/lang/Instants.java
```diff
@@ -0,0 +1,242 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package com.stormpath.sdk.lang;
+
+import java.util.*;
+import java.util.Date;
+
+/**
+ * Utility class to create UTC-based dates and perform time conversions from UTC to other {@link TimeZone} timezones and vice versa
```
mrioan: ... to other {@link TimeZone timezones}

api/src/main/java/com/stormpath/sdk/resource/Auditable.java
```diff
@@ -0,0 +1,40 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package com.stormpath.sdk.resource;
+
+import java.util.Date;
+
+/**
+ * @since 1.0-SNAPSHOT
```
mrioan: Please provide a short description of the purpose of this interface. See `Extendable.java` as an example.

impl/src/main/java/com/stormpath/sdk/impl/provider/AbstractProvider.java
```diff
@@ -30,7 +31,7 @@
  *
  * @since 1.0.beta
  */
-public abstract class AbstractProvider extends AbstractResource implements Provider {
+public abstract class AbstractProvider extends AbstractResource implements Provider, Auditable {
```
mrioan: I think that `Provider` should be the one implementing `Auditable` rather than this abstract class

impl/src/main/java/com/stormpath/sdk/impl/provider/AbstractProviderData.java
```diff
@@ -30,7 +31,7 @@
  *
  * @since 1.0.beta
  */
-public abstract class AbstractProviderData extends AbstractResource implements ProviderData {
+public abstract class AbstractProviderData extends AbstractResource implements ProviderData, Auditable {
```
mrioan: I think that `ProviderData` should be the one implementing `Auditable` rather than this abstract class

impl/src/main/java/com/stormpath/sdk/impl/query/DefaultDateExpressionFactory.java
```diff
@@ -0,0 +1,137 @@
...
+        return new SimpleExpression(propertyName, value, Operator.EQUALS);
+    }
+
+    private Date calculateDateFromDuration(Date begin, Duration duration){
+        Calendar cal = Calendar.getInstance();
+        cal.setTimeInMillis(begin.getTime() + this.DurationToIntMillis(duration));
+        return cal.getTime();
+    }
+
+    private long DurationToIntMillis(Duration duration){
```
mrioan: Shouldn't it be `private long durationToMillis(Duration duration)` ?

impl/src/main/java/com/stormpath/sdk/impl/resource/AbstractExtendableInstanceResource.java
```diff
@@ -19,18 +19,23 @@
...
 import com.stormpath.sdk.resource.Saveable;
 
+import java.util.Date;
 import java.util.Map;
 
 /**
  * @since 1.0.0
  */
-public abstract class AbstractExtendableInstanceResource extends AbstractInstanceResource implements Saveable, Extendable {
+public abstract class AbstractExtendableInstanceResource extends AbstractInstanceResource implements Auditable, Saveable, Extendable {
```
mrioan: The fact that `AbstractExtendableInstanceResource` is implementing `Savable` is redundant since `AbstractInstanceResource` is already implementing it. Can you please remove it from here? I know that this is not part of your code in particular but since I am just noticing it here and this class is changing right now we can just take the opportunity to improve it here. Thanks!

api/src/main/java/com/stormpath/sdk/query/DateExpressionFactory.java
```diff
@@ -0,0 +1,114 @@
...
+ * An {@code DateExpressionFactory} creates a single condition (matches) for resource properties of Date type.
+ *
+ * @since 1.0-SNAPSHOT
+ */
+public interface DateExpressionFactory {
+
+    /**
+     * Convenience method that returns a new equals expression as a String expression, indicating the specified value should match the property value.
+     * Multiple values are allowed, for example:
+     * .createdAt().matches("[,2014-04-05T12:00:00]"): matches all entities created between the beginning of time and noon of 2014/04/05
```
mrioan: Could you please add the missing java object over which the `.createdAt()` method will be executed? Since these are examples they should be complete for the users to fully understand them.
Also, please surround each piece of example with <pre> </pre>

api/src/main/java/com/stormpath/sdk/query/DateExpressionFactory.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param value the value that should equal the property value (ignoring case).
+     * @return a new case-insensitive equals expression reflecting the property name and the specified value.
+     */
+    Criterion matches(String value);
+
+    /**
+     * Returns a new expression indicating the property value must be greater than the specified date.
+     * For example:
+     * .createdAt().gt(new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss").parse("2014-01-01T12:00:00")): matches only those entities created after 2014/01/01 at noon
```
mrioan: Idem here and subsequent methods

impl/src/test/java/com/stormpath/sdk/impl/resource/AuditableTestResource.java
```diff
@@ -0,0 +1,40 @@
...
+
+import com.stormpath.sdk.impl.ds.InternalDataStore;
+
+import java.util.Date;
+import java.util.Map;
+
+/**
+ * @since 1.0-SNAPSHOT
+ */
+public class AuditableTestResource extends AbstractExtendableInstanceResource {
```
mrioan: Please add a comment saying that this resource is just a test resource created for testing purposes

impl/src/main/java/com/stormpath/sdk/impl/resource/AbstractExtendableInstanceResource.java
maranai: Sure, good catch.

mrioan:
@maranai I am reviewing and updating this PR

mrioan:
@maranai this PR is approved. You can merge it. Thanks

maranai:
PR has been merged. Thanks @mrioan

mrioan:
Solves #199

PR #163, opened 2015-03-24
Collection Resources report total collection size

antollinim:


api/src/main/java/com/stormpath/sdk/resource/CollectionResource.java
```diff
@@ -47,4 +47,18 @@
      */
     public int getLimit();
 
+    /**
+     * Returns the total size (quantity of items) of this CollectionResource.
```
lhazlewood: I'd make the first line a little more explicit since it might be so obvious when thinking of pagination.  How about:

> Returns the total count (quantity) of all items across the entire collection.  Pagination does not impact this value.

Issue #127, opened 2015-01-14
Support getSingle*() methods


chrylis:
A number of methods (e.g., `getApplications()`) are frequently used to just select a single result (e.g., by name), but the methods all return `CollectionResource` objects. It would be helpful to have JPA-style `getSingle` methods (e.g., `client.getSingleApplication(criteria)`) that verified that exactly one item matches and return that single item.

Instead of:

    client.getApplications(Applications.where(Applications.name().eqIgnoreCase(stormpathInfo.getApplication()))).iterator().next();
    // plus sanity checks

just

    client.getSingleApplication(Applications.where(Applications.name().eqIgnoreCase(stormpathInfo.getApplication())));

lhazlewood:
I like it - thanks for the suggestion!

antollinim:
Hello @chrylis. We are going to add the `getSingle()` method to the `CollectionResource` interface. This will materialize this kind of operations:
```java
Application app = client.getApplications(Applications.where(Applications.name().eqIgnoreCase(stormpathInfo.getApplication()))).getSingle();
Account account  = application.getAccounts(Accounts.where(Accounts.username().eqIgnoreCase("test@stormpath.com"))).getSingle();
Group group = account.getGroups(Groups.criteria().withAccountMemberships(1, 0).and(Groups.name().endsWithIgnoreCase("X"))).getSingle();
Account account = directory.getAccounts(Accounts.surname().eqIgnoreCase("Doe")).getSingle();
```

The `getSingle()` method will throw when the list contains zero or more than one item.

Any thoughts?

markwoon:
+1

lhazlewood:
Ah, no, RC4 was released earlier today at 12 noon PST.  We could try to squeeze it into a follow up point release though.

chrylis:
It's certainly an improvement!

mrioan:
Implemented in https://github.com/stormpath/stormpath-sdk-java/pull/171

PR #171, opened 2015-04-06
#127 Added CollectionResource#single()

mrioan:


impl/src/test/groovy/com/stormpath/sdk/impl/account/DefaultAccountListTest.groovy
```diff
@@ -21,9 +21,14 @@ import com.stormpath.sdk.impl.resource.ArrayProperty
 import com.stormpath.sdk.impl.resource.IntegerProperty
 import org.testng.annotations.Test
```
maranai: Please remove this blank line

api/src/main/java/com/stormpath/sdk/resource/CollectionResource.java
```diff
@@ -60,4 +60,17 @@
...
     public int getSize();
 
+    /**
+     * This is just a convenience method to retrieve the single element expected to exist in this collection. This method is
+     * intended to be used in cases where the returned list is explicitly expected to contain a single element. This operation will throw
+     * an exception if this list contains zero or more than one element.
+     *
+     * <p>This method is not backed by any specific property obtained from the backend, as already mentioned, this is a convenience method.
+     *
+     * @return the single unique resource that is expect to be contained within this list</p>.
```
maranai: Not sure if this is intended but I think the </p> at the of this line needs to be moved one line above, like:
<p>This method is not ..... this is a convenience method.
</p>.
@return the single unique resource that is expect to be contained within this list

impl/src/test/groovy/com/stormpath/sdk/impl/account/DefaultAccountListTest.groovy
marioan: This blank line is separating classic imports from `static` ones.

PR #176, opened 2015-04-10
Spring security extension

mrioan:


extensions/pom.xml
```diff
@@ -35,6 +35,7 @@
         <module>servlet-plugin</module>
         <module>httpclient</module>
         <module>spring</module>
+        <module>spring-security</module>
```
lhazlewood: This should be a module under the existing spring module and renamed to `stormpath-spring-security`.  It should be a peer to `stormpath-spring`, `stormpath-spring-webmvc`, etc.

marioan: Thanks Les. It makes much more sense to have it under extensions/spring. I did not realize about that...

PR #194, opened 2015-05-18
Convenience method to provide the account verification in Application

maranai:


mrioan:
@maranai back to you. I made a few changes, if you agree with them you can merge since I consider your previous commits to be approved.

mrioan:
@maranai BTW, please remember to document this new method in the Product Guide.

maranai:
@mrioan I reviewed your changes and I see they're OK. The PR for product guide changes is https://github.com/stormpath/stormpath.github.io/pull/173, can you take a look at it?

mrioan:
@maranai I am having second thoughts about this feature... An application` could verify an account belonging to a different application. It still makes sense to give the application the chance to verify an email (it currently can only be execute by the `Tenant` resource). We need to talk to @omgitstom before we can decide what to do with this PR. We might need to request the endpoint to have a new body where we can pass the application href that is requesting the verification so the backend can authorize or reject it based on whether that accounts belongs to that application or not.

mrioan:
This feature will not be implemented

PR #203, opened 2015-06-01
Spring security extension webmvc

mrioan:


examples/spring-security-webmvc/src/main/java/tutorial/WebAppInitializer.java
```diff
@@ -36,21 +40,23 @@ public void onStartup(ServletContext sc) throws ServletException {
...
         //Stormpath Filter
         FilterRegistration.Dynamic filter = sc.addFilter("stormpathFilter", new DelegatingFilterProxy());
         EnumSet<DispatcherType> types =
                 EnumSet.of(DispatcherType.ERROR, DispatcherType.FORWARD, DispatcherType.INCLUDE, DispatcherType.REQUEST);
         filter.addMappingForUrlPatterns(types, false, "/*");
 
+        //Spring Security Filter
+        FilterRegistration.Dynamic securityFilter = sc.addFilter(AbstractSecurityWebApplicationInitializer.DEFAULT_FILTER_NAME, DelegatingFilterProxy.class);
+        securityFilter.addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), false, "/*");
+
```
maranai: Please remove this extra line

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathMethodSecurityConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+import com.stormpath.spring.security.authz.permission.evaluator.WildcardPermissionEvaluator;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.PermissionEvaluator;
+import org.springframework.security.access.expression.method.DefaultMethodSecurityExpressionHandler;
+import org.springframework.security.access.expression.method.MethodSecurityExpressionHandler;
+import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
+import org.springframework.security.config.annotation.method.configuration.GlobalMethodSecurityConfiguration;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+
```
maranai: I think this blank line can be removed as well

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+     *             if an error occurs
+     */
+    @Override
+    protected final void configure(AuthenticationManagerBuilder auth) throws Exception {
+        auth.authenticationProvider(stormpathAuthenticationProvider);
+        config(auth);
+    }
+
+    protected void config(AuthenticationManagerBuilder auth) throws Exception {
+    }
```
maranai: Can this method be overridden to specify config security settings? If so, add a comment letting users know how can it be used. Similar to what you did for the config method above.

omgitstom:
@lhazlewood The java team is blocked on this until you give you input

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathLoginSuccessHandler.java
```diff
@@ -0,0 +1,62 @@
...
+import java.io.IOException;
+
+/**
+ * @since 1.0.RC4.3
+ */
+@Component
+public class StormpathLoginSuccessHandler extends SavedRequestAwareAuthenticationSuccessHandler {
+
+    @Autowired
+    protected StormpathWebMvcConfiguration stormpathWebMvcConfiguration;
```
lhazlewood: We should never auto-wire configuration objects into Beans.  Configuration sits at a level 'above' and external to the beans used by the application.  Instead, you should inject the beans the component depends on directly, and preferably interfaces.  For example:

```java
@Autowired
@Qualifier("stormpathAuthenticationResultSaver")
private Saver<Account> authenticationResultSaver;
```

Notice that the interface is specified, but because there are a few instances of this, we have to qualify it.

It is ok to autowire configuration objects in tests if you're asserting the state of the config object directly, but, like Spring XML, the beans should never have 'knowledge' of the configuration mechanism itself.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+@EnableWebMvcSecurity
+@ComponentScan
+@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true)
+public class StormpathWebSecurityConfiguration extends WebSecurityConfigurerAdapter {
+
+    @Autowired
+    protected StormpathAuthenticationProvider stormpathAuthenticationProvider;
+
+    @Autowired
+    protected StormpathWebMvcConfiguration stormpathWebMvcConfiguration;
```
lhazlewood: Don't autowire config objects - instead, autowire the beans that are required.

The reason for this is that config objects can change, or be broken up if they become too big, or moved around, etc.  The important thing for this class to function is the beans themselves, not the config object, so we shouldn't have an explicit dependency on a Config instance.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathLogoutHandler.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * @since 1.0.RC4.3
+ */
+@Component
+public class StormpathLogoutHandler implements LogoutHandler {
+
+    @Autowired
+    protected StormpathWebMvcConfiguration stormpathWebMvcConfiguration;
```
lhazlewood: Please inject the necessary bean(s), not the config object.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+                .successHandler(stormpathLoginSuccessHandler)
+                .usernameParameter("login")
+                .passwordParameter("password")
+                .and()
+                .logout()
+                .logoutUrl(stormpathWebMvcConfiguration.stormpathInternalConfig().getLogoutUrl())
+                .logoutSuccessUrl(stormpathWebMvcConfiguration.stormpathInternalConfig().getLogoutNextUrl())
+                .addLogoutHandler(stormpathLogoutHandler)
+                .and()
+                .httpBasic()
```
lhazlewood: what about our OAuth support, forgotPassword, etc?  All of these views/controllers need to be supported.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+     *            the {@link HttpSecurity} to be modified
+     * @throws Exception
+     *             if an error occurs
+     * @see #config(org.springframework.security.config.annotation.web.builders.HttpSecurity)
+     */
+    @Override
+    protected final void configure(HttpSecurity http) throws Exception {
+        http
+                .formLogin()
+                .loginPage(stormpathWebMvcConfiguration.stormpathInternalConfig().getLoginUrl())
```
lhazlewood: Before adding a url and handler to HttpSecurity, we need to check 2 things:

1.  Did the user specify their own config for that value on the HttpSecurity object?  If so, we shouldn't override it.
2.  Is that url/handler enabled as indicated by the `stormpath.web.FEATURE_NAME.enabled` property value?  Only then should we modify the corresponding method(s) on the HttpSecurity object.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathSpringSecurityConfiguration.java
```diff
@@ -24,10 +24,9 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.PropertySource;
 
 /**
- * @since 1.0
+ * @since 1.0.RC4.3
  */
 @SuppressWarnings("SpringFacetCodeInspection")
 @Configuration
```
lhazlewood: This class belongs in the stormpath-spring-security module - there is nothing 'web' specific about it.

Another config would need to be created that incorporates web-only concepts that depends on things defined in the stormpath-spring-security config class.

lhazlewood: Additionally, we shouldn't expect users to configure class names - that's not really idiomatic for Spring.  Instead, we should provide a set of default bean instances and they can create their own and override them if they want custom logic.

PR #213, opened 2015-06-11
Login and Logout redirection now works in non-root context path

mrioan:


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/http/CookieSaver.java
```diff
@@ -62,7 +62,8 @@ public void set(HttpServletRequest request, HttpServletResponse response, String
         if (!Strings.hasText(val)) {
             val = Strings.clean(request.getContextPath());
         }
-        if (!Strings.hasText(val) || delete) {
+        //Fix for https://github.com/stormpath/stormpath-sdk-java/issues/207
+        if (!Strings.hasText(val)) {
```
lhazlewood: I think you still want to set it to `/` if `delete` is `true`, no?

examples/spring-webmvc/src/main/java/tutorial/WebAppConfig.java
```diff
@@ -31,6 +32,7 @@
 @EnableStormpath //Stormpath base beans
 @EnableStormpathWebMvc //Stormpath web mvc beans plus out-of-the-box views
 @ComponentScan
+@PropertySource("classpath:application.properties")
```
lhazlewood: Just curious - why is this necessary?  It should work w/out an application.properties file with the defaults.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/http/CookieSaver.java
mrioan: @lhazlewood In order for the delete to work properly the `path` has to have the actual path being used for that cookie. This means that if the cookie path is "/myApp" and we set the path to "/" the cookie is not cleared in the browser side. That was the problem we saw on logout not working ok. If we clear all the properties of the cookie and set the path "/myApp" then the cookie in the client browser will be cleared.

If we arrived to line 65 and `val` is empty then regardless the cookie needs to be deleted or not, the path by default can be set to "/" (as a last resort). But if we found the `path` in the previous lines (either in  `cfg` or the `request`) then we need to keep that value for the delete to be applied to the cookie.

examples/spring-webmvc/src/main/java/tutorial/WebAppConfig.java
mrioan: @lhazlewood In my case I have many Applications defined in Stormpath's Backend. The Example cannot start since it does not know which of the apps to use. If I add `stormpath.application.href = https://api.stormpath.com/v1/applications/XXXXXXXXXXX` to `resources/application.properties` it is still ignored. So, the properties file is being ignored altogether. This means that any other setting in this file will be also ignored. Adding that line got the `applications.properties` file read. BTW, the `Spring Example` has the very same line.

lhazlewood: Hrm - the examples shouldn't have anything application specific in them since users will use them to copy and paste.  We don't want them to pull in your specific URL and have things fail for them.

For your local testing in IntelliJ, you can pass in the value as a system property in your IntelliJ configuration, i.e. 

    -Dstormpath.application.href=your_value_here

We should do this for the Spring Example and all other examples as well.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/http/CookieSaver.java
lhazlewood: I see - thanks for the clarification!

examples/spring-webmvc/src/main/java/tutorial/WebAppConfig.java
mrioan: @lhazlewood The committed code does not have anything application specific. The flow for someone wanting to play with our examples are like this (at least in my mind):

1. A user downloads the source code from Github
2. A user builds the whole project: `#mvn clean source:jar install -DskipTests=true`
3. A user goes to the actual example he wants to try: `#cd examples/spring-webmvc `
4. The user runs the example: `#mvn jetty:run`
5. Assuming the `ApiKey` is in one of the expected locations:
    1. If the users has a single app in his Tenant the sample will run out of the box
    2. If the users has more than one app, the example will fail saying that no specific app has been defined and the error messages instructs him to add `stormpath.application.href = https://api.stormpath.com/v1/applications/XXXXXXXXXXX` to the sample `application.properties` file.
        1. He will add the href of this Application and it will be overlooked because the sample is not reading the `application.properties` file. Why? Because the codebase lacks the `@PropertySource("classpath:application.properties")` line
   3. Since he is just playing with the examples, he might want to tweak some other property. Then he will do so in the `application.properties` file. That tweak will be overlooked because the Example code is not reading the `application.properties` file. Why? Because the codebase lacks the `@PropertySource("classpath:application.properties")` line

In summary, the `@PropertySource("classpath:application.properties")` line is required if we want users to customize properties when trying the examples. Otherwise we need to think of some other mechanism for them to successfully play with the examples.

PR #219, opened 2015-07-06
Send verification email fix

mrioan:


impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -373,7 +376,12 @@ public ResourceDataResult filter(final ResourceDataRequest req) {
                 Map<String, Object> responseBody = getBody(response);
 
                 if (Collections.isEmpty(responseBody)) {
-                    throw new IllegalStateException("Unable to obtain resource data from the API server.");
+                    // Fix for https://github.com/stormpath/stormpath-sdk-java/issues/218
+                    if ( response.getHttpStatus() == 202 ) { //202 means that the request has been accepted for processing, but the processing has not been completed. Therefore we do not have a response body.
+                        responseBody = java.util.Collections.emptyMap();
```
josebarrueta: @mrioan why do you need to assign the reponseBody? 

shouldn't be better something in line: 378

```if(Collections.isEmpty(responseBody) && response.getHttpStatus != 202)```

File: impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java [hunk 2]
```diff
@@ -388,7 +396,11 @@ public ResourceDataResult filter(final ResourceDataRequest req) {
         ResourceDataResult result = chain.filter(request);
 
         Map<String,Object> data = result.getData();
-        Assert.notEmpty(data, "Filter chain returned an empty data result from a persistence request. This is never allowed.");
+
+        // Fix for https://github.com/stormpath/stormpath-sdk-java/issues/218
+        if (!result.getAction().equals(ResourceAction.PROCESS)) {
```
josebarrueta: Wondering why are we asserting empty body twice? any value to do it? @lhazlewood

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
mrioan: a few lines later, when the `DefaultResourceDataResult` is created, there will be an internal Exception: `data map cannot be null - specify an empty map instead of null.`

So, I need to create an empty body if I want the flow to continue its normal path.

josebarrueta: ok

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
mrioan: At line 402, the resuilts has passed through all the filter. This (previously existing) line, validates that the data is still present after this point.

The first check validates that the server returned some data. The second one checks that data still exists after filtering it.

josebarrueta: But actually not true the data from here is the one returned at this line:

```return new DefaultResourceDataResult(responseAction, uri, returnType, responseBody);```

Comment (mrioan) [inline impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java]: take a look at line 396. The `result` variable is assigned from the filter:

```
ResourceDataResult result = chain.filter(request);
```
then, in line 398, the `data` variable is assigned from that filtered result:

```
Map<String,Object> data = result.getData();
```
So, the assertion in line 402 is asserting data from the filtered result. Actually, the assertion error is explicitly saying that:

```
Filter chain returned an empty data result from a persistence request. This is never allowed.
```

Comment (mrioan) [inline impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java]: @josebarrueta is right. The chain's `completionHandler` is checking that `responseBody` is not empty. So, at line 393 we have already validated this and thus we do not need to validate that `data` is not empty again in line 399.

**Comment by mrioan** [comment]
Ready to be reviewed again @josebarrueta

File: impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java [hunk 3]
```diff
@@ -388,7 +393,11 @@ public ResourceDataResult filter(final ResourceDataRequest req) {
         ResourceDataResult result = chain.filter(request);
 
         Map<String,Object> data = result.getData();
-        Assert.notEmpty(data, "Filter chain returned an empty data result from a persistence request. This is never allowed.");
+
+        //Commenting out as this exact same validation is already being done in lines 378-385 where an IllegalStateException is thrown if the body is empty
```
lhazlewood: Please do not commit commented-code into master.  If it should not be in master, we can add it to a branch.  In this case, it looks like we can probably just delete the block entirely.

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -388,7 +393,11 @@ public ResourceDataResult filter(final ResourceDataRequest req) {
...
+        //if (!result.getAction().equals(ResourceAction.PROCESS)) {
```
lhazlewood: Also, there should not be a `ResourceAction.PROCESS` enum value.  Has it been removed?

mrioan: Yes, it has been removed.

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
mrioan: Ok. Makes sense. Thanks

josebarrueta: @mrioan As soon we remove this consider the PR approved.

Issue #225, opened 2015-07-07
Request Builder - simplify the HTTP builder so you dont have to deal with maps


omgitstom:
HttpRequestBuilder builder = HttpRequests.method(HttpMethod.POST);

builder.header(String, String[]);
builder.parameter(String, String[]);

maranai:
PR with code changes for this issue: https://github.com/stormpath/stormpath-sdk-java/pull/233/files

mrioan:
Solved in #233

PR #233, opened 2015-07-18
#225 Http request refactor

maranai:


mrioan:
@maranai I did some changes to the PR. I basically removed the ability to create the builder when solely specifying `parameter` or `header`. I also changed `header(String name, String[])` to `addHeader(String, String[])`, and `parameter(String name, String[])` to `addParameter(String, String[])`.

So, this is not valid any longer: `HttpRequests.parameter("name", "value");`

You now need to do this: `HttpRequests.method(HttpMethod.GET).addParameter("name", "value");`

Please review my changes and if you agree with them this can be considered approved.

BTW, please remember to update the `Product Guide` documentation. Thanks!

maranai:
@mrioan I could not find references to the HttpRequestBuilder or HttpRequests class in the Product Guide. Could you recommend where should this be documented?
Thanks

maranai:
Product Guide update is not required as there's no reference to these classes in it.

Issue #11, opened 2013-07-23
Expansion on account authentication


ecrisostomo:
There is no support for account expansion when authenticating an account.

This functionality needs to be supported on the SDK since it is supported on the REST API.

mrioan:
NOTE: `AccountOptions` is not supported by the server. This means that we cannot expand anything beyond the `Account` itself (like groups for example).

mrioan:
Released

PR #234, opened 2015-07-22
#11 Adding new AuthenticationOptions for Authentication Expansion

mrioan:


mrioan:
Documentation: https://github.com/stormpath/stormpath.github.io/pull/172

api/src/main/java/com/stormpath/sdk/application/Application.java
```diff
@@ -472,6 +469,28 @@
...
+     *
+     * @param request the authentication request representing an account's principals and credentials (e.g.
+     *                username/password) used to verify their identity.
+     * @param options the {@link AuthenticationOptions} to use to customize the AuthenticationResult resource upon retrieval.
+     * @return the result of the authentication.  The authenticated account can be obtained from
+     *         {@code result.}{@link com.stormpath.sdk.authc.AuthenticationResult#getAccount() getAccount()}.
+     * @throws ResourceException if the authentication attempt fails.
+     * @since 1.0.RC4.6
+     */
+    AuthenticationResult authenticateAccount(AuthenticationRequest request, AuthenticationOptions options) throws ResourceException;
```
lhazlewood: A request should encapsulate 100% of all things needed to perform the authentication - there should be no separate options argument.  Please refactor.

api/src/main/java/com/stormpath/sdk/authc/Authentications.java
```diff
@@ -0,0 +1,52 @@
...
+ * Static utility/helper methods serving Authentication-related operations. For example, to construct a {@link com.stormpath.sdk.authc.BasicAuthenticationOptions BasicAuthentication Options}:
+ *
+ * <pre>
+ * AuthenticationOptions options = Authentications.BASIC.options().withAccount();
+ * AuthenticationResult result = app.authenticateAccount(request, options);
+ * </pre>
+ *
+ * @since 1.0.RC4.6
+ */
+public final class Authentications {
```
lhazlewood: Should be removed - the AuthenticationRequest should have a Builder that allows you to set options as desired.

api/src/main/java/com/stormpath/sdk/application/Application.java
lhazlewood: P.S. The request/result paradigm avoids the method overloading problem where there are too many permutations which results in confusing code.  This is called the 'parameter object' pattern: https://sourcemaking.com/refactoring/introduce-parameter-object

mrioan: @lhazlewood totally agree with you but the current implementations of `AuthenticationRequest` (like `UsernamePasswordRequest`) are not builders. I may be able to add a wrapper to construct those requests in a new way while keeping compatibility with the old "new UsernamePasswordRequest(usernameOrEmail, password);". Do you want me to do so and also create a new issue to remove the `new UsernamePasswordRequest(...)` way before 1.0.0 while keeping only a builder interface?

lhazlewood: I don't think a wrapper is necessary - you can have UsernamePasswordRequest.builder().setUsername(username).setPassword(password).setXXX, etc for the options.

This Builder can be a sub-class of an AuthenticationRequest.Builder class that encapsulates common building options across all types of AuthenticationRequests.

mrioan: @lhazlewood Yes, I meant something like that; a `Builder` which at the end creates the specific (already available) authentication requests... wrapper was not the accurate word. 
What I do want to clarify is that we will have this new "builder way" of creating the request but we will also have the old "new UsernamePasswordRequest(username, password);" way. I guess we want to keep the old way as well in order to avoid breaking users' code, right?

lhazlewood: Yep, that should be fine.   They'll just have to use the Builder approach if they want to specify options.  Thanks!

api/src/main/java/com/stormpath/sdk/application/Application.java
```diff
@@ -15,14 +15,11 @@
  */
 package com.stormpath.sdk.application;
 
-import com.stormpath.sdk.account.Account;
-import com.stormpath.sdk.account.AccountCriteria;
-import com.stormpath.sdk.account.AccountList;
-import com.stormpath.sdk.account.CreateAccountRequest;
-import com.stormpath.sdk.account.VerificationEmailRequest;
+import com.stormpath.sdk.account.*;
```
maranai: To follow our imports policy, please replace the package import with specific imports.

extensions/oauth/src/main/java/com/stormpath/sdk/impl/oauth/authc/OauthAuthenticationRequestDispatcher.java
```diff
@@ -16,9 +16,7 @@
 package com.stormpath.sdk.impl.oauth.authc;
 
 import com.stormpath.sdk.application.Application;
-import com.stormpath.sdk.authc.AuthenticationRequest;
-import com.stormpath.sdk.authc.AuthenticationResult;
-import com.stormpath.sdk.authc.UsernamePasswordRequest;
+import com.stormpath.sdk.authc.*;
```
maranai: Same comment about the package imports. Please use specific imports.

impl/src/main/java/com/stormpath/sdk/impl/application/DefaultApplication.java
```diff
@@ -28,6 +28,7 @@
 import com.stormpath.sdk.api.ApiKeyOptions;
 import com.stormpath.sdk.api.*;
 import com.stormpath.sdk.application.*;
```
maranai: Replace package imports with specific imports.

impl/src/main/java/com/stormpath/sdk/impl/authc/AuthenticationRequestDispatcher.java
```diff
@@ -16,9 +16,7 @@
 package com.stormpath.sdk.impl.authc;
 
 import com.stormpath.sdk.application.Application;
-import com.stormpath.sdk.authc.AuthenticationRequest;
-import com.stormpath.sdk.authc.AuthenticationResult;
-import com.stormpath.sdk.authc.UsernamePasswordRequest;
+import com.stormpath.sdk.authc.*;
```
maranai: Another package import to replace here.

maranai:
@mrioan I've completed the review of this PR. I see all of  recommendations have been implemented and look good. It's approved.

mrioan:
Solves #11

PR #269, opened 2015-09-01
Fix typo

nbarbettini:
Noticed this as I was looking through the code :)


maranai:
@nbarbettini thank you, good catch!

nbarbettini:
No worries @maranai :smile:

PR #272, opened 2015-09-03
Stormpath default spring boot starter

dogeared:


extensions/spring/boot/stormpath-default-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <name>Stormpath :: Spring :: Boot :: Default ::  Starter</name>
+    <description>Spring Boot Default Starter for Stormpath with WebMVC, Thymeleaf and Spring Security</description>
+
+    <properties>
+        <tomcat.version>7.0.59</tomcat.version>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
```
lhazlewood: We shouldn't list all of these dependencies - you just need to depend on the stormpath-spring-security-webmvc-spring-boot-starter and the stormpath-thymeleaf-spring-boot-starter.  All the rest will be brought in transitively - which is what you want because if those dependencies change, you don't have to repeat them here.

extensions/spring/boot/stormpath-default-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <version>1.0.RC4.6-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>stormpath-default-spring-boot-starter</artifactId>
+    <name>Stormpath :: Spring :: Boot :: Default ::  Starter</name>
+    <description>Spring Boot Default Starter for Stormpath with WebMVC, Thymeleaf and Spring Security</description>
+
+    <properties>
+        <tomcat.version>7.0.59</tomcat.version>
```
lhazlewood: I'm pretty sure this isn't required since Spring Boot manages the embedded tomcat version already

extensions/spring/boot/stormpath-default-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.stormpath.sdk</groupId>
+            <artifactId>stormpath-sdk-httpclient</artifactId>
+            <scope>runtime</scope>
+        </dependency>
+
+        <!--
+            We need to force Tomcat 7 for the ITs to run in Java 6.
```
lhazlewood: We shouldn't specify tomcat dependencies and let Spring Boot handle this.  Specify a `provided` concrete servlet version (3.1.0) to avoid the major.minor version issue.

dogeared: Won't relying on Spring Boot to handle the tomcat dependencies have it run tomcat 8 which will automatically run servlet 3.1.0?

dogeared:
Seems like NOT having the servlet-api version `3.1.0` called out still works and all ITs run properly. Perhaps we should revisit this in the other pom files and see if it's still necessary to specify specific versions of tomcat and/or servlet?

PR #274, opened 2015-09-06
Added TwoAppTenant annotation along with test.

dogeared:
This fulfills on being able to specify a two-app tenant for certain tests as defined here: https://stormpath.atlassian.net/wiki/display/SDKS/Testing+Environments.

It also removes a bunch of semicolons to make the tests more idiomatic groovy.

To ensure that a test uses the two-app tenant, this PR introduces the: `@TwoAppTenant` annotation. 

The annotation is processed in `ClientIT` and therefor any integration test that extends `ClientIT` can make use of it.

Note: Right now, the re-organized spring boot spring security code is in a different branch. The Spring tests do not use ClientIT at all. Check out this [branch](https://github.com/stormpath/stormpath-sdk-java/compare/spring_security_extension...spring_security_extension_two_app) that addresses the Spring side of things.

Also note that in this PR, no actual tests use the `@TwoAppTenant` annotation except a test that simply exercises the annotation itself.

The next step is to identify those tests that require using a two-app tenant.


extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ClientIT.groovy
```diff
@@ -80,6 +109,8 @@ abstract class ClientIT {
             builder.setCacheManager(Caches.newDisabledCacheManager())
         }
 
+        builder.setApiKey(getApiKeyForTest())
```
mrioan: Different tests use other `buildClient(...)` methods, in this particular case you are only adding support for two-apps-tenant to `Client buildClient(boolean enableCaching=true)`.

Anyways, the overall solution does not look nice to me. Please refer to the "Conversation" tabs for a proposed solution.

mrioan:
Some general comments about this PR:
* This specific feature was required by Spring-related tests. Therefore this feature needs to provide a working solution for those tests.
* There must be a concrete test using this feature, and asserting that it is working properly
* All the Spring tests that require this feature must be changed to use it

mrioan:
@dogeared, one question... If the solution is using Environment Variables,  how are "API_KEY_ID_TWO_APP" and "API_KEY_SECRET_TWO_APP" variables established in Travis?

Could you please clarify that? Thanks

mrioan:
**Proposed solution**

Instead of having a sub-class of ClientIT, the solution should look like this:

1. Add a new method in `ClientIT.groovy`:

    Client buildClient(String apiKeyIdPropertyName, String apiKeySecretPropertyName) {

        def builder = Clients.builder()
        ((DefaultClientBuilder)builder).setBaseUrl(baseUrl)

        ApiKey apiKey = ApiKeys.builder().setIdPropertyName(apiKeyIdPropertyName).setSecretPropertyName(apiKeySecretPropertyName).build()
        builder.setApiKey(apiKey)

        return builder.build()
    }

2. Add these two constants in `ClientIT.groovy`:

        public final static DEFAULT_APIKEY_ID_PROPERTY_NAME_FOR_TENANT_WITH_TWO_APPS = "STORMPATH_API_KEY_ID_TWO_APP"
        public final static    DEFAULT_APIKEY_SECRET_PROPERTY_NAME_FOR_TENANT_WITH_TWO_APPS = "STORMPATH_API_KEY_SECRET_TWO_APP"

3. Any tests that wants to use a specific tenant, can do:

        client = buildClient(apiKeyIDPropertyName, apiKeySecretPropertyName)
    
        //Tests that want to specifically use the Tenant with 2 Apps will do: 
        client = buildClient(ClientIT.DEFAULT_APIKEY_ID_PROPERTY_NAME_FOR_TENANT_WITH_TWO_APPS, ClientIT.DEFAULT_APIKEY_SECRET_PROPERTY_NAME_FOR_TENANT_WITH_TWO_APPS)

4. Newly added `getApiKeyForTest()` method is not required any longer
5. TwoAppTenant Annotation should be deleted as it is not required
6. TwoAppTenantAnnotationIT should be deleted as it is not required

mrioan:
@dogeared back to you

dogeared:
@mrioan - addressing each question/comment below:

* Good point about the other `buildClient` methods. Could add calls to `getApiKeyForTest` in each of them OR just say that it is only that version of the method that is supported for the TwoAppTenant test. Also, ClientIT.groovy, which most (non-Spring) ITs extend, calls that `buildClient` method in its `@BeforeClass`. The result is that *any* test that inherits from ClientIT has support for TwoAppTenant without any additional checking.
* `TwoAppTenantAnnotationIT` is a concrete test that exercises the behavior of this (test only) feature
* Check here for the Spring-related version of this: https://github.com/stormpath/stormpath-sdk-java/pull/272. None of the Spring ITs extend ClientIT, so I am not sure that a single solution that works for both Spring and httpclient is doable here. Open to suggestions.
* In the pull request on the Spring side of things, there's an updated `.travis.yml` file with the `API_KEY_ID_TWO_APP` and `API_KEY_SECRET_TWO_APP` set. I can update it in this PR as well.
* In your proposed solution you say: `Instead of having a sub-class of ClientIT...` Most of the ITs (outside of Spring already extend ClientIT.
* I like the annotation approach, but I am not married to it. Let's discuss.

mrioan:
WRT: 
Regarding the one method that does the annotation processing: You are suggesting that a new method be created that explicitly passes in the values (with some defaults). For tests where this is a requirement, this will force developers to pass in those parameters and be locked into that method signature. The contract I am suggesting requires only that the @TwoAppTenant annotation be added to existing (and new) ITs - they already inherit from ClientIT - when a guarantee of the two app default is a requirement. The result is that any test that inherits from ClientIT has support for TwoAppTenant without any additional checking or coding.

My comment: 
As of today, tests that require specific `Client` instantiations rely on different "buildClient(...)" methods. The underlying `Client` object already supports customizing the ApiKey ID/Secret Property names. We must only expose that to the Tests so they can use it. You are suggesting to create an annotation with the sole purpose of customizing the actual tenant that is being used. This is:
* *Inflexible/Non reusable*: the annotation is of `TwoAppsTenant`. If in the future we need a `ThreeAppsTenant` we will need a new Annotation. This annotation can only be used for a single very specific purpose of `using a tenant that has two apps`. Why can't we just generalize this to `using any tenant you want` ?
* *Non-homogenous*: we have been using the "buildClient(...)" method for Tests to customize the client the way they need to. In this scenario a new `buildClient(...)`method is enough, no need to introduce a new concept which does not add any more value
* *Incomplete*: the TwoAppsTenant annotation will work with one of all the existing `buildClient(...)` methods.
* *Not a good design 1*: One/All existing `buildClient(...)` methods need to be modified to add `if` statements to check if the custom tenant is being used. 
* *Not a good design 2*: All test methods within a test class will be forced to use the `two apps tenant`. This will add an unnecessary restriction when writing tests. Today, each test method can customize the client to its own will.
* *Overdone*: the you are adding a bunch of code for something that the `Client` already supports.
* *Error Prone*: the "buildClient()" will have different behavior depending on whether the functionality the annotation is present in the class or not. This will lead to errors when reusing code for new tests.

My proposed approach of simply adding a new `buildClient(apiKeyIdPropName, apiKeySecretPropName)`:
* *Flexible/Reusable*: the same code can be used for any kind of tests to customize the tenant for any reason, not just for the reason of using a `tenant with two apps`
 * *Homogenous*: adheres to the current way of writing tests
* *Complete*: the tenant will always be customized when using the proper available method
* **A good design 1**: no existing code needs to be modified.
* **A good design 2**: each test method can instantiate the client the way they need to
* *Simple*: just adding a new method in the ClientIT class, which instantiate the client with the provided information
* *Errorless*: when copying code, it will behave as expected. No need to take additional aspects into consideration (like an annotation) to check if a test will change its behavior.

mrioan:
WRT: 
This approach makes it easy to identify existing ITs that would benefit from/require the default two-app setup and to add the @TwoAppTenant annotation as a next step.

My Comment:

Why is that? How will this Annotation help identify the tests that will need to use it?

mrioan:
WRT:
`TwoAppTenantAnnotationIT` is a concrete test that exercises the behavior of this (test only) feature

My Comment:
Yes, it verifies that the functionality is working, not problem with that test. But we are lacking an actual test that validates that the Annotation is in fact solving the problem it needs to solve: having Spring tests using it

mrioan:
WRT: 

Check [here](https://github.com/stormpath/stormpath-sdk-java/compare/spring_security_extension...spring_security_extension_two_app) for the Spring-related version of this. None of the Spring ITs extend ClientIT, so I am not sure that a single solution that works for both Spring and httpclient is doable here. Open to suggestions.

My Comment:

Yes, I am aware of that and this PR should solve exactly this particular problem. Otherwise it is not providing the solution to the problem that we are having.

mrioan:
WRT:

In the pull request on the Spring side of things, there's an updated .travis.yml file with the API_KEY_ID_TWO_APP and API_KEY_SECRET_TWO_APP set. I can update it in this PR as well.

My Comment:

If that file is required for this PR, it should be here.

mrioan:
WRT:

In your proposed solution you say: Instead of having a sub-class of ClientIT... Most of the ITs (outside of Spring already extend ClientIT.

My Comment:

Sorry, I did not mean to say that. I meant to say "Instead of having a new Annotation".

PR #275, opened 2015-09-09
enable dynamic property override when testing locally.

dogeared:


dogeared:
The purpose of `PropertyOverrideStormpathConfigurationIT` is to test that the PropertySource properties file in `PropertyOverrideAppConfig` will be read in and set in the application properly.

This is one place where it makes sense to have hardcoded values.

However, this is the thing that causes this test to fail when running the test suite locally against your own tenant.

This PR allows exercising the property override through the use of an environment variable: `STORMPATH_TEST_APPLICATION_HREF`. It does not interfere with other tests and, more importantly, does not interfere with the hard coded value for use in travis ci builds.

The entire integration test suite can be run locally like this:
```
STORMPATH_TEST_APPLICATION_HREF=https://api.stormpath.com/v1/applications/7WCXUThn8hhE4pFV0r9KRY mvn -DskipITs=false clean verify
```

maranai:
This PR is already merged in spring_security_extension, for this reason I'm closing it.

PR #278, opened 2015-09-11
Spring security extension two app

### LINKED ISSUE #259: Write code to ensure that relevant tests point to two-application-tenant

dogeared:
The core of this PR is `TwoAppTenantStormpathConfiguration`. It's a Spring configuration that can be used to override the default behavior for resolving the ApiKeys.

This also makes the test packages from one module available in other modules. This is so that `TwoAppTenantStormpathConfiguration`, which resides in `stormpath-spring` can be available to all of the Spring and Spring Boot modules.

To have a test use the Two App Tenant, simply add `TwoAppTenantStormpathConfiguration` as an additional class in the classes list on the `@SpringApplicationConfiguration`

```
@SpringApplicationConfiguration(classes = [BeanOverrideStormpathAutoConfigurationApplication.class, TwoAppTenantStormpathConfiguration.class])
```

NOTE: `.travis.yml` is setup to have the environment variables for both the regular tenant and the guaranteed two app tenant.

ALSO NOTE: This PR is to merge this branch into the `spring_security_extension` branch which will ultimately be merged into master.

resolves #259


api/src/main/java/com/stormpath/sdk/authc/AuthenticationOptions.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.stormpath.sdk.query.Options;
+
+/**
+ * Base Authentication-specific options that may be specified during authentication.
+ *
+ * @see com.stormpath.sdk.authc.BasicAuthenticationOptions
+ * @since 1.0.RC5
+ */
+public interface AuthenticationOptions<T extends AuthenticationOptions> extends Options {
```
lhazlewood: How is this related to Two App testing?  Something about this PR seems off...

PR #283, opened 2015-09-16
Remove Import annotations from SpringSec-SpringBoot

mrioan:


mrioan:
@lhazlewood This PR manages to have Spring Security in Spring Boot without using explicit `@Import`

extensions/spring/boot/stormpath-spring-security-webmvc-spring-boot-starter/src/main/java/com/stormpath/spring/boot/autoconfigure/StormpathWebSecurityAutoConfiguration.java
```diff
@@ -32,7 +46,7 @@
 @ConditionalOnClass({Servlet.class, Filter.class, DispatcherServlet.class})
 @ConditionalOnWebApplication
 @EnableWebSecurity
-@Import(StormpathMethodSecurityConfiguration.class)
+@ComponentScan
```
lhazlewood: Why is this necessary?  We don't want Spring to scan for annotated beans in our packages - we want them all to be defined in our *Configuration classes that are automatically picked up via Spring Boot.  I think I'm misunderstanding something here :)

lhazlewood: In other words, why didn't you list the necessary AutoConfiguration class in the `@AutoConfigureAfter` list below ?

mrioan: Without `@ComponentScan` things do not work:
1. The `restricted` page can be accessed by an anonymous user
2. The `StormpathMethodSecurityConfiguration` configuration is being completely ignored.

I have tried many different settings, this was the only thing that got things working properly. It seems that SpringBoot cannot properly auto-configure everything itself without some help like `@Import` or `@ComponentScan`

mrioan: @lhazlewood I tried both `@AutoConfigureAfter` in this class and `@AutoConfigureBefore` in the target class, neither worked.

extensions/spring/boot/stormpath-spring-security-webmvc-spring-boot-starter/src/main/java/com/stormpath/spring/boot/autoconfigure/StormpathMethodSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+package com.stormpath.spring.boot.autoconfigure;
+
+import com.stormpath.spring.config.StormpathMethodSecurityConfiguration;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @since 1.0.RC5
+ */
+@Configuration
+public class StormpathMethodSecurityAutoConfiguration extends StormpathMethodSecurityConfiguration {
```
lhazlewood: This doesn't look like our other AutoConfiguration subclasses that override a method and call super for each method - why the difference?

mrioan: @lhazlewood This class does not differ in behavior to `StormpathMethodSecurityConfiguration` so I was able to extend directly from the concrete Spring configuration. I can move the code to an abstract class and have both `StormpathMethodSecurityConfiguration` and `StormpathMethodSecurityAutoConfiguration` extend it if you consider they could differ in the past.

lhazlewood: My point is how does a user override the beans in the configuration?  Unless I'm missing something, you can't override them in a Spring Boot environment because the methods are not declared as `@ConditionalOnMissingBean` (which would only be possible if they are declared in the subclass).

This is why our other configs always override a parent class and call super - to ensure the annotations work correctly in the two different environments (traditional Spring & Spring Boot).

Does this make sense?

mrioan: @lhazlewood yes, makes complete sense but I did not manage to get this working in a different way. I will give it a second shot. I'll analyze in more detail what Spring Boot is doing in the background in order to try to come up with a new solution.

mrioan:
hey @lhazlewood, I think I got things right this time!

lhazlewood:
Nice! Looking forward to it!

lhazlewood:
This looks really good - is it ready to merge now?  If so, please go ahead and delete the branch.

mrioan:
@lhazlewood I do think it is good enough to be merged. BUT I do want you to be aware of 2 changes: 

1. I added `StormpathMethodSecurityAutoConfiguration` to `stormpath-spring-security-webmvc-spring-boot-starter/src/main/resources/META-INF/spring.factories`
2. In `StormpathWebSecurityAutoConfiguration` this:

    @AutoConfigureAfter({StormpathWebMvcAutoConfiguration.class, StormpathSpringSecurityAutoConfiguration.class})

looks like this now:

    @AutoConfigureBefore(StormpathWebMvcAutoConfiguration.class)
    @AutoConfigureAfter(StormpathSpringSecurityAutoConfiguration.class)

Why? Because we need `stormpathCsrfTokenManager` from `StormpathWebSecurityAutoConfiguration ` to take precedence over the one in `StormpathWebMvcAutoConfiguration`

If you are ok with those changes I'll merge now....

lhazlewood:
Yep, this makes sense!  Please merge :)

Issue #300, opened 2015-10-04
Integration support for multi-tenant ID Site


lhazlewood:
While IdSiteUrlBuilder already has methods in the master branch for `setOrganizationNameKey`, `setUseSubdomain` and `showOrganizationField`, the servlet & Spring/Spring Boot integrations do not yet make use of these.

lhazlewood:
The ID Site Result controller still needs to be updated:

If the user was on an application subdomain when they were redirected to ID site, they need to be returned to the same subdomain after coming back from ID Site.  Because all ID Site callbacks need to be whitelisted, dynamic (tenant-specific) subdomains cannot be pre-registered.  So the callback handler needs to do the redirect after ID Site return (i.e. probably look in the ID SIte jwt response and find the original subdomain and redirect based on that).

lhazlewood:
(note to self:  when sending the user off to ID Site, we can probably use the `state` field on IdSiteUrlBuilder to retain the subdomain in effect and look at that in the result controller).

lhazlewood:
@dogeared Have you had a chance to test things with the callback URL?  I don't think it would work if it is a subdomain (https://foo.bar.com).

lhazlewood:
I believe this issue to be blocked until we allow wildcard domains to be registered as ID Site Authorized Redirect URIs - 'using up' the state attribute is not an effective long term solution, and it puts unnecessary implementation burden on all Stormpath SDK development teams.

mrioan:
@dogeared Is this issue really solved? I am not sure about that. If so, could  you please reference the PR that solves this issue here? Otherwise this issue would needs to be re-open.

lhazlewood:
I reopened per my last comment:

> I believe this issue to be blocked until we allow wildcard domains to be registered as ID Site Authorized Redirect URIs - 'using up' the state attribute is not an effective long term solution, and it puts unnecessary implementation burden on all Stormpath SDK development teams.

dogeared:
Sorry - I confused this issue with the multi-tenant work you did in advance of the conference. My bad.

lhazlewood:
We can close this - the remaining work is only in IAM and the SDK does not need to be updated when that is fixed.

PR #301, opened 2015-10-04
#300: initial servlet controller support for supporting multi-tenant ID Site

lhazlewood:
Work in progress - do not merge yet.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/util/SubdomainResolver.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/util/SubdomainResolver.java
@@ -1,0 +1,185 @@
+package com.stormpath.sdk.servlet.util;
+
+import com.stormpath.sdk.lang.Collections;
+import com.stormpath.sdk.lang.Strings;
+import com.stormpath.sdk.servlet.http.Resolver;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
+/**
+ * @since 1.0.RC5
+ */
+public class SubdomainResolver implements Resolver<List<String>> {
+
+    private String baseDomainName;
+    private MatchMode mode = MatchMode.ALL;
+
+    public void setBaseDomainName(String baseDomainName) {
+        this.baseDomainName = baseDomainName;
+    }
+
+    public void setMode(MatchMode mode) {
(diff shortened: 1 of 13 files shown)


lhazlewood:
Still need to finish JavaDoc before we can merge this.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/util/SubdomainResolver.java
```diff
@@ -0,0 +1,185 @@
...
+        if (i > -1) {
+            host = host.substring(0, i);
+        }
+
+        return host;
+    }
+
+    public enum MatchMode {
+
+        CLOSEST,
```
dogeared: perhaps `CLOSEST_TO_BASE` and `FURTHEST_FROM_BASE`? At first I thought it was closest and furthest from the beginning.

Maybe just docs? My suggestion is more verbose.

dogeared:
The build is failing due to the presence of multiple tenants during testing. This is resolved with the TwoAppTenantStormpathConfiguration in the spring_security_extension branch. Maybe this pull request should go against that instead of master?

extensions/spring/stormpath-spring-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebMvcConfiguration.java
```diff
@@ -358,6 +361,15 @@
     @Value("#{ @environment['stormpath.web.idSite.result.uri'] ?: '/idSiteResult' }")
     protected String idSiteResultUri;
 
+    @Value("#{ @environment['stormpath.web.idSite.useSubdomain'] }")
+    protected Boolean idSiteUseSubdomain;
+
+    @Value("#{ @environment['stormpath.web.idSite.showOrganizationField] }")
```
dogeared: missing single quote. I'll add it to my pr to your branch.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/util/SubdomainResolver.java
lhazlewood: I'm thinking of removing this entirely since we don't need it.  The caller of this component configures 'ALL' and checks to see if there is exactly one subdomain, and if so, assumes that to be the organization name key.  If there are more than one then something is probably wrong (i.e. `stormpath.web.application.domain` is not configured). 

Thoughts?

PR #302, opened 2015-10-04
Added assertion in catch for invalid mode name

dogeared:


dogeared:
The build is failing due to the presence of multiple tenants during testing. This is resolved with the `TwoAppTenantStormpathConfiguration` in the spring_security_extension branch. Maybe this pull request should go against that instead of master?

lhazlewood:
This PR has no dependencies whatsoever on Spring or Spring Security, so it should be merged to master, and then master should be merged to the `spring_security_extension` branch.

Does this mean that master will always fail until we merge the `spring_security_extension` branch into master?

Issue #299, opened 2015-10-03
If a user restricts "/**" then no Stormpath page can be accessed


mrioan:
If a user does something like this:
```
   @Override
   protected void configure(HttpSecurity http) throws Exception {
        http
                .authorizeRequests()
                .antMatchers("/restricted").authenticated()
                .antMatchers("/**").denyAll();
    }
```
Then our endpoints (like /login, /register, etc) are not accessible.

lhazlewood:
Hrm - need to find out why.  I thought our configuration occurs first in
Spring Boot startup before the users executes and so ours would be allowed.

--
Les

mrioan:
I already have a fix for this. Spring Security has a way to force patterns
to be always available. I did so for all of our *enabled* endpoints

On Oct 3, 2015 14:55, "Les Hazlewood" <notifications@github.com> wrote:
>
> Hrm - need to find out why. I thought our configuration occurs first in
> Spring Boot startup before the users executes and so ours would be
allowed.
>
> --
> Les
>

lhazlewood:
Nice!

--
Les
@lhazlewood

dogeared:
Awesome!

==============================================
Micah Silverman, CSM <http://bit.ly/WikiScrum>
Phone: 631.606.8928
Co-Author: Mastering Enterprise Javabeans 3.0 <http://bit.ly/MasteringEJB3>
==============================================
"Debugging is twice as hard as writing the code in the first
place. Therefore, if you write the code as cleverly as
possible, you are, by definition, not smart enough to debug
it." by Brian Kernighan
==============================================
[image: Namez]
<http://namez.com/profiles/1720-micah-silverman/?autoPlay=true>

mrioan:
Solved in #304

Issue #295, opened 2015-10-02
CSRF not working in register page


mrioan:

mrioan:
Solved in #304

PR #304, opened 2015-10-06
Spring security extension configuration fix

### LINKED ISSUE #303: Anonymous user not redirect to login when trying to access a resource that requires Authentication
User is not redirected to login if an anonymous user tries to access a resource that requires authentication as specified in sample configuration via `configure(HttpSecurity http)`. They see the “access denied” page right away.

We had the false impression that sample's `configure(HttpSecurity http)` was working since `HelloService` was doing that because of `@PreAuthorize("hasRole('" + MY_GROUP + "') and hasPermission('say', 'hello')")`. If we remove the `@PreAuthorize` annotation then the we see an "access denied" page instead of being redirected to the `login` page.

mrioan:
Fixes #303, #299 and #295


dogeared:
This looks good @mrioan - please merge!

Issue #206, opened 2015-06-03
Loading apiKeys from classpath: resource doesnt work


maxkremer:
I'm trying to load the api keys like so:

```
ApiKey apiKey  =ApiKeys.builder().set("classpath:stormpathkeys.properties").build();
```

Causes

```
java.lang.IllegalArgumentException: Unable to read properties from specified apiKeyFileLocation [classpath:stormpathApiKey.properties].
	at com.stormpath.sdk.impl.api.ClientApiKeyBuilder.build(ClientApiKeyBuilder.java:220)
```

This used to work in version 0.8 of the api

lhazlewood:
Shouldn't it be the following?

```java
ApiKey apiKey  = ApiKeys.builder().setFileLocation("classpath:stormpathkeys.properties").build();
```

lhazlewood:
Also, something feels weird - the exception message references a location of `classpath:stormpathApiKey.properties` but your code example references `classpath:stormpathkeys.properties`

maxkremer:
Sorry my bad, a few typos in the example code. It should say:
```
ApiKey apiKey = ApiKeys.builder().setFileLocation("classpath:stormpathApiKey.properties").build();
```
and
```
java.lang.IllegalArgumentException: Unable to read properties from specified apiKeyFileLocation [classpath:stormpathApiKey.properties]
```

I've also confirmed that if I specify the location via file path it works fine.
```
//This works
ApiKey apiKey  =ApiKeys.builder().setFileLocation("/home/max/rest-play/conf/stormpathApiKey.properties").build();
```

maxkremer:
I took a look at the source and it looks like the `ClientApiKeyBuilder` makes use of the `FileResource` but I can't find any logic that uses the `ClasspathResource`.

It appears there is a `DefaultResourceFactory` which seems to be responsible for deciding which type of resource to create (URL, File or Classpath) but it doesn't seem to be used anywhere.

maxkremer:
@lhazlewood have you had a chance to confirm this bug?

mrioan:
@maxkremer I have just tried this and it works:

`ApiKey apiKey = ApiKeys.builder().setFileLocation("stormpathApiKey.properties").build();`

*Note the missing `classpath:` prefix.

Can you please confirm that it works for you?

maxkremer:
No that doesn't work. 

```
[...]

PR #311, opened 2015-10-16
#206 apikey file can now be obtained from classpath

mrioan:
Like this: `classpath:apikey.properties`


mrioan:
solves #206

mrioan:
Waiting on user response. We are not still sure this is a valid scenario as we do not want users to add their ApiKeys in the classpath

maranai:
PR approved. Merge pending, see Mario's last comment.

PR #312, opened 2015-10-16
Spring security extension documentation

dogeared:
- Added in tutorials section.
- Renamed `tutorial` package to `example` package in examples folders to avoid confusion
- updated docs to add tutorial docs to go along with code.

Also: Don't be alarmed by the removal of a good chunk of the top of the quickstart doc. I moved that into its own file as its now used in both the quickstart and tutorial docs.

You can see the docs live, including the new tutorial docs in the `java_updates` branch of https://github.com/stormpath/stormpath.github.io


tutorials/spring-boot-default/04-a-finer-grain-of-control/src/main/java/com/stormpath/tutorial/controller/HelloController.java
```diff
@@ -0,0 +1,56 @@
...
+        model.addAttribute("status", req.getParameter("status"));
+        return "home";
+    }
+
+    @RequestMapping("/me")
+    String me() {
+        return "me";
+    }
+
+    @RequestMapping("/restricted")
```
lhazlewood: This URI implies it is restricted, but it's not really - there is not authz check.  I think we should change the URL to reflect what is happening, like "/hello".

lhazlewood: Ah, I see - you're restricting the service via `@PreAuthorize`

Issue #281, opened 2015-09-15
Stormpath default spring boot starter


dogeared:
Add a default starter that incorporates `stormpath-spring-security-webmvc-spring-boot-starter` and `stormpath-thymeleaf-spring-boot-starter`

mrioan:
Released

Issue #271, opened 2015-09-03
CSRF incompatibility with Spring Security


mrioan:
CSRF handling: Right now there are two separate systems for csrf. There's the one from Stormpath (that's also baked into servlet). There's also the default csrf system built into Spring Security. These systems are incompatible. So, if you run the spring-security-spring-boot-webmvc example app, for instance, you won't be able to submit the /login page because Spring Security complains that there's no csrf token

Possible solution: CsrfManager interface that delegates to Spring Security's CsrfTokenRepository

mrioan:
Released

Issue #267, opened 2015-08-28
Ensure all Spring Boot modules support property metadata for IDE autocompletion


lhazlewood:
http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#configuration-metadata-additional-metadata

mrioan:
Released

PR #315, opened 2015-10-19
Spring security extension

mrioan:


mrioan:
Solves  #179 #183 #189 #214 #227 #240 #241 #267 #271 #281 #295 #303

PR #327, opened 2015-10-30
Refactored examples to all be in examples package. 

dogeared:
Updated main pom.xml to exclude impl, examples and tutorial packages. Updated build and publish doc scripts to build and publish javadocs for release.

addresses #326


pom.xml
```diff
@@ -706,6 +706,9 @@
     </script>
     <script src="//cdn.optimizely.com/js/225847041.js"></script>
                                 ]]></top>
+                            <excludePackageNames>
+                                *.impl:*.examples:*.tutorial.*:quickstart
```
mrioan: Not an expert on wildcards in this environment but I think something like this will be excluded: `com.stormpath.sdk.account.some.other.api.package.impl`, right? In that case, we need to be more explicit here, like this: `com.stormpath.sdk.impl` and `com.stormpath.spring.boot.examples` and `com.stormpath.spring.examples` etc....
 Besides the quickstart separator should be `:*.` instead of `.*:`

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/authz/permission/evaluator/WildcardPermissionEvaluator.java
```diff
@@ -29,7 +29,8 @@
  * A {@link PermissionEvaluator} that determines if a {@link WildcardPermission} matches a given permission.
  * <h3>Usage</h3>
  * In order to use it you need to configure Spring this way:
```
mrioan: When checking the javadoc output I found that this javadoc was not being well formatted. I just fixed it myself.

mrioan:
@dogeared back to you

pom.xml
dogeared: I thought we wanted to always exclude `impl` packages?

The pattern is correct. for tutorial, it's: `*.tutorial.*`

For quickstart it's just: `quickstart` (that's the entire package path for quickstart in the examples section)

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/authz/permission/evaluator/WildcardPermissionEvaluator.java
dogeared: great catch!

pom.xml
mrioan: In muy example I justo try to bring up that a user developing something in the api module could be creating an package called impl, which I think could be a valid scenario. So I would just make the exclusion just more explicit so online the simple module is left aside but not every package that has a name impl.

Issue #328, opened 2015-10-30
Fix bug where empty value in key value pair is causing servlet to blow up


dogeared:
We recently added properties to `default.stormpath.properties` that don't have values. 

This is explicitly not allowed by the current `DefaultPropertiesParser`

mrioan:
The fix is to change `DefaultPropertiesParser` to allow that or we will be removing `default.stormpath.properties` altogether?

dogeared:
This was addressed by #329

lhazlewood:
I'm confused - all of our config properties start with `stormpath.` by design (so they can be filtered, auto-completed, easily discovered in projects, etc).  Do we still have a `default.stormpath.properties` ?

PR #329, opened 2015-10-30
Refactored DefaultPropertiesParser

dogeared:
Refactored to allow empty properties. Added test for `DefaultPropertiesParser`.

Current `default.stormpath.properties` has empty keys in it which is causing `DefaultPropertiesParser` to blow up in servlet.

addresses #328

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/config/DefaultPropertiesParser.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/config/DefaultPropertiesParser.java
@@ -151,8 +151,8 @@ public class DefaultPropertiesParser implements PropertiesParser {
         String key = Strings.clean(keyBuffer.toString());
         String value = Strings.clean(valueBuffer.toString());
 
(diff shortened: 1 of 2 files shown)


mrioan:
If I do not understand wrongly, empty values are now allowed and they are mapped to `null`, right? Why was not an issue in the past?

mrioan:
Back to you

dogeared:
In the past, we explicitly did not allow keys with empty values.

At the beginning of October, @lhazlewood introduced two empty properties in `default.stormpath.properties` to support multi-tenant (organizations) and sub-domains.

While this works fine with Spring Boot and ID Site, it was breaking servlet because the `DefaultPropertiesParser` was barfing on empty properties values.

If you try to fire up the servlet example right now, it will fail because it is not able to load `default.stormpath.properties`.

mrioan:
@dogeared Understood! Now that I understand what is happening this is approved! Thanks for taking core of this.

Issue #343, opened 2015-11-09
Add sphinx admonition about the circular redirect issue


dogeared:
Google Chrome aggressively caches 302 redirects in a way that can result in a redirect loop when you are first getting Started with the Stormpath Spring Boot integration.

The solution to clear cache.

This should be added as an admonition to the docs.

dogeared:
Addressed by #346

Issue #342, opened 2015-11-09
Add information to the Quickstart "having problems" section: spring-boot-parent examples


dogeared:
When using the `spring-boot-starter-parent`, you *must* override the Spring Security version to 4.0.2.

`spring-boot-parent` is often used in conjunction with Spring Intializr. We don't recommend using the spring-boot-parent, but we should have an answer if people run into this.

We should reference this in the quickstart.

dogeared:
Addressed by #346

PR #346, opened 2015-11-11
Updated quickstart to include issues with spring-boot-starter-parent

dogeared:
addresses #342 
addresses #343

### DIFF
--- a/extensions/spring/boot/docs/source/quickstart.rst
+++ b/extensions/spring/boot/docs/source/quickstart.rst
@@ -70,6 +70,17 @@ Don’t believe it? Try it! Start up your Spring Boot web application, and we'll
 
 Wasn't that easy?!
 
+.. ATTENTION::
+
+   You may encounter a circular redirect in your Google Chrome browser when you are in the process of developing your
+   Spring Boot + Spring Security + Stormpath application.
+
+   This is due to a known issue with Google Chrome whereby it aggressively caches ``302`` redirects.
+
+   The solution is to clear cache or use an Icognito window while testing.
+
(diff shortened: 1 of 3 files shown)


extensions/spring/boot/docs/source/quickstart.rst
```diff
@@ -97,8 +108,38 @@ Did you experience any problems with this quickstart?  It might not have worked
 
   By default, the ``StormpathFilter`` is ordered as ``Ordered.HIGHEST_PRECEDENCE``, but if you have multiple filters with that same order value, you might have to change the order of the other filters as well.
 
-
-
+* you're using the ``spring-boot-starter-parent`` as a ``parent`` and you are seeing the wrong version of Spring Security. The current release of the ``spring-boot-starter-parent`` uses Spring Security, version 3.2.8. The ``stormpath-default-spring-boot-starter`` needs Spring Security, version 4.0.x. An easy solution is to override the Spring Security version in your ``pom.xml``
+
```
mrioan: please explain here what will happen when they do this.

mrioan: Instead of "An easy solution is to override..." I would say "There is a simple solution to this which is to override...". The phrase "an easy solution is to..." makes me thing that you will also present another (difficult) solution afterwards.

mrioan:
Back to you @dogeared

dogeared:
@mrioan - I implemented your suggestion about the "simple solution".

I was not sure what you meant by tell them what happens here.

I did update to let the reader know that they might see errors related to Spring Security until they override the version property.

Back to you... ;)

PR #373, opened 2015-12-01
338 token management servlet plugin

maranai:


maranai:
Closing as this PR was created against master and should have been against the token management branch.

Issue #292, opened 2015-09-29
Exchanging ID Site token for a Stormpath Access Token


omgitstom:
https://stormpath.atlassian.net/wiki/display/AM/ID+Site+OAuth+Access+Token+Grant+Type

omgitstom:
Open question here from Support:

How does this feature effect customers developing against:
stormpath-shiro-servlet-0.7.1-idsite 
stormpath-shiro-core-0.7.1-idsite

Is there any code changes / configuration needed to take advantage of using access tokens?

maranai:
Related PR (https://github.com/stormpath/stormpath-sdk-java/pull/436) needs to be reviewed. I has been updated to modify the grant_type as per Les' comment.

dogeared:
closes #436

PR #376, opened 2015-12-02
 #292 Exchanging ID Site token for a Stormpath Access Token

maranai:


impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultIdSiteAuthenticationRequest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.stormpath.sdk.lang.Assert;
+import com.stormpath.sdk.oauth.IdSiteAuthenticationRequest;
+
+/**
+ * @since 1.0.RC7
+ */
+public class DefaultIdSiteAuthenticationRequest implements IdSiteAuthenticationRequest {
+
+    private final static String grant_type = "id_site_token";
```
lhazlewood: this should be `stormpath_token` per the latest spec

PR #380, opened 2015-12-04
setup ID Site to set Spring Security authentication when applicable

dogeared:
The motivation for this design is that:

1. We do not need a different `IDSiteResultController` when Spring Security is in use

2. By the time we get into the methods of an `IDSiteResultListener`, authentication via ID Site has already occurred. In the context of Spring Security, all that is necessary is to set the authentication in the `SecurityContextHolder`

With that in mind, this is a summary of the primary changes in this PR:

1. `IdSiteResultController` now has a `List` of `IdSiteResultListener` rather than just one `IdSiteResultListener`

2. `IdSiteCallbackHandler` now has a `List` of `IdSiteResultListener` rather than just one `IdSiteResultListener`

3. In `AbstractStormpathWebMvcConfiguration`, inside the `stormpathIdSiteResultController` method, the following lined is executed: `controller.addIdSiteResultListener(springSecurityIdSiteResultListener)`.

4. The Spring Security `IdSiteResultListener` is exposed as a bean in `StormpathWebSecurityAutoConfiguration` and `StormpathWebSecurityConfiguration`

The upshot of this configuration is that when Spring Security is enabled, in addition to the default `IdSiteListener` being registered in `IdSiteResultController`, a Spring Security `IdSiteListener` will be registered with `IdSiteResultController`.

This approach introduces minimal changes to `IdSiteResultController`, alleviates the need for a Spring Security specific `IdSiteResultController` and ensures that authentication, registration and logout work through the stack with Spring Security and ID Site.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
@@ -19,6 +19,7 @@ import com.stormpath.sdk.account.Account;
 import com.stormpath.sdk.account.AccountStatus;
 import com.stormpath.sdk.application.Application;
 import com.stormpath.sdk.authc.AuthenticationResult;
+import com.stormpath.sdk.idsite.IdSiteCallbackHandler;
 import com.stormpath.sdk.idsite.IdSiteResultListener;
 import com.stormpath.sdk.idsite.LogoutResult;
 import com.stormpath.sdk.idsite.RegistrationResult;
@@ -33,6 +34,8 @@ import com.stormpath.sdk.servlet.http.Saver;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+import java.util.List;
 
 public class IdSiteResultController extends AbstractController {
 
@@ -42,6 +45,8 @@ public class IdSiteResultController extends AbstractController {
     private Controller logoutController;
     private Saver<AuthenticationResult> authenticationResultSaver;
     private Publisher<RequestEvent> eventPublisher;
+
+    private List<IdSiteResultListener> idSiteResultListeners = new ArrayList<IdSiteResultListener>();
 
     public void setRegisterNextUri(String registerNextUri) {
         this.registerNextUri = registerNextUri;
@@ -71,6 +76,12 @@ public class IdSiteResultController extends AbstractController {
         this.authenticationResultSaver = authenticationResultSaver;
     }
 
+    public void addIdSiteResultListener(IdSiteResultListener resultListener) {
+        if (resultListener != null) {
+            idSiteResultListeners.add(resultListener);
+        }
+    }
+
     public void init() {
         Assert.hasText(registerNextUri, "registerNextUri must be configured.");
         Assert.hasText(loginNextUri, "loginNextUri must be configured.");
@@ -90,7 +101,7 @@ public class IdSiteResultController extends AbstractController {
 
         final ViewModel[] viewModel = new ViewModel[1];
 
-        app.newIdSiteCallbackHandler(request).setResultListener(new IdSiteResultListener() {
+        IdSiteCallbackHandler idSiteCallbackHandler = app.newIdSiteCallbackHandler(request).setResultListener(new IdSiteResultListener() {
             @Override
             public void onRegistered(RegistrationResult result) {
                 viewModel[0] = IdSiteResultController.this.onRegistration(request, response, app, result);
@@ -106,7 +117,13 @@ public class IdSiteResultController extends AbstractController {
                 viewModel[0] = IdSiteResultController.this.onLogout(request, response, app, result);
 
             }
-        }).getAccountResult();
+        });
+
+        for (IdSiteResultListener resultListener : idSiteResultListeners) {
+            idSiteCallbackHandler.addResultListener(resultListener);
+        }
+
+        idSiteCallbackHandler.getAccountResult();
 
         return viewModel[0];
     }
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
@@ -77,9 +77,8 @@ public class IdSiteResultController extends AbstractController {
     }
 
     public void addIdSiteResultListener(IdSiteResultListener resultListener) {
-        if (resultListener != null) {
-            idSiteResultListeners.add(resultListener);
-        }
+        Assert.notNull(resultListener, "resultListener cannot be null");
+        idSiteResultListeners.add(resultListener);
     }
 
     public void init() {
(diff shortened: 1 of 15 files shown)


mrioan:
@dogeared I committed a few changes to this branch. Could you please take a look at them? Thanks

dogeared:
@mrioan - I will look at this, but the change has caused a failing test that was not failing before:

```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.web.servlet.mvc.Controller]: Factory method 'stormpathIdSiteResultController' threw exception; nested exception is java.lang.IllegalArgumentException: resultListener cannot be null
```

dogeared:
Hey @mrioan - I implemented most of the changes you suggested over in https://github.com/stormpath/stormpath-sdk-java/pull/395.

I would like to keep this PR as the one that will be ultimately merged as it has the latest from master rebased in, keeping the commits orderly. Sorry if this creates extra work for you!

The one thing I did *not* do was to add in the `PreAuthenticatedAuthenticationToken` class and the associated changes.

It's a little dated, but I got the pattern that I used from an SO post from Rob Winch: http://stackoverflow.com/questions/11313309/spring-security-programmatic-login-without-a-password/11314388#11314388

Given the age of his response, I realize it may not be best practices now.

However, the solution I have has a very small footprint and ensures that all granted authorities (groups & permissions) are registered with Spring Security.

Since, in this circumstance, we *know* that the user has authenticated to Stormpath, this seems like a reasonable approach. 

Let's discuss tomorrow.

Thanks!

dogeared:
@mrioan - Please review the latest. I've incorporated all the suggested changes you've made, including using the `PreAuthenticatedAuthenticationToken`.

However, rather than use a custom version of the `PreAuthenticatedAuthenticationToken`, I used the one that's already built into Spring Security:

`org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken`

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/security/provider/SpringSecurityIdSiteResultListener.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public void onLogout(LogoutResult result) {
+        SecurityContextHolder.clearContext();
+    }
+
+    private void doAuthenticate(Account account) {
+        SecurityContextHolder.clearContext();
+        Authentication authentication = new PreAuthenticatedAuthenticationToken(
+            account.getEmail(), null, authenticationProvider.getGrantedAuthorities(account)
```
mrioan: @dogeared Rather than doing 

`Authentication authentication = new PreAuthenticatedAuthenticationToken(account.getEmail(), null, authenticationProvider.getGrantedAuthorities(account)`

you should do
```
Authentication authentication = new PreAuthenticatedAuthenticationToken(account.getEmail(), null, account);
authentication = authenticationProvider.authenticate(authentication);
```

this way you leave all the responsibility of how to populate roles and granted authorities to the `stormpathProvider.`

Otherwise, if in the future the Provider needs to do new things to authenticated tokens then this code will need to change as well.

mrioan: WRT: http://stackoverflow.com/questions/11313309/spring-security-programmatic-login-without-a-password/11314388#11314388

It is saying "you do not need to" rather than "you must not to". I think we should always let a token pass through the provider in order for it to be the sole responsible to create and populate authenticated tokens with its corresponding roles and granted authorities.

mrioan:
@dogeared back to you

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
```diff
@@ -90,7 +100,7 @@ protected ViewModel doGet(final HttpServletRequest request, final HttpServletRes
 
         final ViewModel[] viewModel = new ViewModel[1];
 
-        app.newIdSiteCallbackHandler(request).setResultListener(new IdSiteResultListener() {
+        IdSiteCallbackHandler idSiteCallbackHandler = app.newIdSiteCallbackHandler(request).setResultListener(new IdSiteResultListener() {
```
lhazlewood: We need to remove the internal calls, correct?  I thought we were extracting this logic out of the IdSiteResultController, no?  Thoughts/

impl/src/test/java/com/stormpath/sdk/impl/idsite/IdSiteResultListenerType.java
```diff
@@ -0,0 +1,5 @@
+package com.stormpath.sdk.impl.idsite;
+
+public enum IdSiteResultListenerType {
```
lhazlewood: Hrm - this has a funny smell to me (usually creating any 'type' concept in a type-safe language is a yellow flag for me).  Why is this necessary?

lhazlewood: Ah, this is for testing only - I'm much less concerned now :)

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/security/provider/SpringSecurityIdSiteResultListener.java
dogeared: Got it. I will move all responsibility into authenticationProvider so we don't have to go hunting for where we authenticate in various places later. Thanks.

impl/src/test/java/com/stormpath/sdk/impl/idsite/IdSiteResultListenerType.java
dogeared: yeah, just testing. ;)

dogeared:
@mrioan - I've addressed ensuring that all authentication related activity goes through the `authenticationProvider`.

Thanks for the feedback - it's much tighter now.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
dogeared: @lhazlewood - we did discuss that originally. I thought this implementation has a much smaller footprint, which is one of the reasons I wanted you to take a look at it.

As it stands, IdSite integration with Spring Security is accomplished without having to make big changes here.

If we still want to pull this logic out, we can do it separately now.

lhazlewood: Nah, no reason to pull it out just yet.  Since this is an implementation detail that SDK users don't see, it should be ok.

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/token/IdSiteAuthentication.java
```diff
@@ -0,0 +1,59 @@
...
+package com.stormpath.spring.security.token;
+
+import com.stormpath.sdk.account.Account;
+import com.stormpath.sdk.lang.Assert;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+
+/**
+ * @since 1.0.RC7.2
+ */
+public class IdSiteAuthentication extends AbstractAuthenticationToken {
```
lhazlewood: This should be called `IdSiteAuthenticationToken`

dogeared: Yeah - it's a little strange.

`AbstractAuthenticationToken` implements `Authentication` and the `AuthenticationProvider.authenticate` take an `Authentication` object as its parameter.

While it looks a little odd here, I thought it looked better in `StormpathAuthenticationProvider.authenticate`

I'll push up the refactor and let me know what you think. It's easy enough to revert if we don't like it...

lhazlewood:
LGTM!  Please merge after CI is successful.

Issue #390, opened 2015-12-08
If all paths are locked down in servlet, Stormpath view assets require authentication


dogeared:
If a user specifies locking down all paths in `stormpath.properties`:

```
stormpath.web.uris./** = authc
```

It causes assets that the Stormpath views for login, registration, etc. to require authentication as well.

Currently, these assets are:

```
/assets/css/stormpath.css
/assets/css/custom.stormpath.css
```

These should always be publicly accessible regardless of user configuration.

PR #391, opened 2015-12-08
Ensure that assets required to render Stormpath views are always accessible without authentication regardless of user configuration

dogeared:
If a user specifies locking down all paths in `stormpath.properties` when using the servlet plugin:

```
stormpath.web.uris./** = authc
```
It causes assets that the Stormpath views for login, registration, etc. to require authentication as well.

Currently, these assets are:

```
/assets/css/stormpath.css
/assets/css/custom.stormpath.css
```

These should always be publicly accessible regardless of user configuration.

Addresses #390


maranai:
PR is approved.

Issue #392, opened 2015-12-09
Spring Security: Can't use defaultSuccessUrl and  successHandler together


dogeared:
*Note:* I think this is actually a Spring Security bug and I will log an issue with Rob Winch.

In the current (and many previous releases) of Spring Security, the call to `defaultSuccessUrl` instantiates a `AuthenticationSuccessHandler`, sets the `defaultSuccessUrl` in it and installs it as the configured `AuthenticationSuccessHandler`.

A subsequent call to `successHandler` clobbers the `AuthenticationSuccessHandler` that had been previously set and you end up with the default `defaultSuccessUrl` which is `/`.

However, there's an easy workaround for this which is to set the `defaultSuccessUrl` when the `StormpathLoginSuccessHandler` is instantiated.

qmn1711:
@dogeared: I know it's two years later already but did you report this bug on somewhere? I'm running Spring Security 4.1.3.RELEASE and meet exactly the same your scenario.

PR #393, opened 2015-12-09
set loginNextrUri when creating StormpathSuccessHandler so it doesn't get clobbered in Spring Security config.

dogeared:
*Note:* I think this is actually a Spring Security bug and I will log an issue with Rob Winch.

In the current (and many previous releases) of Spring Security, the call to `defaultSuccessUrl` instantiates a `AuthenticationSuccessHandler`, sets the `defaultSuccessUrl` in it and installs it as the configured `AuthenticationSuccessHandler`.

A subsequent call to `successHandler` clobbers the `AuthenticationSuccessHandler` that had been previously set and you end up with the default `defaultSuccessUrl` which is `/`.

However, there's an easy workaround for this which is to set the `defaultSuccessUrl` when the `StormpathLoginSuccessHandler` is instantiated.

Addresses #392


mrioan:
Approved

PR #395, opened 2015-12-09
Idsite spring security mrioan

dogeared:


api/src/main/java/com/stormpath/sdk/idsite/IdSiteCallbackHandler.java
```diff
@@ -107,7 +107,9 @@
      * });
      * Account account = callbackHandler.getAccountResult().getAccount();
      * </code>
-     * </pre
+     * </pre>
+     * <p>Usage Note: setting a listener using this method will reset the current list removing any listener previously added via
+     * the {@link #addResultListener(IdSiteResultListener)} method.</p>
      *
```
dogeared: I will incorporate this into my branch. Thanks!

api/src/main/java/com/stormpath/sdk/idsite/IdSiteCallbackHandler.java
```diff
@@ -118,4 +120,17 @@
...
+     *
+     * @param resultListener the {@link IdSiteResultListener} that will be added to the list of listeners which will be notified
+     *                       about the actual operation of the ID Site invocation: registration, authentication or logout.
+     *                       will be sent.
+     * @return this instance for method chaining
+     * @see IdSiteResultListener
+     * @since 1.0.RC7.2
+     */
+    IdSiteCallbackHandler addResultListener(IdSiteResultListener resultListener);
 }
```
dogeared: I will incorporate this into my branch. Thanks!

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
```diff
@@ -71,6 +76,11 @@ public void setAuthenticationResultSaver(Saver<AuthenticationResult> authenticat
         this.authenticationResultSaver = authenticationResultSaver;
     }
 
+    public void addIdSiteResultListener(IdSiteResultListener resultListener) {
+        Assert.notNull(resultListener, "resultListener cannot be null");
+        idSiteResultListeners.add(resultListener);
+    }
+
```
dogeared: There needs to be a null check here. This is why some of the tests are failing.

extensions/spring/boot/stormpath-spring-security-webmvc-spring-boot-starter/src/main/java/com/stormpath/spring/boot/autoconfigure/StormpathWebSecurityAutoConfiguration.java
```diff
@@ -85,4 +85,11 @@ public CsrfTokenManager stormpathCsrfTokenManager() {
     public ErrorModelFactory stormpathLoginErrorModelFactory() {
         return super.stormpathLoginErrorModelFactory();
     }
+
+    @Bean
+    @ConditionalOnMissingBean(name="springSecurityIdSiteResultListener")
+    @ConditionalOnProperty(name="stormpath.web.idSite.enabled")
```
dogeared: Les told us in prior PRs to remove the `@ConditionalOnProperty`'s here. The `@ConditonalOnProperty` at the top of the class takes care of it.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/security/listener/SpringSecurityIdSiteResultListener.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public void onLogout(LogoutResult result) {
+        SecurityContextHolder.clearContext();
+    }
+
+    protected void doAuthenticate(Account account) {
+        SecurityContextHolder.clearContext();
+        Authentication authentication = new PreAuthenticatedAuthenticationToken(account.getEmail(), null, account);
```
dogeared: I think the `PreAuthenticationToken` class adds an unnecessary layer of complexity. It leads to branching and `instanceof` checks later on.

Do you know if the GrantedAuthorities are taken care of here? In my original PR, the `authenticationProvider.getGrantedAuthorities(account)` ensures that all `group` and `permissions` are registered.

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -351,7 +352,12 @@ public Authentication authenticate(Authentication authentication) throws Authent
         Account account;
 
         try {
-            account = application.authenticateAccount(request).getAccount();
+            if (authentication instanceof PreAuthenticatedAuthenticationToken) {
```
dogeared: Don't like having an extra `if` and `instanceof` calls unnecessarily, per above. Why is this important?

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/token/PreAuthenticatedAuthenticationToken.java
```diff
@@ -0,0 +1,64 @@
...
+package com.stormpath.spring.security.token;
+
+import com.stormpath.sdk.account.Account;
+import com.stormpath.sdk.lang.Assert;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+
+/**
+ * @since 1.0.RC7.2
+ */
+public class PreAuthenticatedAuthenticationToken extends AbstractAuthenticationToken {
```
dogeared: Why is this needed?

impl/src/main/java/com/stormpath/sdk/impl/idsite/DefaultIdSiteCallbackHandler.java
```diff
@@ -250,29 +261,35 @@ private JwtSignatureValidator getJwtSignatureValidator(String jwtApiKeyId) {
...
      * <ul>
-     *     <li> Registered -> {@link com.stormpath.sdk.idsite.IdSiteResultListener#onRegistered(com.stormpath.sdk.idsite.RegistrationResult) IdSiteResultListener#onRegistered(RegistrationResult)}</li>
-     *     <li> Authenticated -> {@link com.stormpath.sdk.idsite.IdSiteResultListener#onAuthenticated(com.stormpath.sdk.idsite.AuthenticationResult) IdSiteResultListener#onAuthenticated(AuthenticationResult)} </li>
-     *     <li> Logout -> {@link com.stormpath.sdk.idsite.IdSiteResultListener#onLogout(com.stormpath.sdk.idsite.LogoutResult) IdSiteResultListener#onLogout(LogoutResult)} </li>
+     *     <li> Registered -> {@link IdSiteResultListener#onRegistered(com.stormpath.sdk.idsite.RegistrationResult) IdSiteResultListener#onRegistered(RegistrationResult)}</li>
+     *     <li> Authenticated -> {@link IdSiteResultListener#onAuthenticated(com.stormpath.sdk.idsite.AuthenticationResult) IdSiteResultListener#onAuthenticated(AuthenticationResult)} </li>
+     *     <li> Logout -> {@link IdSiteResultListener#onLogout(com.stormpath.sdk.idsite.LogoutResult) IdSiteResultListener#onLogout(LogoutResult)} </li>
      * </ul>
```
dogeared: Why remove the fully qualified package names in some cases and leave them in in others? Is it because they're in the same package?

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
```diff
@@ -150,33 +153,44 @@ public void init(HttpSecurity http) throws Exception {
         ApplicationContext context = http.getSharedObject(ApplicationContext.class);
         context.getAutowireCapableBeanFactory().autowireBean(this);
 
-        if (stormpathWebEnabled) {
+        if (idSiteEnabled) { //IDSite
```
dogeared: Not sure this branch is necessary. What about registration, forgot, etc. pages?

We *do* need to add a rule to permit the defined `stormpath.web.idSite.result.uri` - I forgot that.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
```diff
@@ -150,33 +153,44 @@ public void init(HttpSecurity http) throws Exception {
...
-                    .formLogin()
-                    .loginPage(loginUri)
-                    .defaultSuccessUrl(loginNextUri)
-                    .successHandler(successHandler)
-                    .failureHandler(failureHandler)
-                    .usernameParameter("login")
-                    .passwordParameter("password")
-                    .and().authorizeRequests()
-                    .antMatchers(loginUriMatch).permitAll();
+                        .formLogin()
```
dogeared: Why the extra indentation?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
mrioan: The Assert.notNull does not allow the code to continue if resultListener is null and it also throws an exception to the invoker of this method so they can get to know that this is not a valid invocation. We use the `Assert.notNull` strategy everywhere we know that a null parameter is not allowed rather than just checking for null. The assertion works as a "firewall" that does not allow the code to continue if the method is invoked with unsupported values. This way you catch problems earlier, your algorithm can work safely and the caller is notified (with an exception) that what he is trying to do is not supported by the inner implementation of the method.

extensions/spring/boot/stormpath-spring-security-webmvc-spring-boot-starter/src/main/java/com/stormpath/spring/boot/autoconfigure/StormpathWebSecurityAutoConfiguration.java
mrioan: Not sure a class-level annotation will achieve the same thing. This condition in this particular method is only valid for the IdSite-related bean, the class can be loaded but this particular bean should not if IDSite is not enabled.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
dogeared: I think you may be missing my point: In a non-spring security setup, like Spring WebMVC, it is EXPECTED that the `resultListener` will be null here.

In the `IdSiteResultController`, we always call `addResultListener`. I suppose a null check could be there. 

Which do you think would be better?

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/security/listener/SpringSecurityIdSiteResultListener.java
mrioan: Prior to this PR the responsibility creating Authenticated objects and how and when to populate them with the required Roles and Granted Authorities was completely and solely achieved by the AuthenticationProvider. The `AuthenicationProvider` interface is provided by Spring Security and it is one of the key doors that Spring Security provides to plug your login-specific business logic. It provides a single method `Authentication authenticate(Authentication)` (there's an additional one where you can ask if the `authenticationProvider` supports a particular authenticationToken). Spring Security is designed in a way where you encapsulate login requests in Authentication classes. So there is nothing wrong about having a new `PreAuthenticatedAuthenticationToken ` but actually is the desired approach. Spring provides different tokens: `JaasAuthentication`, `UsernamePasswordAuthentication`, `AnonymousAuthentication`, which are intended to be processed by an `AuthenticationProvider`. In your implementation you are trying to mimic what the StormpathAuthenticationProvider does now so you are basically implementing the same responsibility in two different places. Any future modification in the provider will need to be carried out here as well and it is very likely that we will forget about this method in the Listener.

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
mrioan: Agree that this does not look pretty but Spring Security provides this sole method for the Provider to get an Authentication token in order to do the "logins". Maybe we can think of other pattern but all the tokens need to come to this sole method since this is what Spring Security offers in the `AuthenicationProvier` interface

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/token/PreAuthenticatedAuthenticationToken.java
mrioan: It denotes a kind of `Authentication` (see http://docs.spring.io/spring-security/site/docs/current/apidocs/org/springframework/security/core/Authentication.html)

impl/src/main/java/com/stormpath/sdk/impl/idsite/DefaultIdSiteCallbackHandler.java
mrioan: When packages are imported the Javadoc can read the class, if they have not been imported in the class then you need to the fqcn. When javadoc is compiled they will all be seen in the same way (without FQCN as I did things like:  

{@link IdSiteResultListener#onAuthenticated(com.stormpath.sdk.idsite.AuthenticationResult) IdSiteResultListener#onAuthenticated(AuthenticationResult)}

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
mrioan: The branch for both IDSite and stormpathWebEnabled is below.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
mrioan: Just IntelliJ

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
mrioan: Unless impossible, a class should care about its inside business. If a null listener is not appropriate then we should not allow it to enter here. Can't the invoker check for null itself before invoking this method?

extensions/spring/boot/stormpath-spring-security-webmvc-spring-boot-starter/src/main/java/com/stormpath/spring/boot/autoconfigure/StormpathWebSecurityAutoConfiguration.java
dogeared: yes, that makes sense as we can have Spring Security without id site being enabled

dogeared:
I've addressed all the issues [here](https://github.com/stormpath/stormpath-sdk-java/pull/380) Closing this one now.

Issue #397, opened 2015-12-10
Session saver throws Exception after login 


mrioan:
In our servlet Plugin we get a
`IllegalArgumentException: An authenticated account must be represented with a specific request authType` when `session` is used rather than `cookie` (`stormpath.web.authc.savers = session`)

mrioan:
In order for this fix to work, the following configuration is required:

stormpath.web.authc.savers = session
stormpath.web.account.resolvers = session

lhazlewood:
@mrioan is this a bug?  Or just a misconfiguration?

If the only thing you needed to change was the two specified properties, it sounds like misconfiguration to me and not a bug.  Thoughts?

mrioan:
@lhazlewood It was a bug: missing `request.setProperty` in `SessionAccountResolver.java`

mrioan:
Solved in #402

PR #402, opened 2015-12-14
#397 Setting AuthType = form in request during SessionAccountResolver

mrioan:


mrioan:
@lhazlewood I basically copied what is being done in [CookieAccountResolver](https://github.com/stormpath/stormpath-sdk-java/blob/master/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java#L87)

I understand what you are saying... If this fix is wrong then I assume what is currently being done in `CookieAccountResolver` is also wrong. Right?

mrioan:
Looking at the servlet request API, it appears we are allowed to return an implementation-specific String if desired.  We need to change this to be some other string that accurately reflects how they authenticated, for example "jwt" or "jwt-cookie" or "session".

PR #419, opened 2015-12-21
Fix for providerAccountResult.isNewAccount() always returning true

mrioan:


maranai:
@mrioan this PR is approved.

Issue #425, opened 2015-12-23
LimitTo on accountCriteria do not work


tomekl007:
given : 
>
int limit = 1;
int offset = n;
client.getAccounts(Accounts.criteria().offsetBy(offset).limitTo(limit));

operation returns more than one elements.
If we have 10 accounts, 
setting offset to 3 and limit to 1, return 10 - 3 = 7 records
setting offset to 8 and limit to 1, return 10 - 8 = 2 records,
setting offset to 0 and limit to 1, return 10 - 0 = 10 records.
Property limit is totally ignored

lhazlewood:
What version of the SDK are you using?

tomekl007:
I am using most recent version, that is 1.0.RC7.6

tomekl007:
I've menage to replicate that bug, see my pull request : 
https://github.com/stormpath/stormpath-sdk-java/pull/426

lhazlewood:
Thanks so much for the PR!

tomekl007:
Hi, is there a possibility that this issue will be resolved soon?

mrioan:
Hi @tomekl007 ,

What you are experiencing is not a bug. It is the actual expected behavior. Please take a look at our Javadoc: https://github.com/stormpath/stormpath-sdk-java/blob/master/api/src/main/java/com/stormpath/sdk/query/Criteria.java#L129

The total size of the list is given by the quantity of items matching the criteria like `Accounts.where(Accounts.name().containsIgnoreCase("Foo")`. 

The whole list is divided in pages. `limitTo` sets the quantity of items per page, rather than the size of the whole list. When you iterate over the list, the backend will be contacted every time the page is exhausted and only X quantity of items will be locally available per request (where X is the page size). Paging happens behind the scenes, you are basically given the chance to set the page size so you can tweak the quantity of items that will travel through the wire at once, but besides that, paging is a transparent concept once the list is being traversed.

So, for example, if the query matches 10 elements and each page contains 1 element (as stated by `limitTo(1)`) then you will be contacting the backend 11 times: 1 for each page + 1 to know that the list has finally been completely traversed. However if you specify `limitTo(25)` then the backend will be contacted only once since the underlying page retrieve and contain all the 10 items that the criteria matched.

BTW, you can always invoke `list.getSize()` to know the actual quantity of items matching the criteria that the whole list will contain.

If you agree with what I am explaining here please let us know so we can close this issue.

Thanks!

[...]

PR #426, opened 2015-12-23
test that shows bug in limit method

tomekl007:


dogeared:
Issue #425 was closed as it addressed the correct usage of limit in Stormpath.

Issue #433, opened 2015-12-29
AccountStore is null when ApplicationAccountStoreMapping is an Organization


dogeared:
Setup an `Application` with only `Organization`s as its `AccountStore`s.

In the code below, `accountStore` is `null`

```
ApplicationAccountStoreMappingList accountStoreMappings = application.getAccountStoreMappings();
for (ApplicationAccountStoreMapping mapping : accountStoreMappings) {
    AccountStore accountStore = mapping.getAccountStore();
    if (!(accountStore instanceof Organization)) { continue; }

    Organization organization = (Organization) accountStore;
    System.out.println(organization.getName());
}
```

maranai:
Pull request ready for review: https://github.com/stormpath/stormpath-sdk-java/pull/441

dogeared:
closes #441

PR #441, opened 2016-01-07
#433 AccountStore is null when ApplicationAccountStoreMapping is an Organization

maranai:


impl/src/main/java/com/stormpath/sdk/impl/application/AbstractAccountStoreMapping.java
```diff
@@ -54,6 +55,8 @@ public AccountStore getAccountStore() {
             accountStore = getDataStore().getResource(href, Directory.class);
         } else if (href.contains("groups")) {
             accountStore = getDataStore().getResource(href, Group.class);
+        } else if (href.contains("organizations")){
```
mrioan: please do it this way: `href.contains("/organizations/")` (note the slash at the beginning and end)

Can you also please do:  `href.contains("/groups/")` and `href.contains("/directories/")` ?

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
```diff
@@ -426,6 +428,38 @@ class ApplicationIT extends ClientIT {
     }
 
     /**
+     * @since 1.0.RC9
```
mrioan: 1.0.RC8.2 please

mrioan:
@maranai Review ready. Back to you. Thanks!

Issue #164, opened 2015-03-26
Cache does not return correctly resources with custom data or expansions


maranai:
Resources' custom data as well as properties that have been expanded are not returned correctly from Cache. If a expanded resource is requested and it exists in the cache, the cache will return it with no expansions.

mrioan:
When working on this pleas read the comment of `DirectoryIT#testDirectoryExpansion`

mrioan:
@maranai could you please create a test that fails because of the custom data issue you are describing? I have not been able to reproduce it. I did, however, found the issue with expansions.

maranai:
@mrioan I've created a test to demonstrate the problem with the cache and expanded resources.
Since we created a workaround for this in the past, the WriteCacheFilter does not cache any expanded resources so for the test to fail as expected, you may need to comment out or remove the line starting with (!request.getUri().hasQuery() || !request.getUri().getQuery().containsKey("expand") ..... in WriteCacheFilter.java and enable the test.
I'm not sure if the implemented workaround will end up being the actual solution (or if we will ever want to cache expanded resources) but the test clarifies one of the scenarios where expansions and actual cache behavior cause problems.

dogeared:
branch 164_cache_expansions_tests can be used as a starting point to address this issue. PR #454 was closed as the branch was not yet ready to be merged.

jarias:
Looks like there was a problem with Anai's test and the code is actually working fine, I'll update the test and remove the code that prevents writing entities with expanded attributes into the cache.

However the expanded entity and the not expanded entity are store with different keys, at least for me it should be the same key so future retrieves of that entity would have the expanded attributes.

Current Workflow:

1. Get an app, it gets store in the cached with its href
2. Get the app again expanding something (accounts for example), it gets cached with the app herf + the query string expand=something
3. Get the app again and it would be the one from step 1 without the expanded attribute

At least for me again I would like to have the app in step 3 the one with the expanded accounts collection.

@mrioan @lhazlewood @dogeared @omgitstom what do you think?

PR #454, opened 2016-01-22
 #164 Cache does not return correctly resources with custom data or e…

maranai:
…xpansions


dogeared:
Resolves #164

lhazlewood:
@dogeared another closed issue with an orphaned branch

dogeared:
Issue #164 now references the 164_cache_expansions_tests branch. The branch has a test that is set not to execute (enabled=false) and the PR should probably not have been opened. The branch can still be used as a starting point to address #164.

PR #455, opened 2016-01-22
Refactored to ensure that Spring Security WebMVC works as well as Spring Boot Spring Security WebMVC

dogeared:
Also, updated new example to be clear that it's a spring security webmvc bare bones example.


extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
```diff
@@ -161,16 +164,12 @@ public void init(HttpSecurity http) throws Exception {
 
         // autowire this bean
         ApplicationContext context = http.getSharedObject(ApplicationContext.class);
-        AutowireCapableBeanFactory beanFactory = context.getAutowireCapableBeanFactory();
-        beanFactory.autowireBean(this);
+        context.getAutowireCapableBeanFactory().autowireBean(this);
```
mrioan: @dogeared Is this line really needed? We hadn't had the need to do something like this in the past...

mrioan:
@dogeared back to you. Just a single comment...

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathWebSecurityConfigurer.java
dogeared: Actually, I am not sure. This change puts it back to the way it was. I will see what happens if I remove that line and the one before it, but I think you had it there for a reason. ;)

dogeared: @mrioan - I just confirmed that removing those two lines breaks Spring Security.

I am very curious now to know why this is and how you discovered it originally!

mrioan: @dogeared ah! I did not realize it was old code that was getting back in. I thought it was a new line of code. I think I recall that Rob explained to us that this was needed when the `WebSecurityConfigurerAdapter` is not used (when we switched for it to the `http.apply(stormpath())` strategy

PR #460, opened 2016-01-30
Abstracted out boilerplate setup code for register controller

dogeared:
for use in customized versions.

NOTE: This PR is not intended to be merged - it is a POC


dogeared:
This refactor enables easy custom configurations without all the boilerplate.

Let's say you have a customized registration controller called: `CustomRegisterController`.

And, you have a Spring Boot config file to override the `stormpathRegisterController` bean.

With this refactor, here's what `CustomRegisterControllerConfig` would look like:

```
@Configuration
public class CustomRegisterControllerConfig extends AbstractStormpathWebMvcConfiguration {

    @Bean
    public Controller stormpathRegisterController() {
        return setupRegisterController(new CustomRegisterController());
    }

}
```

If we like this pattern, we can replicate for LoginController and all the other ones.

Without the change in this PR, you'd need the following to accomplish the same overridden bean:

```
@Configuration
public class CustomRegisterControllerConfig extends AbstractStormpathWebMvcConfiguration {

    @Bean
    public Controller stormpathRegisterController() {
        if (idSiteEnabled) {
            return createIdSiteController(idSiteRegisterUri);
        }

        //otherwise standard registration:
        CustomRegisterController controller = new CustomRegisterController();
        controller.setCsrfTokenManager(stormpathCsrfTokenManager());
        controller.setClient(client);
        controller.setEventPublisher(stormpathRequestEventPublisher());
        controller.setFormFields(toDefaultFields(stormpathRegisterFormFields()));
        controller.setLocaleResolver(stormpathLocaleResolver());
        controller.setMessageSource(stormpathMessageSource());
        controller.setAuthenticationResultSaver(stormpathAuthenticationResultSaver());
        controller.setUri(registerUri);
        controller.setView(registerView);
        controller.setNextUri(registerNextUri);
        controller.setLoginUri(loginUri);
        controller.setVerifyViewName(verifyView);
        controller.init();

        SpringController springController = new SpringController(controller);
        if (urlPathHelper != null) {
            springController.setUrlPathHelper(urlPathHelper);
        }
        return springController;
    }

    private List<DefaultField> toDefaultFields(List<Field> fields) {
        List<DefaultField> defaultFields = new ArrayList<DefaultField>(fields.size());
        for (Field field : fields) {
            Assert.isInstanceOf(DefaultField.class, field);
            defaultFields.add((DefaultField) field);
        }

        return defaultFields;
    }
}
```

lhazlewood:
I think Spring Boot / Security allows similar overrides/callback method concepts - we should use the same pattern/naming that they do.

extensions/spring/stormpath-spring-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebMvcConfiguration.java
```diff
@@ -969,13 +969,14 @@ public String getMessage(String key, Locale locale, Object... args) {
     }
 
     public Controller stormpathRegisterController() {
+        return setupRegisterController(new RegisterController());
+    }
 
+    public Controller setupRegisterController(RegisterController controller) {
```
mzumbadoate: since this is a new method and it looks like it is called just from stormpathRegisterController should this be public? private maybe?

dogeared: Thanks for the feedback @mzumbadoate. This is likely *not* the approach we will be taking to solve this issue. It was just an experiment. As is, the method could be protected as it's used in a sub-class, but it wouldn't work for it to be private.

lhazlewood:
@dogeared are you going to delete the now orphaned branch as well?

dogeared:
@lhazlewood - yes, I've deleted the branch. Going forward, rather than creating a PR for a question or something experimental, I will just send the compare link in an email.

PR #461, opened 2016-01-31
Get rid of specific size tests in order to enable travis concurrency.

dogeared:


dogeared:
Will revisit when a concurrency solution is found.

PR #465, opened 2016-02-02
Added permitAll for accessTokenUri for Spring Security

dogeared:


dogeared:
@mrioan - without this addition, you get redirected to `/login` when you try to hit `/oauth/token` with Spring Security enabled.

PR #436, opened 2016-01-01
 #292 Exchanging ID Site token for a Stormpath Access Token

dogeared:
Please refer to #376 for previous comments

PR #467, opened 2016-02-03
Renamed `Oauth2Requests.IDSITE_AUTHENTICATION_REQUEST` to `Oauth2Requests.ID_SITE_AUTHENTICATION_REQUEST`

kaarelnu:
Thank you very much for #436 but when I started using it, I noticed a slight imperfection:

```scala
val request = Oauth2Requests.IDSITE_AUTHENTICATION_REQUEST.builder()
  .setToken(jwt)
  .build()

val authenticationResult = Authenticators.ID_SITE_AUTHENTICATOR
  .forApplication(application)
  .authenticate(request)
```

Would be nice if there was a consistency in naming. I suggest `Oauth2Requests.ID_SITE_AUTHENTICATION_REQUEST` and `Authenticators.ID_SITE_AUTHENTICATOR`.


dogeared:
Hey @kaarelnu  Thanks very much for this. What we will do is support both for a time and deprecate the inconsistent version for removal at a later release.

This is because there may be people already using the code as it is and we don't want to force a change on them immediately.

I've set this up as issue #468 and you can track its progress.

kaarelnu:
Thanks, @dogeared!

PR #471, opened 2016-02-04
Updated code to differentiate between AccessToken and RefreshToken.

dogeared:
Previously, RefreshToken inherited from AccessToken. Per the oauth2 spec, this is incorrect.

This PR moves common implementation into a base class and provides a method for ensuring that an AccessToken is really an AccessToken.

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
@@ -15,12 +15,6 @@
 */
 package com.stormpath.sdk.oauth;
 
-import com.stormpath.sdk.account.Account;
-import com.stormpath.sdk.application.Application;
-import com.stormpath.sdk.resource.Deletable;
-import com.stormpath.sdk.resource.Resource;
-import com.stormpath.sdk.tenant.Tenant;
-
 /**
  * This interface represents an Authentication OAuth2 token created in Stormpath.
  * Specifically, an
--- a/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
@@ -22,13 +22,13 @@ import com.stormpath.sdk.resource.Resource;
 import com.stormpath.sdk.tenant.Tenant;
 
 /**
- * This class represents an Authentication OAuth2 token created in Stormpath.
+ * This interface represents an Authentication OAuth2 token created in Stormpath.
  * Specifically, an
  * <a href="https://docs.stormpath.com/guides/token-management/#using-stormpath-to-generate-an-oauth-20-access-tokens">AccessToken</a>
  *
  * @since 1.0.RC7
  */
 public interface AccessToken extends BaseOauth2Token {
-    AccessToken ensureAccessToken();
+
 }
 
--- a/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
@@ -23,37 +23,12 @@ import com.stormpath.sdk.tenant.Tenant;
 
 /**
  * This class represents an Authentication OAuth2 token created in Stormpath.
+ * Specifically, an
+ * <a href="https://docs.stormpath.com/guides/token-management/#using-stormpath-to-generate-an-oauth-20-access-tokens">AccessToken</a>
  *
  * @since 1.0.RC7
  */
-public interface AccessToken extends Resource, Deletable {
-
-    /**
-     * Returns the String representation of the Json Web Token.
-     *
-     * @return a String value denoting a JWT.
-     */
-    String getJwt();
-
-    /**
-     * Returns the {@link Account} associated to this {@link AccessToken}.
-     *
-     * @return the {@link Account} associated to this {@link AccessToken}.
-     */
-    Account getAccount();
-
-    /**
-     * Returns the {@link Application} associated to this {@link AccessToken}.
-     *
-     * @return the {@link Application} associated to this {@link AccessToken}.
-     */
-    Application getApplication();
-
-    /**
-     * Returns the {@link Tenant} this {@link AccessToken} belongs to.
-     *
-     * @return the {@link Tenant} this {@link AccessToken} belongs to.
-     */
-    Tenant getTenant();
+public interface AccessToken extends BaseOauth2Token {
+    AccessToken ensureAccessToken();
 }
 
--- a/api/src/main/java/com/stormpath/sdk/oauth/BaseOauth2Token.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/BaseOauth2Token.java
@@ -31,6 +31,7 @@ import com.stormpath.sdk.tenant.Tenant;
  * @since 1.0.RC8.3
  */
 public interface BaseOauth2Token extends Resource, Deletable {
+
     /**
      * Returns the String representation of the Json Web Token.
      *
@@ -58,4 +59,5 @@ public interface BaseOauth2Token extends Resource, Deletable {
      * @return the {@link Tenant} this {@link AccessToken} belongs to.
      */
     Tenant getTenant();
+
 }
--- a/api/src/main/java/com/stormpath/sdk/oauth/BaseOauth2Token.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/BaseOauth2Token.java
@@ -1,0 +1,61 @@
+/*
+* Copyright 2015 Stormpath, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+* http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package com.stormpath.sdk.oauth;
+
+import com.stormpath.sdk.account.Account;
+import com.stormpath.sdk.application.Application;
+import com.stormpath.sdk.resource.Deletable;
+import com.stormpath.sdk.resource.Resource;
+import com.stormpath.sdk.tenant.Tenant;
+
+/**
+ * This is a base interface for those methods that Stormpath
+ * {@link AccessToken}s and {@link RefreshToken}s have in common.
+ *
+ * @see AccessToken
+ * @see RefreshToken
+ *
+ * @since 1.0.RC8.3
+ */
+public interface BaseOauth2Token extends Resource, Deletable {
+    /**
+     * Returns the String representation of the Json Web Token.
+     *
+     * @return a String value denoting a JWT.
+     */
+    String getJwt();
+
+    /**
+     * Returns the {@link Account} associated to this {@link AccessToken}.
+     *
+     * @return the {@link Account} associated to this {@link AccessToken}.
+     */
+    Account getAccount();
+
+    /**
+     * Returns the {@link Application} associated to this {@link AccessToken}.
+     *
+     * @return the {@link Application} associated to this {@link AccessToken}.
+     */
+    Application getApplication();
+
+    /**
+     * Returns the {@link Tenant} this {@link AccessToken} belongs to.
+     *
+     * @return the {@link Tenant} this {@link AccessToken} belongs to.
+     */
+    Tenant getTenant();
+}
--- a/api/src/main/java/com/stormpath/sdk/oauth/RefreshToken.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/RefreshToken.java
@@ -16,9 +16,12 @@
 package com.stormpath.sdk.oauth;
 
 /**
- * This interfaces denotes a specific kind of {@link AccessToken}: a
- * <a href="https://docs.stormpath.com/guides/token-management/#refreshing-access-tokens">Refresh Token</a>.
+ * This class represents an Authentication OAuth2 token created in Stormpath.
+ * Specifically, a
+ * <a href="https://docs.stormpath.com/guides/token-management/#refreshing-access-tokens">Refresh Token</a>
  *
+ * @since 1.0.RC7
  */
-public interface RefreshToken extends AccessToken {
+public interface RefreshToken extends BaseOauth2Token {
+
 }
--- a/api/src/main/java/com/stormpath/sdk/oauth/RefreshToken.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/RefreshToken.java
@@ -16,7 +16,7 @@
 package com.stormpath.sdk.oauth;
 
 /**
- * This class represents an Authentication OAuth2 token created in Stormpath.
+ * This interface represents an Authentication OAuth2 token created in Stormpath.
  * Specifically, a
  * <a href="https://docs.stormpath.com/guides/token-management/#refreshing-access-tokens">Refresh Token</a>
  *
--- a/extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
+++ b/extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
@@ -1504,6 +1504,63 @@ class ApplicationIT extends ClientIT {
         assertNotNull Authenticators.JWT_AUTHENTICATOR.forApplication(app).withLocalValidation().authenticate(authRequest)
     }
 
+    @Test
+    void testAttemptAuthenticationWithRefreshToken() {
+        def app = createTempApp()
+        def account = createTestAccount(app)
+
+        PasswordGrantRequest grantRequest = Oauth2Requests.PASSWORD_GRANT_REQUEST.builder()
+            .setLogin(account.email)
+            .setPassword("Changeme1!")
+            .build()
+
+        def grantResult = Authenticators.PASSWORD_GRANT_AUTHENTICATOR
+            .forApplication(app)
+            .authenticate(grantRequest)
+
+        // Authenticate token against Stormpath using refresh token <--- INVALID
+        JwtAuthenticationRequest authRequest = Oauth2Requests.JWT_AUTHENTICATION_REQUEST.builder()
+            .setJwt(grantResult.getRefreshTokenString())
+            .build()
+
+        try {
+            Authenticators.JWT_AUTHENTICATOR.forApplication(app).authenticate(authRequest)
+            fail("Should have thrown")
+        } catch (Exception e) {
+            def message = e.getMessage()
+            assertTrue message.equals("JWT failed validation; it cannot be trusted.")
+        }
+    }
+
+    @Test
+    void testAttemptAuthenticationWithRefreshTokenWithLocalValidation() {
+        def app = createTempApp()
+        def account = createTestAccount(app)
+
+        PasswordGrantRequest grantRequest = Oauth2Requests.PASSWORD_GRANT_REQUEST.builder()
+                .setLogin(account.email)
+                .setPassword("Changeme1!")
+                .build()
+
+        def grantResult = Authenticators.PASSWORD_GRANT_AUTHENTICATOR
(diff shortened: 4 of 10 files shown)


api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
```diff
@@ -23,37 +23,12 @@
...
-    Application getApplication();
-
-    /**
-     * Returns the {@link Tenant} this {@link AccessToken} belongs to.
-     *
-     * @return the {@link Tenant} this {@link AccessToken} belongs to.
-     */
-    Tenant getTenant();
+public interface AccessToken extends BaseOauth2Token {
+    AccessToken ensureAccessToken();
```
mrioan: The interfaces for Resources must always reflect what the resource actually provides in the backend. As an example, note that his interface only has getters but not setters. `AccessToken` cannot be created/modify by the developers, so we do not provide anything other than what the backend provides. We can sometimes have convenience methods to help the users do something but they must always be clearly stated that is is a convenience method (see Application#createAccount(Account) for example).

 Additionally, any class/operation that is part of the API module must always have detailed documentation. We can overlook documentation in other modules, but API must have it.

 Additionally, we should adhere to coding standards, in this case, a blank line is need before line 32 and after line 32.

You can put helper methods/functions in the concrete implementation (without it been exposed in the interface) an flagged as protected/private when possible. We only expose (exposing means having them in the API) helper operations if that is strictly necessary and will help customers not to write complex pieces of code.

api/src/main/java/com/stormpath/sdk/oauth/BaseOauth2Token.java
```diff
@@ -0,0 +1,61 @@
...
+ * This is a base interface for those methods that Stormpath
+ * {@link AccessToken}s and {@link RefreshToken}s have in common.
+ *
+ * @see AccessToken
+ * @see RefreshToken
+ *
+ * @since 1.0.RC8.3
+ */
+public interface BaseOauth2Token extends Resource, Deletable {
+    /**
```
mrioan: Line break here

api/src/main/java/com/stormpath/sdk/oauth/RefreshToken.java
```diff
@@ -16,9 +16,12 @@
 package com.stormpath.sdk.oauth;
 
 /**
- * This interfaces denotes a specific kind of {@link AccessToken}: a
- * <a href="https://docs.stormpath.com/guides/token-management/#refreshing-access-tokens">Refresh Token</a>.
+ * This class represents an Authentication OAuth2 token created in Stormpath.
```
mrioan: This interface represents...

api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
```diff
@@ -23,37 +23,12 @@
 
 /**
  * This class represents an Authentication OAuth2 token created in Stormpath.
```
mrioan: This interface represents...

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
```diff
@@ -1504,6 +1504,63 @@ class ApplicationIT extends ClientIT {
         assertNotNull Authenticators.JWT_AUTHENTICATOR.forApplication(app).withLocalValidation().authenticate(authRequest)
     }
 
+    @Test
```
mrioan: /** @since 1.0.RC8.3 */

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
```diff
@@ -1504,6 +1504,63 @@ class ApplicationIT extends ClientIT {
...
+        try {
+            Authenticators.JWT_AUTHENTICATOR.forApplication(app).authenticate(authRequest)
+            fail("Should have thrown")
+        } catch (Exception e) {
+            def message = e.getMessage()
+            assertTrue message.equals("JWT failed validation; it cannot be trusted.")
+        }
+    }
+
+    @Test
```
mrioan: /** @since 1.0.RC8.3 */

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -53,37 +38,19 @@ public DefaultAccessToken(InternalDataStore dataStore, Map<String, Object> prope
...
-
-    @Override
-    public Tenant getTenant() {
-        return getResourceProperty(TENANT);
-    }
-
     @Override
-    public void delete() {
-        getDataStore().delete(this);
+    public AccessToken ensureAccessToken() {
```
mrioan: Not clear to me. If this instance is already instantiated then this Object cannot be other thing than a Access Token. If the underlying logic is creating an `AccessToken` when the JWT is not really an access token then that is a bug. This class should not have to check anything. It should always return `true` if it gets asked "are you an access token"? If there is a chance that this code ever returns false then this is a bug. I think this method must be removed. What I would do (if necessary) is to have the the constructor asserting this JWT, that way this object will be never created unless the jwt is really an access token

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultBaseOauth2Token.java
```diff
@@ -0,0 +1,76 @@
...
+import com.stormpath.sdk.impl.resource.ResourceReference;
+import com.stormpath.sdk.impl.resource.StringProperty;
+import com.stormpath.sdk.oauth.BaseOauth2Token;
+import com.stormpath.sdk.tenant.Tenant;
+
+import java.util.Date;
+import java.util.Map;
+
+public class DefaultBaseOauth2Token extends AbstractInstanceResource implements BaseOauth2Token {
+    static final String ACCOUNT_PROP_NAME = "account";
```
mrioan: Line break

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultJwtAuthenticator.java
```diff
@@ -81,7 +84,7 @@ public JwtAuthenticationResult authenticate(Oauth2AuthenticationRequest authenti
                 properties.put(DefaultAccessToken.JWT_PROP_NAME, jwtRequest.getJwt());
                 properties.put(DefaultAccessToken.TENANT_PROP_NAME, application.getTenant());
 
-                AccessToken accessToken = new DefaultAccessToken(dataStore, properties);
+                AccessToken accessToken = new DefaultAccessToken(dataStore, properties).ensureAccessToken();
```
mrioan: See line 76. This is explicitly adding the "/accessTokens/" string to the HREF. This is by no doubts an access token. Therefore, the `.ensureAccessToken()` is not needed. If there is a chance that the JWT is not an access token then all this piece of code is buggy as it is constructing a wrong object.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultJwtAuthenticator.java
```diff
@@ -94,7 +97,9 @@ public JwtAuthenticationResult authenticate(Oauth2AuthenticationRequest authenti
         StringBuilder stringBuilder = new StringBuilder(application.getHref());
         stringBuilder.append(OAUTH_TOKEN_PATH);
         stringBuilder.append(jwtRequest.getJwt());
-        AccessToken accessToken = dataStore.getResource(stringBuilder.toString(), AccessToken.class);
+        AccessToken accessToken =
+            dataStore.getResource(stringBuilder.toString(), AccessToken.class).ensureAccessToken();
```
mrioan: here, the dataStore is returning what we are asking for: an AccessToken. If there is a chance that the DataStore is constructing a wrong object then the DataStore has a bug and it should be there, not here trying to validate that the datastore did what it was asked to do.

mrioan:
@dogeared back to you

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
dogeared: I originally had ensureAccessToken called from the constructor. I was not sure that was best practice *and* it added a lot of complexity to some of the unit tests. If you think it is worth it, I will go back to that.

I think since it is no longer exposed in the api (per your comments above), that it is ok the way it is. I will take your feedback on this.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultJwtAuthenticator.java
dogeared: As it was, you *could* pass in a refresh token and use it for authentication. line 76 did not guard against that for local validation.

I just confirmed this in the IT tests I added to ApplicationIT

dogeared:
@mrioan - thanks! back to you...

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultJwtAuthenticator.java
mrioan: Exactly, that is the whole purpose of this PR. I am not saying that there is a problem. I am saying that the way that this problem is being solved is not good enough. When we have an Account we know it is an account (there is no a single chance it is a group), when we have an application we are sure we have an application (there is no chance it is a CustomData). We cannot basically be creating objects and not be confident that they are what they should be. If we keep the code as you suggest, every time we manipulate an Access Token aanywhre in the code we first need to validate that is is acatullay true. As you can see this is really error prom. We have a bug, our ITs failed to catch it but your propose solution is not fixing the bug, it is trying to avoid being hit by the bug. The real soltuion is to be 100% sure that when she have an AccessToken instantiated it is really an Access Token.

dogeared: ok. Let me bring `ensureAccessToken` back into the constructor. I'll take another pass at it.

api/src/main/java/com/stormpath/sdk/oauth/AccessToken.java
```diff
@@ -22,38 +22,13 @@
 import com.stormpath.sdk.tenant.Tenant;
```
mrioan: All these imports are not required any longer

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultBaseOauth2Token.java
```diff
@@ -0,0 +1,77 @@
...
+import com.stormpath.sdk.impl.resource.Property;
+import com.stormpath.sdk.impl.resource.ResourceReference;
+import com.stormpath.sdk.impl.resource.StringProperty;
+import com.stormpath.sdk.oauth.BaseOauth2Token;
+import com.stormpath.sdk.tenant.Tenant;
+
+import java.util.Date;
+import java.util.Map;
+
+public class DefaultBaseOauth2Token extends AbstractInstanceResource implements BaseOauth2Token {
```
mrioan: This should be an abstract class. We do not want this class to ever be instantiated. This class only provides common operations for its concrete children. Note that the name of the class must also be `Abstract*` and not `Default*`
```
public abstract class AbstractBaseOauth2Token extends AbstractInstanceResource implements BaseOauth2Token {
```

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -15,75 +15,37 @@
...
-    @Override
-    public Application getApplication() {
-        return getResourceProperty(APPLICATION);
-    }
-
-    @Override
-    public Tenant getTenant() {
-        return getResourceProperty(TENANT);
-    }
+    private void ensureAccessToken() {
```
mrioan: Please add a very short note saying what this method does and that it will throw if the jwt does not belong to a `access_token`

dogeared: Isn't that what line 44 does?

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
```diff
@@ -41,47 +45,47 @@ import static org.testng.Assert.*
 class DefaultAccessTokenTest {
 
     @Test
-    void testGetPropertyDescriptors() {
```
mrioan: Please do not delete this test. Every resource is always validated to check that it contains the proper quantity and type of internal properties. As a rule of thumb we should not delete tests unless they are deprecated for some reason. If you are having issues with a test then you could temporarily ignore it (`@Test(enabled = false)`) and add a comment about why you are disabling it.

This is a working version for this deleted test:


```@Test
    void testGetPropertyDescriptors() {
        def secret = "a_very_secret_key"
        def href = "https://api.stormpath.com/v1/accessTokens/5hFj6FUwNb28OQrp93phPP"

        String jwt = Jwts.builder()
                .setSubject(href)
                .claim("rti", "abcdefg")
                .signWith(SignatureAlgorithm.HS256, secret.getBytes("UTF-8"))
                .compact();

        def properties = [
                href: href,
                jwt: jwt,
                tenant: [href: "https://api.stormpath.com/v1/tenants/jdhrgojeorigjj09etiij"],
                application: [href: "https://api.stormpath.com/v1/applications/928glsjeorigjj09etiij"],
                account: [href: "https://api.stormpath.com/v1/accounts/apsd98f2kj09etiij"],
                created_at: "2015-01-01T00:00:00Z"
        ]

        def internalDataStore = createStrictMock(InternalDataStore)
        def apiKey = createStrictMock(ApiKey)

        expect(internalDataStore.getApiKey()).andReturn(apiKey)
        expect(apiKey.getSecret()).andReturn(secret)

        replay internalDataStore, apiKey

        def defaultAccessToken = new DefaultAccessToken(internalDataStore, properties)

        def propertyDescriptors = defaultAccessToken.getPropertyDescriptors()

        assertEquals(propertyDescriptors.size(), 5)

        assertTrue(propertyDescriptors.get("jwt") instanceof StringProperty)
        assertTrue(propertyDescriptors.get("created_at") instanceof DateProperty)

        assertTrue(propertyDescriptors.get("tenant") instanceof ResourceReference && propertyDescriptors.get("tenant").getType().equals(Tenant))
        assertTrue(propertyDescriptors.get("application") instanceof ResourceReference && propertyDescriptors.get("application").getType().equals(Application))
        assertTrue(propertyDescriptors.get("account") instanceof ResourceReference && propertyDescriptors.get("account").getType().equals(Account))


        verify internalDataStore, apiKey
    }
```

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
```diff
@@ -92,4 +96,68 @@ class DefaultAccessTokenTest {
...
+        try {
+            new DefaultAccessToken(internalDataStore, properties)
+            fail("should have thrown")
+        } catch (Exception e) {
+            def message = e.getMessage()
+            assertTrue message.equals("JWT failed validation; it cannot be trusted.")
+        }
+    }
+
+    @Test
```
mrioan: /** @since 1.0.RC8.3 */

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
```diff
@@ -92,4 +96,68 @@ class DefaultAccessTokenTest {
         def application = defaultAccessToken.getApplication()
         assertTrue(application instanceof Application && application.getHref().equals(properties.application.href))
     }
-}
\ No newline at end of file
+
+    @Test
```
mrioan: /** @since 1.0.RC8.3 */

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
mrioan: Yes, but I think it does not hurt to add a comment outside (similar to javadoc, no real need to be a real javadoc as it is not in the API module). Someone seeing this method should be able to infer what it is about without having to read the internal code. I think something like is enough:

```
/**
* This method will validate that the received jwt corresponds to an `access_token` (as opposite to a `refresh_token`). If that is not the case then this operation will throw an IllegalArgumentException.
* @since 1.0.RC8.3
*/
```

mrioan:
@dogeared back to you

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
dogeared: The original test exercise instantiating with no properties. That is invalid now as the properties *must* be present at construction time in order to validate the JWT.

The constructor that does not take properties has been removed.

mrioan: Yeap I realized about that, but I provided a working test that (using the new constructor) validates the same things it was validating before

dogeared: Gotcha. I will add it back in.

mrioan: Maybe you have a misunderstanding of what I mean with "deprecated" in my sentence:
"As a rule of thumb we should not delete tests unless they are deprecated for some reason". You may be considering a test as deprecated because the resources being tested has changed. That is wrong, this does not mean that the test is deprecated, this means the test is broken. The test was yet valid but it was failing to be compliant with a new constructor for the resource being tested. Therefore the test is still needed as it was testing important things and must be improved to be compliant with the new resource being tested.

dogeared:
@mrioan - I believe I've addressed everything. Great feedback. Thanks!

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -15,75 +15,45 @@
...
+    }
+
+    /**
+     * This method will validate that the received jwt corresponds to an `access_token` (as opposed to a
+     * `refresh_token`). If that is not the case then this operation will throw a JwtException. It is called
+     * from the constructor, so an AccessToken cannot be instantiated without verifying the jwt is a proper
+     * `access_token`
+     * 
+     * @since 1.0.RC8.3
+     */
```
mrioan: nice

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
```diff
@@ -40,10 +44,34 @@ import static org.testng.Assert.*
  */
 class DefaultAccessTokenTest {
 
-    @Test
```
mrioan: This line was mistakenly erased

impl/src/main/java/com/stormpath/sdk/impl/oauth/AbstractBaseOauth2Token.java
```diff
@@ -0,0 +1,77 @@
...
+import com.stormpath.sdk.impl.resource.Property;
+import com.stormpath.sdk.impl.resource.ResourceReference;
+import com.stormpath.sdk.impl.resource.StringProperty;
+import com.stormpath.sdk.oauth.BaseOauth2Token;
+import com.stormpath.sdk.tenant.Tenant;
+
+import java.util.Date;
+import java.util.Map;
+
+public abstract class AbstractBaseOauth2Token extends AbstractInstanceResource implements BaseOauth2Token {
```
mrioan: /** @since 1.0.RC8.3 */

impl/src/test/groovy/com/stormpath/sdk/impl/oauth/DefaultAccessTokenTest.groovy
dogeared: ooo! good catch

mrioan:
@dogeared back to you. You can consider this PR approved if you agree with all the proposed changes, Thanks! Good work!

PR #459, opened 2016-01-28
Resend password verification

dogeared:
This implements #442 as outlined in https://github.com/stormpath/stormpath-framework-spec/blob/master/email-verification.md

It is implemented in Servlet, Spring and Spring Security

PR #472, opened 2016-02-05
Resend account verification email for webmvc

dogeared:


dogeared:
All of the previous commentary on this pr can be found in #459

PR #494, opened 2016-02-25
Avoid 2 api calls revoke tokens

### LINKED ISSUE #479: Avoid TokenRevocationRequestEventListener to call the backend twice
When tokens are been revoked (during logout) the access and refresh tokens are first retrieved and then deleted. We should be able to delete them without retrieving them first.

mrioan:
Resolves #479 

We are calling the backend twice per token in order to revoke it: 1) get the resource, 2) delete it.

This PR instantiates the Resource without contacting the backend and then issues the delete operation which does contact the backend.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/event/TokenRevocationRequestEventListener.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/event/TokenRevocationRequestEventListener.java
@@ -1,21 +1,22 @@
 /*
- * Copyright 2016 Stormpath, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+* Copyright 2016 Stormpath, Inc.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
 package com.stormpath.sdk.servlet.event;
 
 import com.stormpath.sdk.client.Client;
+import com.stormpath.sdk.impl.ds.DefaultDataStore;
 import com.stormpath.sdk.oauth.AccessToken;
 import com.stormpath.sdk.oauth.RefreshToken;
 import com.stormpath.sdk.resource.ResourceException;
@@ -35,10 +36,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.security.Key;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 /**
- * @since 1.0.RC8.3
- */
+* @since 1.0.RC8.3
+*/
 public class TokenRevocationRequestEventListener implements RequestEventListener {
 
     private static final Logger log = LoggerFactory.getLogger(TokenRevocationRequestEventListener.class);
@@ -46,8 +49,6 @@ public class TokenRevocationRequestEventListener implements RequestEventListener
     private final TokenExtractor tokenExtractor = new BearerHeaderTokenExtractor();
 
     private final JwtTokenSigningKeyResolver jwtTokenSigningKeyResolver = new JwtTokenSigningKeyResolver();
-
-    private final SignatureAlgorithm alg = SignatureAlgorithm.HS256;
 
     private Client client = null;
 
@@ -76,7 +77,7 @@ public class TokenRevocationRequestEventListener implements RequestEventListener
         String jwt = tokenExtractor.getAccessToken(e.getRequest());
         if (jwt != null) {
             if (this.client == null) {
-                this.client = ClientResolver.INSTANCE.getClient(e.getRequest());
+                this.client = ClientResolver.INSTANCE.getClient(e.getRequest()); //will throw if not found
             }
 
             Key signingKey = jwtTokenSigningKeyResolver.getSigningKey(e.getRequest(), e.getResponse(), null, SignatureAlgorithm.HS256);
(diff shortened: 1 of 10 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/event/TokenRevocationRequestEventListener.java
```diff
@@ -16,6 +16,7 @@
 package com.stormpath.sdk.servlet.event;
 
 import com.stormpath.sdk.client.Client;
+import com.stormpath.sdk.impl.ds.DefaultDataStore;
```
lhazlewood: You can't have direct access to this class from the servlet module - it is an SDK internal implementation component only.

dogeared:
FYI - closes #479

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/event/TokenRevocationRequestEventListener.java
mrioan: @lhazlewood I did that in order to avoid exposing the newly added operation in the API.
```
* **Only intended for internal use, not exposed in the API**
     * <p>Convenience method that returns a new resource instance without contacting the backend. This operation allows
     * the <code>href</code> to be a fragment, where the <code>baseUrl</code> can be missing and will be added automatically.<p/>
     * @param hrefFragment when <code>true</code>, the baseUrl will be appended to the value found in the href key of the properties map.
     *                     If <code>false</code> the href will not be altered and will be kept as-is.
     * @return a resource instance corresponding to the specified clazz.
     * @version 1.0.RC9
     */
    public <T extends Resource> T instantiate(Class<T> clazz, Map<String, Object> properties, QueryString qs, boolean hrefFragment) {
        if (hrefFragment) {
            Assert.hasText((String) properties.get("href"), "when hrefFragment is set to true the properties map must contain an href key.");
            String hrefValue = (String) properties.get("href");
            hrefValue = qualify(hrefValue);
            properties.put("href", hrefValue);
        }
        return this.instantiate(clazz, properties, qs);
    }
```

The DataStore is not really visible by customers so I guess it will not hurt to do add this operation to the API anyways. So I will follow your advise.

BTW, we always have a trade-off between adding new (tweak) operations intended for internal use only (which end up requiring a cast to the implementation class) vs exposing that new operation in the API. Since we are the maintainers of the every sub-module and they are always built together I tend to think that adding a dependency to the imp module does less harm than having a new intended-for-external-use-only operation exposed in the API that customer could see and use. Do not worry, I never depend on `impl`. The separation between API and IMPL is totally clear to me, I was relaxing that in this case on purpose just to avoid exposing that operation in the API.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/event/TokenRevocationRequestEventListener.java
```diff
@@ -76,7 +77,7 @@ public void on(LogoutRequestEvent e) {
         String jwt = tokenExtractor.getAccessToken(e.getRequest());
```
dogeared: small nit - can we call the variable something other than `e` so that it is clear what it is when used in the method? maybe: `logoutRequestEvent` or even `event`. Not a blocker.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/MinimalStormpathSpringSecurityWebMvcConfigurationIT.groovy
```diff
@@ -188,7 +228,7 @@ class MinimalStormpathSpringSecurityWebMvcConfigurationIT extends AbstractTestNG
         return account
     }
 
-    private Directory createTempDir() {
+    protected Directory createTempDir() {
```
dogeared: Why change private to protected?

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/MinimalStormpathSpringSecurityWebMvcConfigurationIT.groovy
```diff
@@ -201,23 +241,18 @@ class MinimalStormpathSpringSecurityWebMvcConfigurationIT extends AbstractTestNG
...
         def reversed = resourcesToDelete.reverse() //delete in opposite order (cleaner - children deleted before parents)
-
-        for (def r : reversed) {
-            try {
-                r.delete()
-            } catch (Throwable t) {
-                log.error('Unable to delete resource ' + r, t)
-            }
-        }
+        reversed.collect { it.delete() }
```
dogeared: Nice refactor!

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -45,13 +45,15 @@ public DefaultAccessToken(InternalDataStore dataStore, Map<String, Object> prope
      */
     private void ensureAccessToken() {
         try {
-            Claims claims = Jwts.parser()
-                .setSigningKey(getDataStore().getApiKey().getSecret().getBytes("UTF-8"))
-                .parseClaimsJws(getString(JWT)).getBody();
+            if(isMaterialized()) {
```
dogeared: space after if. If it's not materialized, then nothing happens here. What's the circumstance where the access token would not be materialized? Is it an error if this method is called and the access token is not materialized?

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/MinimalStormpathSpringSecurityWebMvcConfigurationIT.groovy
mrioan: Just for the sake of allowing potential subclasses of this IT to have access to create a temp directory. This is an IT, so I do not think there is a strong need on obscuring operations really.

dogeared:
I added a few comments in addition to @lhazlewood. Back to you @mrioan

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
mrioan: In particular due to the need to create the resource without contacting the server. We are instantiating this class only with the HREF so we do not contact the server. If I do not add the "is materialized" check then this code will try to read the JWT which is not really present and things blow up

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/MinimalStormpathSpringSecurityWebMvcConfigurationIT.groovy
dogeared: :+1:

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
dogeared: since this is a private method anyway, I think it would be more clear to have the if in the constructor. `ensureAccessToken` should *always* ensure that it's an accessToken according to the rules we've defined. So, in the constructor:

```
public DefaultAccessToken(InternalDataStore dataStore, Map<String, Object> properties) {
         super(dataStore, properties);
         if (isMaterialized()) {
             ensureAccessToken();
        }
     }
```

What do you think?

mrioan: What do you think if I do two things: 1. If it is materialized this method analyzes the JWT as of now. 2. If it is not materialized it analyzes that the href has the `/accessTokens/` string.

I mean, it is called "ensureAccessToken" so it should be trying to ensure in different scenarios. If you like that then the isMaterialized is not required in the constructor, only in the ensureAccessToken method

dogeared: yes, I like that. so, once you do that, the PR is approved. :+1:

mrioan: great

PR #495, opened 2016-02-25
Expose expandedJwt in AccessToken and RefreshToken resources

### LINKED ISSUE #485: We are not exposing the `expandedJwt` property of the Access and Refresh Tokens
See here: https://stormpath.atlassian.net/wiki/display/AM/Token+Authentication+and+Management

mrioan:
resolves #485 

The Backend was returning an `expandedJwt` property for both AccessToken and RefreshToken. We were not exposing that in our API.

I also discovered a bug where the `RefreshToken` was having the HREF of its associated `AccessToken` rather than the correct `RefreshToken` one.

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultGrantAuthenticationToken.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultGrantAuthenticationToken.java
@@ -81,8 +81,8 @@ public class DefaultGrantAuthenticationToken extends AbstractInstanceResource im
 
     public RefreshToken getAsRefreshToken(){
         Map<String, Object> props = new LinkedHashMap<String, Object>(1);
-        String accessTokenID = (String)((Map)this.getAsAccessToken().getExpandedJwt().get("claims")).get("rti");
-        props.put("href", "/refreshTokens/" + accessTokenID);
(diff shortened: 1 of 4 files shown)


impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultGrantAuthenticationToken.java
```diff
@@ -81,7 +81,8 @@ public AccessToken getAsAccessToken(){
 
     public RefreshToken getAsRefreshToken(){
         Map<String, Object> props = new LinkedHashMap<String, Object>(1);
-        props.put("href", this.getAccessTokenHref());
+        String accessTokenID = (String)((Map)this.getAsAccessToken().getExpandedJwt().get("claims")).get("rti");
```
dogeared: should this variable be called `refreshTokenID`?

dogeared:
one small change. Once complete, this is approved @mrioan.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultGrantAuthenticationToken.java
mrioan: Totally! Thanks!

PR #496, opened 2016-02-28
Coverage jacoco coveralls

### LINKED ISSUE #431: Integrate code coverage plugin for unit and integration tests
Jacoco? I say: Yes

Cobertura is not actively maintained.

IntelliJ supports using Jacoco as its coverage tool so command line will match IntelliJ

Need to determine baseline: What % coverage satisfies this ticket to start with? That should be enforced in the pom.xml.

dogeared:
resolves #431 

* Adds Jacoco test configuration for surefire and failsafe plugins
* Adds jacoco-mulit-coverage module for creating local aggregated coverage reports for unit and integration tests. This is necessary as jacoco does *not* natively support aggregated reports for multi-module maven projects
* Adds coveralls.io build to .travis.yml to report on aggregated code coverage for unit and integration tests.
* Moves docs build to oraclejdk7 from oraclejdk8
* Moves ITs run from jdk6 to jdk8 (reason is that maven coveralls plugin has to be jdk7 or newer AND we want the coveralls report to run after both unit and integration tests are run)
* Adds an initial unit test to oauth package as none existed before. This ensures that the oauth package is included in the stats for coverage
* Adds shell script to run all unit and integration tests and produce local aggregated coverage report

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - STORMPATH_API_KEY_ID_TWO_APP=VN3N09Z6BUE7QWL4RYX4FID3X
   - secure: ZAs6XDgRX1YIZAKf0IcR6DlfQtGLKJS5fyTOextZsRW3MCwu7WTBe5v6dAS6xoLqZ0iEkICMdIRhs+6DQEW6rju/W39mjxe8HJEcMNI+wZyW+rZ+j7NmPkjovMO+XbiR0aVVvn7ZGjWEKdozCTBVtJSLd8YVHh6Ey+yxOdDoc1A=
   - secure: GD3ZS7+oXVngRZdxPAjDbLNSjHNvKQ8IBXlo+DyDQ1kYAfyxSHD6eY0mW2SuqpUZ67e9XKhXd5z0VLgGAtn7JLco+56jyuBQKk8ak+KLctmUKIyFPNk6/11q/+vGIMEeKnmA5lI6UE+rcJOoSTtsubxbQkZE/mlZ79bckcx9W70=
+  - secure: XaSpWM+WSiXI3HnUL2WV3JD4Sl+dlqv37lvI+TFM0ns5UhCPylSE39h62ioJB85xLJKSpL0EZ/oVXPBdjY7ue3iaTbDtfiwpElCKyuizkzFHxy+3kGwpmq3SjVMaW57D86ubdpgkdzaHQr4aVzB+FhXl0Sdk6s20wGU+Sg/mqB4=
 before_install:
 - openssl aes-256-cbc -K $encrypted_da634cfd642b_key -iv $encrypted_da634cfd642b_iv
   -in deployKey_stormpath_github_io_id_rsa.enc -out ~/.ssh/id_rsa -d
@@ -35,6 +36,7 @@ notifications:
     on_failure: always
     rooms:
       secure: WXJLZTnleKQd5DIfdPSNYk4ZT7bssV9esdaacZjwb9sdKLZAQ+ujL97uuVK4oDMbua9XJxYTsKK45QQp56+PgJc4KG0Regvs/e7NtoxahrNnpsHCob4+Hc2MSmFfJgsdAQ9dFemNZ3BQ2ec1gO6SToYk+7pwMPScUYnmAOrsHjU=
-    template:
-      '%{repository}#%{build_number} (%{branch} - %{commit} : %{author}): %{message} (<a href="%{build_url}">Details</a>/<a href="%{compare_url}">Change view</a>)' 
+    template: '%{repository}#%{build_number} (%{branch} - %{commit} : %{author}):
+      %{message} (<a href="%{build_url}">Details</a>/<a href="%{compare_url}">Change
+      view</a>)'
     format: html
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ script:
 - test -z "$BUILD_DOCS" || ./build_docs.sh
 after_success:
 - test -z "$BUILD_DOCS" || test -z "$IS_RELEASE" || ./publish_docs.sh
+- mvn jacoco:report coveralls:jacoco
 notifications:
   hipchat:
     on_success: always
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
 - export IS_RELEASE="$([ ${RELEASE_VERSION/SNAPSHOT} == $RELEASE_VERSION ] && [ $TRAVIS_BRANCH
   == 'master' ] && echo 'true')"
 - export BUILD_DOCS="$([ $TRAVIS_JDK_VERSION == 'oraclejdk8' ] && echo 'true')"
-- export COVERALLS_REPORT="$([ $TRAVIS_JDK_VERSION == 'openjdk6' ] && echo 'true')"
+- export COVERALLS_REPORT="$([ $TRAVIS_JDK_VERSION == 'openjdk7' ] && echo 'true')"
 install:
 - test -z "$BUILD_DOCS" || pip -q install --user sphinx
 - echo "No need to run mvn install -DskipTests then mvn install.  Running mvn install."
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 - test -z "$BUILD_DOCS" || ./build_docs.sh
 after_success:
 - test -z "$BUILD_DOCS" || test -z "$IS_RELEASE" || ./publish_docs.sh
-- mvn clean test jacoco:report coveralls:report
+- mvn org.jacoco:jacoco-maven-plugin:prepare-agent verify coveralls:report
 notifications:
   hipchat:
     on_success: always
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ script:
 - test -z "$BUILD_DOCS" || ./build_docs.sh
 after_success:
 - test -z "$BUILD_DOCS" || test -z "$IS_RELEASE" || ./publish_docs.sh
-- mvn jacoco:report coveralls:jacoco
+- mvn clean test jacoco:report coveralls:report
 notifications:
   hipchat:
     on_success: always
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ before_install:
   -Dexpression=project.version|grep -Ev '(^\[|Download\w+:)')"
 - export IS_RELEASE="$([ ${RELEASE_VERSION/SNAPSHOT} == $RELEASE_VERSION ] && [ $TRAVIS_BRANCH
   == 'master' ] && echo 'true')"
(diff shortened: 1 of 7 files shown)


extensions/oauth/src/main/java/com/stormpath/sdk/impl/oauth/authc/AccessTokenAuthenticationRequest.java
```diff
@@ -122,6 +122,6 @@ public AccountStore getAccountStore() {
     /* @since 1.0.RC5 */
     @Override
     public AuthenticationOptions getResponseOptions() {
-        throw new UnsupportedOperationException(getClass().getName() + " .getResponseOptions() is not supported.");
+        throw new UnsupportedOperationException(getClass().getName() + ".getResponseOptions() is not supported.");
```
dogeared: Writing the test for this exposed the extra space in the message.

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+            accessTokenAuthenticationRequest.clear()
+            fail("shouldn't be here")
+        } catch (UnsupportedOperationException e) {
+            assertEquals e.getMessage(), "clear() method hasn't been implemented."
+        }
+    }
+
+    @Test
+    public void testGetResponseOptions() {
+
```
mrioan: Remove the first blank line in each method please

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+        ]
+
+        expect(httpRequest.getParameters())
+                .andReturn(params).times(6)
+        expect(httpRequest.getMethod())
+                .andReturn(HttpMethod.POST)
+        expect(httpRequest.getHeader("Content-Type"))
+                .andReturn("application/x-www-form-urlencoded")
+        expect(httpRequest.getHeader("Authorization"))
+                .andReturn("Basic " + new BASE64Encoder().encode("id:secret".bytes)).times(2)
```
mrioan: Just sharing info here: we have `com.stormpath.sdk.impl.util.Base64` which has static methods like `Base64.decodeBase64(...)` and `Base64.encodeBase64(...)`

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+        try {
+            accessTokenAuthenticationRequest.getAccountStore()
+            fail("shouldn't be here")
+        } catch (UnsupportedOperationException e) {
+            assertEquals e.getMessage(), "getAccountStore() method hasn't been implemented."
+        }
+    }
+
+    @Test
+    public void testDefaultTtl() {
```
mrioan: You could have a single test called something like "testGetters" were assert hasScopeFactory, getClientId, getClientSecret, etc in order to avoid having many small tests. Just a comment for future tests, no problem here

jacoco-multi-coverage/pom.xml
```diff
@@ -0,0 +1,275 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
mrioan: I am not sure I like the idea of having a new project in the root of the sdk for this... I cannot think of another alternative other than moving it to an internal sub-module but we do not have anything related to tests or similar... I think we will need to discuss with the team if the can think of another solution

run_all_tests_with_coverage.sh
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
```
mrioan: Can't this be automatically run using during the build process? I mean, during the build process all the tests will be executed but without coverage. Then we will need to have this command executed to test everything again. The coverage ideally should be processed during normal maven cycles. 

Additionally, if I am not wrong the message "All tests passed. Hooray!" will always be displayed disregarding errors because the command is flagged with --fail-never, right?

mrioan:
@dogeared back to you

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
dogeared: I've always had it that unit tests should test one thing. You end up with many small tests, but if something changed with `getAccountStore`, for instance, only that one test would break.

jacoco-multi-coverage/pom.xml
dogeared: Definitely open to discussion. Jacoco (and cobertura and others), do not naturally support aggregating test results. I thought this was a good way to keep it isolated with a relatively small footprint. After we achieve a certain level of coverage, we might even opt to remove it from the project.

run_all_tests_with_coverage.sh
dogeared: The --fail-never flag makes it so the tests won't halt on failure. But, it should execute with a non-zero value if there were failures along the way.

generating the aggregated reports takes some time, so I thought it would be better to keep it separate.

What we could do is to always update the instrumentation and then just have the script generate the tests coverage reports. 

What do you think?

.travis.yml
```diff
@@ -19,7 +20,8 @@ before_install:
   -Dexpression=project.version|grep -Ev '(^\[|Download\w+:)')"
 - export IS_RELEASE="$([ ${RELEASE_VERSION/SNAPSHOT} == $RELEASE_VERSION ] && [ $TRAVIS_BRANCH
   == 'master' ] && echo 'true')"
-- export BUILD_DOCS="$([ $TRAVIS_JDK_VERSION == 'oraclejdk8' ] && echo 'true')"
+- export BUILD_DOCS="$([ $TRAVIS_JDK_VERSION == 'oraclejdk7' ] && echo 'true')"
```
lhazlewood: Why did you change this to JDK7?

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
lhazlewood: Yeah, we should change this to not reference sun internal implementation classes.  JDK8 also has a built-in Base64 codec:  https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

jacoco-multi-coverage/pom.xml
lhazlewood: If there was a way to add this to the root of the project, in the root pom (i.e. stormpath-sdk-java/pom.xml), that'd be better.  Is this possible?  Also, did you find anything out after talking to @jeffwysong about what we did in other projects?  Or does the current PR reflect those things?

jacoco-multi-coverage/pom.xml
```diff
@@ -0,0 +1,275 @@
...
+        <classes.stormpath-spring-security>${build.stormpath-spring-security}/classes</classes.stormpath-spring-security>
+        <classes.stormpath-spring-security-webmvc>${build.stormpath-spring-security-webmvc}/classes</classes.stormpath-spring-security-webmvc>
+        <classes.stormpath-spring-boot-starter>${build.stormpath-spring-boot-starter}/classes</classes.stormpath-spring-boot-starter>
+        <classes.stormpath-webmvc-spring-boot-starter>${build.stormpath-webmvc-spring-boot-starter}/classes</classes.stormpath-webmvc-spring-boot-starter>
+        <classes.stormpath-spring-security-spring-boot-starter>${build.stormpath-spring-security-spring-boot-starter}/classes</classes.stormpath-spring-security-spring-boot-starter>
+        <classes.stormpath-spring-security-webmvc-spring-boot-starter>${build.stormpath-spring-security-webmvc-spring-boot-starter}/classes</classes.stormpath-spring-security-webmvc-spring-boot-starter>
+
+
+        <!-- sources -->
+        <sources.api>${base.api}/src/main/java</sources.api>
```
lhazlewood: Unless I'm mistaken, this does not look sustainable: we wouldn't naturally remember to update this stuff if/when we add new modules.

lhazlewood: Perhaps its time to enable Clover?

https://confluence.atlassian.com/display/CLOVER/Clover-for-Maven+2+and+3+User's+Guide#Clover-for-Maven2and3User'sGuide-Workingwithmulti-moduleprojects

.travis.yml
dogeared: I moved the run of the ITs to JDK8. So, I moved the build docs to JDK7 to keep the time for each JDK build as short as possible.

jacoco-multi-coverage/pom.xml
dogeared: I did talk to @jeffwysong. The engineering team bailed on trying to make Jacoco provide aggregated results and went with Clover.

I like having this as a separate module because it keeps it contained and easily removed if we so choose later on.

jacoco-multi-coverage/pom.xml
dogeared: The intention of this module is to provide a local, aggregated, complete set of test coverage results. I don't like all the manual labor this represents either.

I am looking into some maven or ant plugins that would allow this to automatically discover the folders we need from the other modules.

I also think there may be value in having it setup this way as a stop-gap until we achieve a minimum level of coverage. At that point, we could remove this module. It doesn't seem likely that we will be adding new modules while we are working on expanded coverage, so I think this would work in the short term.

.travis.yml
lhazlewood: Cool - thanks!

extensions/oauth/src/test/groovy/com/stormpath/sdk/impl/oauth/AccessTokenAuthenticationRequestTest.groovy
lhazlewood: @dogeared still waiting on this change before I can approve the PR

dogeared: I will definitely make this change. However, I am going to close this PR and create a new branch using `clover`. We can compare the results, but I believe if we can afford the expense, clover is the way to go.

It will provide aggregated local reports as well as centralized historical reports online.

dogeared:
Closing (for now) to investigate clover.

PR #502, opened 2016-03-01
#445 Disabled Group still allows access in endpoint protected by Spring Security

### LINKED ISSUE #445: Disabled Group still allows access in endpoint protected by Spring Security
As reported by @jaschenk:

Using the latest Java SDK 1.0.RC8, I have found the following security issue with Groups. 
Current flow:

* Application Boots (Java Spring Boot) 
* Perform a Login and test Restricted Access. Access correctly ALLOWED per Group.

* Go to Stormpath Admin Console and Disable the Group.

* Attempt a Login and test Restricted Access, Access ALLOWED when a Group was Disabled. This does not seem correct!

* Log out. 
* Go to Stormpath Admin Console Re-enable disabled Group. 
* Go to specific Login Account for Application and Remove Group. 
* Login to App and test Restricted Access, Access Forbidden. 
* Re-Add Group to Account. 
* Login to App and test Restricted Access, Access Correctly ALLOWED.

I verified this behavior configuring the [Spring Boot tutorial](https://github.com/stormpath/stormpath-sdk-java/tree/master/tutorials/spring-boot/03-spring-security-refined) to use a Group. When I disable the Group, users in that Group still have access to the `/restricted` page - even after rebooting the application.

We will need to check and see if this is just a Spring Security issue or if it is a general issue in the JSDK.

jarias:
Resolves #445

### DIFF
--- a/extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
+++ b/extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
@@ -44,11 +44,10 @@ import java.util.Set;
  * A {@code AuthenticationProvider} implementation that uses the <a href="http://www.stormpath.com">Stormpath</a> Cloud Identity
  * Management service for authentication and authorization operations for a single Application.
  * <p/>
- * The Stormpath-registered
- * <a href="https://www.stormpath.com/docs/libraries/application-rest-url">Application's Stormpath REST URL</a>
- * must be configured.
+ * It requires the <a href="https://docs.stormpath.com/java/spring-boot-web/config.html#stormpath-application">Stormpath Application</a> be configured
+ * <p/>
  * <h3>Authentication</h3>
- * Once your application's REST URL is configured, this provider implementation automatically executes authentication
+ * Once your application's is configured, this provider implementation automatically executes authentication
  * attempts without any need of further configuration by interacting with the Application's
  * <a href="http://www.stormpath.com/docs/rest/api#ApplicationLoginAttempts">loginAttempts endpoint</a>.
  * <h3>Authorization</h3>
@@ -142,7 +141,6 @@ import java.util.Set;
  */
 public class StormpathAuthenticationProvider implements AuthenticationProvider {
 
-    private final Client client;
     private final Application application;
 
     private GroupGrantedAuthorityResolver groupGrantedAuthorityResolver;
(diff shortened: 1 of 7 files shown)


extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -150,7 +152,16 @@
 
     private Application application; //acquired via the client at runtime, not configurable by the StormpathAuthenticationProvider user
 
-    public StormpathAuthenticationProvider() {
+    public StormpathAuthenticationProvider(Client client, String applicationRestUrl) {
```
lhazlewood: Why pass in the `applicationRestUrl` and not the `Application` instance?

In fact, why is the url needed at all?  Every Stormpath Spring application should have an Application instance created at startup - there shouldn't be a need to look it up here in the provider.  Thoughts?

jarias: I'll check the code, since that change was just a refactor to do the state check for those two fields in the constructor instead on every authentication, something that I found while creating my test, thus I kept it almost the same.

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -44,7 +46,7 @@
  * <p/>
  * The Stormpath-registered
  * <a href="https://www.stormpath.com/docs/libraries/application-rest-url">Application's Stormpath REST URL</a>
```
lhazlewood: I'm not sure this makes that much sense here given the implementation change to use an Application.  Yes, an App HREF is required, but I think we should be more specific to our integration via this URL:

https://docs.stormpath.com/java/spring-boot-web/config.html#stormpath-application

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -44,7 +46,7 @@
...
- * must be configured as the {@code applicationRestUrl} property.
+ * must be configured.
  * <h3>Authentication</h3>
  * Once your application's REST URL is configured, this provider implementation automatically executes authentication
```
lhazlewood: same

extensions/spring/stormpath-spring-security/src/test/groovy/com/stormpath/spring/security/provider/StormpathAuthenticationProviderTest.groovy
```diff
@@ -25,87 +25,171 @@ import com.stormpath.sdk.directory.CustomData
...
+    def acctGivenName = 'John'
+    def acctMiddleName = 'A'
+    def acctSurname = 'Smith'
+    def acctStatus = AccountStatus.ENABLED
+
+    @BeforeMethod
     void setUp() {
-        authenticationProvider = new StormpathAuthenticationProvider()
+        client = createStrictMock(Client)
+        authenticationProvider = new StormpathAuthenticationProvider(client, applicationUrl)
```
lhazlewood: Shouldn't this take an application not a url?

lhazlewood:
LGTM - please merge!

PR #505, opened 2016-03-04
enabled setting baseUrl in servlet plugin

dogeared:


dogeared:
resolves #506

jeremypepper:
Thanks!

PR #507, opened 2016-03-04
498 fixing error 500 by checking username hasText

### LINKED ISSUE #498: Leaving username empty in login form results in user seeing a 500 error
If the user enters an invalid username in the login form then they see a reasonable error message and can correct their mistake.

However if they leave the username empty they see a raw 500 error with a fairly non-end user friendly message. E.g. with the custom error controller used in most of the Spring examples I see:

image

If we look at the stack trace for this issue we see there seem to be different expectations at different levels:
```
Daemon Thread [http-nio-8080-exec-6]
    DefaultUsernamePasswordRequestBuilder.setUsernameOrEmail(String) line: 37    
    StormpathAuthenticationProvider.createAuthenticationRequest(Authentication) line: 425    
    StormpathAuthenticationProvider.handleUsernamePasswordAuthentication(Authentication) line: 378    
    StormpathAuthenticationProvider.authenticate(Authentication) line: 357    
    ProviderManager.authenticate(Authentication) line: 167    
    ProviderManager.authenticate(Authentication) line: 192    
    UsernamePasswordAuthenticationFilter.attemptAuthentication(HttpServletRequest, HttpServletResponse) line: 93    
    ...
```
In the `UsernamePasswordAuthenticationFilter.attemptAuthentication(...)` code we see:
```
if (username == null) {
    username = "";
}
```
I.e. it seems to be OK with the idea of the username not being present, however then in the lower down `DefaultUsernamePasswordRequestBuilder.setUsernameOrEmail(String)` call (that we see in the stack trace above) we see:
```
Assert.hasText(usernameOrEmail, "usernameOrEmail cannot be null or empty.");
```
It's this assertion that results in things blowing up and the user seeing the 500 error.

I think an empty username should be handled as nicely, from the end user's point of view, as an invalid username.

This issue is seen when using the Spring Boot setup and using version 1.0.RC8.4.

mzumbadoate:
Resolves #498

### DIFF
--- a/extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
+++ b/extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
@@ -422,6 +422,11 @@ public class StormpathAuthenticationProvider implements AuthenticationProvider {
     protected AuthenticationRequest createAuthenticationRequest(Authentication authentication) {
         String username = (String) authentication.getPrincipal();
         String password = (String) authentication.getCredentials();
+
+        if (!Strings.hasText(username)){
+            throw new AuthenticationServiceException("must have username and password");
+        }
+
         return UsernamePasswordRequest.builder().setUsernameOrEmail(username).setPassword(password).build();
     }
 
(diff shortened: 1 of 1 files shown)


mzumbadoate:
Resolves #498

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -422,6 +422,11 @@ protected final Application ensureApplicationReference() {
     protected AuthenticationRequest createAuthenticationRequest(Authentication authentication) {
         String username = (String) authentication.getPrincipal();
         String password = (String) authentication.getCredentials();
+
+        if (!Strings.hasText(username)){
+            throw new AuthenticationServiceException("must have username and password");
```
lhazlewood: This isn't strictly true - Stormpath supports both usernames and email addresses for logins.  Maybe something like "Login and password required" to be generic?

lhazlewood:
LGTM, please merge.

PR #508, opened 2016-03-07
Fixes #500

### LINKED ISSUE #500: Create Internationalization Integration Test
Related to #427

Because there was confusion as to how to setup internationalization, this will be valuable in making it clear.

jarias:
Create Internationalization Integration Test

#500

### DIFF
--- a/examples/servlet/pom.xml
+++ b/examples/servlet/pom.xml
@@ -200,7 +200,7 @@
             <activation>
                 <!-- only run the IT tests on one of the JVM builds in Travis CI.  This ensures that we don't hammer
                 the REST api unnecessarily. -->
-                <jdk>1.8</jdk>
+                <jdk>[1.7,1.8)</jdk>
             </activation>
             <properties>
                 <skipITs>false</skipITs>
--- a/examples/servlet/pom.xml
+++ b/examples/servlet/pom.xml
@@ -19,7 +19,13 @@
 
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>com.stormpath.sdk</groupId>
+    <parent>
+        <groupId>com.stormpath.sdk</groupId>
+        <artifactId>stormpath-sdk-examples</artifactId>
+        <version>1.0.RC9-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
     <artifactId>stormpath-sdk-examples-servlet</artifactId>
     <version>1.0.RC9-SNAPSHOT</version>
     <name>Stormpath Java SDK :: Examples :: Servlet Container-based Webapp</name>
@@ -32,12 +38,6 @@
         <logback.version>1.1.2</logback.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <servlet.version>3.0.1</servlet.version>
-        <groovy.version>1.8.8</groovy.version>
-        <testng.version>6.8</testng.version>
-        <restassured.version>2.8.0</restassured.version>
-        <slf4j.version>1.7.18</slf4j.version>
-        <gmavenplus.version>1.5</gmavenplus.version>
-        <skipITs>true</skipITs>
     </properties>
 
     <dependencies>
@@ -66,32 +66,23 @@
         <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
-            <version>${testng.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.codehaus.groovy</groupId>
             <artifactId>groovy-all</artifactId>
-            <version>${groovy.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>com.jayway.restassured</groupId>
             <artifactId>rest-assured</artifactId>
-            <version>${restassured.version}</version>
             <scope>test</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>commons-logging</groupId>
-                    <artifactId>commons-logging</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <!-- JCL is needed by Rest Assured, but we want to use SLF4J's instead: -->
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
-            <version>${slf4j.version}</version>
+            <scope>runtime</scope>
         </dependency>
     </dependencies>
     <build>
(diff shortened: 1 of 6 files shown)


examples/servlet/pom.xml
```diff
@@ -31,6 +32,12 @@
         <logback.version>1.1.2</logback.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <servlet.version>3.0.1</servlet.version>
+        <groovy.version>1.8.8</groovy.version>
+        <testng.version>6.8</testng.version>
+        <restassured.version>2.8.0</restassured.version>
```
mrioan: Can you please move newly added dependencies to the root pom? They should be referenced here but not added.

examples/servlet/pom.xml
```diff
@@ -56,11 +63,83 @@
...
+                            <goal>testCompile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <invokeDynamic>true</invokeDynamic>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
mrioan: We already have surefire and failsafe in the root pom, didn't it work here? It should and therefore these plugins should not be required here. Take a look at `stormpath-spring-security-webmvc`. That project hast ITs running but the plugin does not show up in its `pom.xml`

examples/servlet/pom.xml
```diff
@@ -92,4 +194,17 @@
         </plugins>
     </build>
 
+    <profiles>
```
mrioan: We do not follow this procedure in the other modules. We basically have defined a project wide `skiptITs` so ITs never run unless explicitly adding `skipITs=true` when executing maven. This may be a good approach but we need to define this project-wise. Therefore, for now please remove the profile. Thanks!

examples/servlet/src/test/groovy/com/stormpath/sdk/examples/servlet/i18n/I18nIT.groovy
```diff
@@ -0,0 +1,42 @@
...
+import org.testng.annotations.Test
+
+import static com.jayway.restassured.RestAssured.given
+import static org.hamcrest.Matchers.equalTo
+
+/**
+ *
+ */
+class I18nIT {
+    @Test
```
mrioan: Blank line here please

examples/servlet/src/test/groovy/com/stormpath/sdk/examples/servlet/i18n/I18nIT.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+import com.jayway.restassured.RestAssured
+import com.jayway.restassured.config.DecoderConfig
+import org.testng.annotations.Test
+
+import static com.jayway.restassured.RestAssured.given
+import static org.hamcrest.Matchers.equalTo
+
+/**
+ *
```
mrioan: `@since 1.0.XX` here please

examples/servlet/src/test/groovy/com/stormpath/sdk/examples/servlet/i18n/I18nIT.groovy
```diff
@@ -0,0 +1,42 @@
...
+                .header("Accept-Language", "es")
+                .when()
+                .get("http://localhost:8080/register")
+                .then()
+                .body(
+                "html.head.title", equalTo("Crear Cuenta"),
+                "**.findAll { it.@type == 'submit' }", equalTo("Crear Cuenta"),
+                "**.findAll { it.@class == 'to-login' }", equalTo("Regresar a Iniciar Sesión")
+        )
+    }
```
mrioan: Please add another test asserting that by default (no Accept-language) the language texts are displayed in English

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
```diff
@@ -36,12 +37,15 @@ public String getMessage(String key, Locale locale) {
...
         ResourceBundle bundle = getBundle(locale);
-        String msg = bundle.getString(key);
 
         try {
+            String msg = new String(bundle.getString(key).getBytes("ISO-8859-1"), "UTF-8");
             return MessageFormat.format(msg, args);
         } catch (MissingResourceException e) {
             return '!' + key + '!';
+        } catch (UnsupportedEncodingException e) {
+            //Should not happen
```
mrioan: can you elaborate that comment please?: "Should not happen because... "

mrioan:
@jarias First round of review ready, back to you. Thanks!

examples/servlet/pom.xml
jarias: There is no root pom for the examples, I could move the test to the servlet module instead but then It would have a dependency on the servlet example war to be able to the start the app.

examples/servlet/pom.xml
jarias: Same as above the examples don't have a root pom.

examples/servlet/src/test/groovy/com/stormpath/sdk/examples/servlet/i18n/I18nIT.groovy
jarias: Sure thing

examples/servlet/pom.xml
lhazlewood: I think @mrioan is saying that all dependencies - no matter what for - are declared in the project root pom.xml and referenced by submodules.  This better ensures that two different submodules do not accidentally pull in different versions of the same library because the module author can always check the root project pom to see what is globally referenceable

lhazlewood: In other words, the examples/servlet/pom.xml should even have a `<properties>` section that defines library versions since those libs/versions should be declared in the root pom.

If a submodule needs a different version of a library for a very specific reason - perhaps that example is not compatible with the servlet API used in the rest of the project (for example), only then would it make sense to declare that particular version in the local pom.xml file IMO.

examples/servlet/pom.xml
lhazlewood: They do have a root pom - the examples pom extends from the project root pom.

examples/servlet/src/test/groovy/com/stormpath/sdk/examples/servlet/i18n/I18nIT.groovy
```diff
@@ -0,0 +1,42 @@
...
+                    .body(
+                        "html.head.title", equalTo("Iniciar Sesión"),
+                        "**.findAll { it.@class == 'header' }.span.a", equalTo("Crear Cuenta"),
+                        "**.findAll { it.@type == 'submit' }", equalTo("Iniciar Sesión"),
+                        "**.findAll { it.@class == 'to-login' }", equalTo("Se te olvidó tu contraseña?")
+                    )
+    }
+
+    @Test
+    void testRegisterPage() {
```
lhazlewood: Very nice test!  I like the `findAll` approach of verifying body contents :)

examples/servlet/pom.xml
jarias: Yeah but the the servlet example doesn't have a parent.

lhazlewood: Ahhhh - interesting.  @mrioan and @dogeared is this because we wanted to run them standalone for copy-paste reasons (so other people didn't have to depend on our SDK pom?)

mrioan: @lhazlewood That was the original idea. It seems that if we want to run ITs directly in the examples then we will need a root pom for the examples. This way we will still be able to depend on our jar files (rather than on the source code) while having all plugins/dependencies available in the root pom of the examples.

The question is, can the examples' root pom have the SDK as parent (as it currently is)?

In summary: example1, example2 and example3 should examples/pom.xml as their parent. But, is it really possible? Right now examples/pom.xml has the root SDK as parent. So, by transitive nature, example1, example2, example3, will then have the root SDK pom as parent as well?

lhazlewood: What was the outcome of the standup today?  Will we have a shared parent or not for the examples?

lhazlewood:
@jarias build is failing - can you please investigate?

jarias:
@lhazlewood yeah I'm still doing some local changes, the outcome was to depend on the examples pom instead of the root pom and publish that examples pom so people can run the examples standalone, at the same time sharing configuration and dependencies between examples

dogeared:
LGTM. After CI passes, please merge.

PR #514, opened 2016-03-09
Updated quickstart documentation to better clarify how to enable and disable Spring Security

### LINKED ISSUE #484: Spring Boot Spring Security WebMVC does not start when stormpath.spring.security.enabled is disabled
If we set:

stormpath.spring.security.enabled = false
security.basic.enabled = false

Then, when we `start stormpath-sdk-examples-spring-security-spring-boot-webmvc` we get:

`AlreadyBuiltException: This object has already been built`

BTW, these properties are mentioned in our tutorial: https://docs.stormpath.com/java/spring-boot-web/quickstart.html

lhazlewood:
Fixes #484

Documentation (as opposed to code) was the chosen solution because it is not possible to both disable Spring Security and keep a SpringSecurity-specific configuration class in the code project.  There isn't anything we can do code-wise to automatically disable our behavior when user code explicitly calls `apply(stormpath())`.

The only real approach is to document that that method call should be removed as well as set some config properties.

### DIFF
--- a/extensions/spring/boot/docs/source/quickstart.rst
+++ b/extensions/spring/boot/docs/source/quickstart.rst
@@ -42,7 +42,11 @@ Using your favorite dependency resolution build tool like Maven or Gradle, add t
     }
 
 
-In order to connect Stormpath and Spring Security, we need one small configuration class in your project:
+Spring Security
+^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


mzumbadoate:
CR +1 - please merge

PR #516, opened 2016-03-09
Deprecated UsernamePasswordRequest and introduced UsernamePasswordRequests

### LINKED ISSUE #408: Deprecated UsernamePasswordRequest has no alternative
All the methods in `UsernamePasswordRequest` appear to be deprecated. Why not just deprecate the class. Since this is a public api the docs for deprecated stuff should point the user at the new stuff like so

```
/**
 * Does some thing in old style.
 *
 * @deprecated use {@link #new()} instead.  
 */
@Deprecated
public void old() {
// ...
}
```

lhazlewood:
Fixes #408 

Alternative class now reflects the same pattern we use elsewhere in the SDK for utility classes (e.g. `Accounts`, `Directories`, `Applications`, etc).

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/authc/UsernamePasswordRequest.java
+++ b/api/src/main/java/com/stormpath/sdk/authc/UsernamePasswordRequest.java
@@ -20,25 +20,28 @@ import com.stormpath.sdk.lang.Assert;
 import com.stormpath.sdk.lang.Classes;
 
 /**
- * A {@code UsernamePasswordRequest} is an {@code AuthenticationRequest} that represents a username (or email) +
- * password pair.  It optionally supports
- * {@link #setAccountStore(com.stormpath.sdk.directory.AccountStore) targeting an specific account store} as well for
- * customized authentication behavior.
  * <p>
- * NOTE: This class has been deprecated and will be removed in version 1.0. You can now use the new fluent interface
- * to create Username/Password Requests. For example:</p>
+ * NOTE: This class has been deprecated and will be removed in version 1.0. Use the
+ * {@link UsernamePasswordRequests#builder() UsernamePasswordRequests.builder()} instead.  For example:</p>
  * <pre>
- * AuthenticationRequest request = UsernamePasswordRequest.builder()
+ * AuthenticationRequest request = UsernamePasswordRequests.builder()
  *                         .setUsernameOrEmail(username)
  *                         .setPassword(submittedRawPlaintextPassword)
  *                         .build();
  * Account authenticated = application.authenticateAccount(request).getAccount();
  * </pre>
  *
+ * <p>A {@code UsernamePasswordRequest} is an {@code AuthenticationRequest} that represents a username (or email) +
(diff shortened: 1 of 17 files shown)


api/src/main/java/com/stormpath/sdk/authc/UsernamePasswordRequest.java
```diff
@@ -20,25 +20,28 @@
...
+ * AuthenticationRequest request = UsernamePasswordRequests.builder()
  *                         .setUsernameOrEmail(username)
  *                         .setPassword(submittedRawPlaintextPassword)
  *                         .build();
  * Account authenticated = application.authenticateAccount(request).getAccount();
  * </pre>
  *
+ * <p>A {@code UsernamePasswordRequest} is an {@code AuthenticationRequest} that represents a username (or email) +
+ * password pair.  It optionally supports
+ * {@link #setAccountStore(com.stormpath.sdk.directory.AccountStore) targeting an specific account store} as well for
```
mrioan: inAccountStore(AccountStore)

mrioan:
Looks good to me. Approved. Thanks! Back to you @lhazlewood

lhazlewood:
Thanks!  I have some tests to update that I found in addition to your comments.  I'll update the PR shortly.

lhazlewood:
Also, FWIW @mrioan this issue wasn't in the 'Needs Review' column in waffle.io - it wasn't ready to be reviewed just yet.  But I appreciate your eagerness!!! :-D

mrioan:
@lhazlewood I removed the single comment I made. The `inAccountStore` operation is part of the builder but in `AuthenticationRequest` it is called setAacountStore. I was confused because the javadoc was new but for the old code. So everything in your commit was OK.

lhazlewood:
Thanks!

lhazlewood:
Travis is being really slow today for some reason.  I think this issue is complete assuming Travis passes.

examples/quickstart/src/main/java/quickstart/Quickstart.java
```diff
@@ -104,10 +104,10 @@ public static void main(String[] args) {
...
 
         // Create an authentication request using the credentials
-        AuthenticationRequest request = UsernamePasswordRequest.builder()
-            .setUsernameOrEmail(usernameOrEmail)
-            .setPassword(rawPassword)
-            .build();
+        AuthenticationRequest request = UsernamePasswordRequests.builder()
+                .setUsernameOrEmail(usernameOrEmail)
+                .setPassword(rawPassword)
+                .build();
```
dogeared: Not a major issue, but I notice that in some of our code we have a double-indent for code continuation (like this) and elsewhere we have single-indent.

I suspect that is based on our individual IDE settings. Have we established a standard that we should all conform to?

lhazlewood: Until we decide otherwise, we should probably be using the [Java Coding Style Standard](http://www.oracle.com/technetwork/java/codeconvtoc-136057.html)

PR #523, opened 2016-03-11
396 adding the validation to spring sec

### LINKED ISSUE #396: i18n messages in SpringSecurityLoginErrorModelFactory & LoginErrorModelFactory
Currently, both these classes have a hard-coded error message in them:

```
"Invalid username or password."
```

They should be refactored to us i18n messages.

cc @lhazlewood & @george-hawkins-aa

mzumbadoate:
resolves #396

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginErrorModelFactory.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginErrorModelFactory.java
@@ -1,12 +1,15 @@
 package com.stormpath.sdk.servlet.mvc;
 
 import com.stormpath.sdk.servlet.form.Form;
+import com.stormpath.sdk.servlet.i18n.DefaultMessageSource;
+import com.stormpath.sdk.servlet.i18n.MessageSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletRequest;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 /**
  * Generates the error model on a failed login attempt.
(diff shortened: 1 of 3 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginErrorModelFactory.java
```diff
@@ -16,14 +19,20 @@
...
             List<String> errors = new ArrayList<String>(1);
-            errors.add("Invalid username or password.");
+            errors.add(getInvalidLoginMessage(request.getLocale()));
             return errors;
         }
         return null;
     }
+
+    private String getInvalidLoginMessage(Locale locale) {
+        MessageSource messageSource = new DefaultMessageSource();
```
jarias: Should should get the MessageSource singleton from the configuration

See RegisterFilter

Should be something like this:

```java
public static final String MESSAGE_SOURCE = "stormpath.web.message.source";

Config config = ConfigResolver.INSTANCE.getConfig(request.getServletContext())
MessageSource messageSource = config.getInstance(MESSAGE_SOURCE);
```
Otherwise if the user configures a different message source you would still use the default one

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/SpringSecurityLoginErrorModelFactory.java
```diff
@@ -2,25 +2,35 @@
...
         if (query != null && query.contains("error")) {
-            return Collections.singletonList("Invalid login or password.");
+            return Collections.singletonList(getInvalidLoginMessage(request.getLocale()));
         }
 
         return null;
     }
+
+    private String getInvalidLoginMessage(Locale locale) {
+        MessageSource messageSource = new DefaultMessageSource();
```
jarias: Similar to the above but since this spring the message source would probably come from a spring bean.

jarias:
Back to you @mzumbadoate

jarias:
@mzumbadoate LGTM

PR #524, opened 2016-03-11
164 cache expansions tests

### LINKED ISSUE #164: Cache does not return correctly resources with custom data or expansions

jarias:
Fixes #164


extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
```diff
@@ -617,6 +617,33 @@ class ApplicationIT extends ClientIT {
...
+            def account = createTestAccount(client, app)
+            deleteOnTeardown(account)
+        }
+
+        //Get application without expanding the accounts attribute
+        def appWithoutAccounts = client.getApplications(Applications.where(Applications.name().eqIgnoreCase(app.name))).first()
+        assertNull getValue(AbstractResource, appWithoutAccounts, "properties").get("accounts").get("items"), "Application->Accounts shouldn't be expanded"
+
+        //Get the application again twice expanding the accounts attribute
+        2.times {
```
dogeared: What's the purpose of the `2.times` here? Won't it be getting the first application and the same accounts each time?

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
```diff
@@ -132,10 +132,7 @@ private boolean isCacheable(ResourceDataRequest request, ResourceDataResult resu
             AbstractResource.isMaterialized(result.getData()) &&
 
             //@since 1.0.RC7: Let's not cache Access Tokens
-            !AccessToken.class.isAssignableFrom(clazz) &&
-
-            //@since 1.0.RC4.6: Fix for https://github.com/stormpath/stormpath-sdk-java/issues/164. Let's not cache expanded resources
```
dogeared: I'm confused by this. Why are we reversing what was added to fix an issue before?

I guess the question is, do we want to cache expanded resources or not?

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
jarias: Yeah the idea is that the second time we hit the cache

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
jarias: Well the test that Anai did was to prove the expanded resources caching was broken, I remove this line and run the fixed test and it seems the caching resources with expanded attributes is not broken at all. I guess your question is valid do we want to cache resources with expanded attributes I think we should, but we should figure out whats the expected behavior which is something I put in as comment in the ticket

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
jarias: On another note since I change Anai's test the diff doesn't show the prev version, you can look at it in the commit list just for reference, I found there was an error in the test.

mrioan:
@jarias So, in summary. You mean that:

1. Expanded Resources are working fine with Cache
2. We have a good test that will fail if bullet 1 ever gets false

Right?

If so, this PR looks good to me.

jarias:
@mrioan That is correct

PR #528, opened 2016-03-15
409 fix

### LINKED ISSUE #409: Enhance user experience regarding secure connection requirements
For any connection other than localhost, Stormpath requires a secure connection.

This may not be clear to people deploying to non-SSL protected environments.

This could result in confusion when code that was working locally suddenly stops working in a deployment environment.

lhazlewood:
Resolves #409 

Ensured log messages would be printed if not in localhost development

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAuthenticationResultSaver.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAuthenticationResultSaver.java
@@ -23,6 +23,9 @@ import com.stormpath.sdk.servlet.config.CookieConfig;
 import com.stormpath.sdk.servlet.http.CookieSaver;
 import com.stormpath.sdk.servlet.http.Resolver;
 import com.stormpath.sdk.servlet.http.Saver;
+import com.stormpath.sdk.servlet.util.SecureRequiredExceptForLocalhostResolver;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -157,7 +158,18 @@
             <version>${easymock.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.powermock</groupId>
```
dogeared: is powermock required because of static or constructor tests?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAuthenticationResultSaver.java
```diff
@@ -84,8 +91,39 @@ protected void remove(HttpServletRequest request, HttpServletResponse response)
...
+        boolean likelyLocalhost = resolver instanceof SecureRequiredExceptForLocalhostResolver;
+
+        boolean warnable = !configSecure || (!resolverSecure && !likelyLocalhost);
+
+        if (!secureWarned && warnable) {
+            secureWarned = true;
+            String msg = "INSECURE IDENTITY COOKIE CONFIGURATION: Your current Stormpath SDK account cookie " +
+                    "configuration allows insecure identity cookies (transmission over non-HTTPS connections)!  " +
+                    "This should never occur on staging/production machines otherwise your users will be " +
+                    "susceptible to man-in-the-middle attacks.  If you see this warning on a development-only " +
```
dogeared: In a development only environment that's *not* localhost, they'd still have the situation where the account cookie would not be set.

pom.xml
lhazlewood: Yes, `LoggerFactory.getLogger` is a static call.

dogeared: :+1:

dogeared:
@lhazlewood - pending my note about the development environment sentence, this is good to go.

PR #533, opened 2016-03-17
#421 Most views should redirect if already authenticated

### LINKED ISSUE #421: Most views should redirect if already logged in
All stormpath views *except* 'change password' (note this is change password - not 'forgot password') should redirect an already-logged-in user to the login nextUri.

Allowing users to indicate they forgot their password or register another user while already logged in is unintuitive at best, and might be a security concern at worst.

All GET requests should all be redirected and any POST should return an error.  The reason is that if a GET always redirects, it should never be possible to POST (since you'd never see the original form that the GET would have returned).  If POSTing occurs, then, it is probably someone trying to be sneaky, or something went wrong (e.g. CDN caching a view which shouldn't be cached), both of which should be prevented for safety/security.

jarias:
Fixes #421


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
```diff
@@ -28,10 +30,30 @@
  */
 public abstract class AbstractController implements Controller {
 
-    private static final HttpServlet DEFAULT_HANDLER = new HttpServlet(){};
+    private static final HttpServlet DEFAULT_HANDLER = new HttpServlet() {
+    };
```
dogeared: FWIW, I think it's ok on one line. The space is good, though.

dogeared:
nice work! LGTM - please merge

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
jarias: Yeah that's Intellij formatting :(

PR #538, opened 2016-03-22
Refactored to make sure a redirect is not treated like a retry.

### LINKED ISSUE #539: Fix artificial wait on redirects
This occurs because currently, a redirect is treated as a retry.

#538 address this.

dogeared:
resolves #539 

This ensures that there will not be an artificial wait when there's a redirect.

### DIFF
--- a/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -238,7 +238,7 @@ public class HttpClientRequestExecutor implements RequestExecutor {
                 );
             }
 
-            if (retryCount > 0) {
+            if (retryCount > 0 || redirectUri != null) {
                 request.setQueryString(originalQuery);
                 request.setHeaders(originalHeaders);
             }
@@ -268,7 +268,6 @@ public class HttpClientRequestExecutor implements RequestExecutor {
                 }
 
                 exception = null;
-                retryCount++;
 
                 //long start = System.currentTimeMillis();
                 httpResponse = httpClient.execute(httpRequest);
@@ -282,6 +281,7 @@ public class HttpClientRequestExecutor implements RequestExecutor {
                     redirectUri = URI.create(location);
                     httpRequest.setURI(redirectUri);
                 } else {
+                    retryCount++;
 
                     Response response = toSdkResponse(httpResponse);
 
--- a/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -208,10 +208,6 @@ public class HttpClientRequestExecutor implements RequestExecutor {
 
         Assert.notNull(request, "Request argument cannot be null.");
 
-        /*if (requestLog.isDebugEnabled()) {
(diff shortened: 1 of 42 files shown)


extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -238,7 +238,7 @@ public Response executeRequest(Request request) throws RestException {
                 );
             }
 
-            if (retryCount > 0) {
+            if (retryCount > 0 || redirectUri != null) {
```
mrioan: Please add a short comment about what adding this is actually doing and add a the link to issue 539

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -268,7 +268,6 @@ public Response executeRequest(Request request) throws RestException {
                 }
 
                 exception = null;
-                retryCount++;
```
mrioan: I artificially created a timeout and the retryCount is never incremented. What is the rationale to avoid it to be in line 271? Your fix will not work if it just stays here?

dogeared: The retry count will always be incremented unless it's a redirect

mrioan: I just tried it myself. retryCount is in line 284. If there is a timeout, `httpResponse = httpClient.execute(httpRequest);` will throw an Exception (line 273) and it will cause the flow to jump to line 301. Therefore line 284 is never executed and `retryCount` is not incremented

lhazlewood:
It seems like we need to have a test case that asserts the fix is correct.

mrioan:
As @lhazlewood mentioned, it would be great to have an IT that validates that the response time for a redirect is does not pause. Otherwise a Unit Test will still be enough.

mrioan:
Additioally, I would put a link to "https://github.com/stormpath/stormpath-sdk-java/issues/539" so we know how we discovered that (this is an spaghetti algorithm and having info about changes would be good)

mrioan:
Everything looks good to me @dogeared maybe an IT that times and expects the response in, let's say, 800 ms is simple to write. What do you think?

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutorIT.groovy
```diff
@@ -0,0 +1,71 @@
...
+        Callable<Response> callable = new Callable<Response>() {
+
+            @Override
+            Response call() throws Exception {
+                return httpClientRequestExecutor.executeRequest(new DefaultRequest(HttpMethod.GET, verifyUri))
+            }
+        }
+
+        ExecutorService executorService = Executors.newCachedThreadPool()
+        Future<Response> task = executorService.submit(callable)
```
mrioan: Nice! :)

mrioan:
LGTM. You can consider this PR approved once CI passes. Whether you keep the UT is your call, I do not have a preference. Thanks!

lhazlewood:
LGTM

mattjonesorg:
Thanks for the quick turnaround on this!  LGTM

dogeared:
This branch was rebased to master rather than being merged in order to preserve the order of release and subsequent commits.

PR #541, opened 2016-03-24
Add accountStore support to Spring Security

### LINKED ISSUE #513: Authentication in Spring Security does not support specifying an `accountStore`
StormpathAuthenticationProvider#createAuthenticationRequest is missing `inAccountStore(AccountStore)`

```
protected AuthenticationRequest createAuthenticationRequest(Authentication authentication) {
        String username = (String) authentication.getPrincipal();
        String password = (String) authentication.getCredentials();
        return UsernamePasswordRequest.builder().setUsernameOrEmail(username).setPassword(password).build();
    }
```

mraible:
If a 'stormpath.web.accountStore' property is set as a configuration variable, inAccountStore() is used when authenticating.

Fixes #513.


lhazlewood:
Hi Matt,

I think either this PR might be off or my understanding of the problem could be incorrect.

Usually account store resolution is a runtime thing - i.e. what account store to target when a certain tenant is in use.  For example, if I visit https://foo.myapp.com the 'foo' account store should be targeted during authentication attempts, whereas for https://bar.myapp.com, the 'bar' account store should be targeted.

We have an `AccountStoreResolver` to help resolve this at runtime - why the need for the static configuration property?

Thanks for any clarification!

mraible:
@lhazlewood Where is the `accountStoreHref` value stored? This implementation could certainly be wrong, but I did it based on my interpretation of Mario's advice in HipChat:

> In the context of the Servlet and Spring we have a StormpathAccountStoreResolver
> By défault it is disabled
> Which is then used here
> DefaultUsernamePasswordRequestFactory
> but Spring Security does not use this DefaultUsernamePasswordRequestFactory, I am just sharing  how is the non-Spring Security environment finding the AccountStore
> I repeat that by default it is disabled (meaning that the actual instance is a DisabledAccountStoreResolver)
> and we do not have an alternative class
> but in support questions I have replied to users something like this

> By default the account store functionality is disabled. If you want to enable it you can simply specify a Bean called `stormpathAccountStoreResolver`.

> For example, you can do something like this:

```
@Service("stormpathAccountStoreResolver")
public class CustomAccountStoreResolver implements AccountStoreResolver {

    @Value("#{ @environment['stormpath.web.accountStore'] ?: '' }")
    protected String accountStoreHref;

    @Override
    public AccountStore getAccountStore(HttpServletRequest request, HttpServletResponse response) {
        Client client = ClientResolver.INSTANCE.getClient(request);
        if (Strings.hasText(accountStoreHref)) {
            return client.getResource(accountStoreHref, Directory.class);
        }
        return null;
    }
}
```

> In this case, since we are reading a new `stormpath.web.accountStore` property you also need to add something like this to your `application.properties`:

```
stormpath.web.accountStore = https://api.stormpath.com/v1/directories/44GUL2lWkWAOudZPB3ByYy
```

I started to go down the `AccountStoreResolver` path, but then realized it wasn't in any of the stormpath-spring-security module's dependencies:

> So it looks like AccountStoreResolver is an interface in stormpath-sdk-servlet, yet stormpath-spring-security doesn't depend on it.

So I thought I could create a new AccountStoreResolver just for Spring Security.

> StormpathConfiguration has a Client `@Bean`, so I think we can simplify AccountStoreResolver to:

```
public interface AccountStoreResolver {
    AccountStore getAccountStore();
}
```

I'm happy to create this interface and make it work, just need to know where I should get `accountStoreHref` from.

mrioan:
@lhazlewood we actually never had a working `AccountStoreResolver`. We only have a `DisabledAccountStoreResolver` implementation which was introduced in `1.0.RC3`. Besides that, our Spring Security module is handling authentication itself, and it does not even try to use any `AccountStore`. This issue was only trying to have Spring Security been AccountStore-aware. On another issue we will need to implement a working `AccountStoreReolver`. I am reviewing it and I will focus on having Spring Sec using the `AccountStoreReolver` Bean

mrioan:
Just to be in sync. I have re-designed this whole approach. I did great progress but things are not yet in a shareable state.

In summary, I have a new `WebAuthenticationDetailsFactory` which has an `AccountStoreResolver`. The `AccountStore` put into SpringSec's Authorization token as an "authentication detail". This way, the `StormpathAuthenticationProvider` ends up receiving not only the username and password but the account store that it must use for each particular authentication attempt.

jarias:
@mrioan @lhazlewood This looks related or duplicated of #543 that @mzumbadoate is working on. Just FYI

mrioan:
Thanks @jarias #543 is complementary to #541

mraible:
@mrioan Are your changes on a branch or have they already been merged?

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/config/AbstractStormpathSpringSecurityConfiguration.java
```diff
@@ -58,6 +61,12 @@ public AuthenticationTokenFactory stormpathAuthenticationTokenFactory() {
         return new UsernamePasswordAuthenticationTokenFactory();
     }
 
+    @Value("#{ @environment['stormpath.web.accountStore'] ?: '' }")
+    private String accountStoreHref;
```
lhazlewood: Unless I'm missing something, I think we need to remove this since the specific AccountStore used during authentication can change depending on the request.

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
```diff
@@ -354,7 +387,14 @@ protected AuthenticationRequest createAuthenticationRequest(Authentication authe
             throw new AuthenticationServiceException("Login and password required");
         }
 
-        return UsernamePasswordRequests.builder().setUsernameOrEmail(username).setPassword(password).build();
+        UsernamePasswordRequestBuilder requestBuilder = UsernamePasswordRequests.builder().setUsernameOrEmail(username).setPassword(password);
+
+        if (Strings.hasText(accountStoreHref)) {
+            AccountStore accountStore = client.getResource(accountStoreHref, Directory.class);
```
lhazlewood: This is where the `AccountStoreResolver` (or `OrganizationNameKeyResolver`?) would be used to find out which account store is used during authentication.

For example, if I visit https://companyA.myapi.com  the account store would be the `companyA` Organization (organizations - usually used for multi-tenancy - are account stores).  If visiting https://companyB.myapi.com the account store would be the `companyB` organization.

lhazlewood: On second thought, you could have a `TenantAccountStoreResolver` implementation that delegated to the `OrganizationNameKeyResolver` to get the org name key.  Based on that, it could look up the corresponding account store and return that.

Thoughts?

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/config/AbstractStormpathSpringSecurityConfiguration.java
mrioan: I sent an email with detailed information some mins ago, my code is in a different branch: `spring-security-accountstore_dynamic`

extensions/spring/stormpath-spring-security/src/main/java/com/stormpath/spring/security/provider/StormpathAuthenticationProvider.java
mrioan: @lhazlewood I completely refactored this implementation in a different branch: `spring-security-accountstore_dynamic`. This particular concern is address there. Please take a look at an email I sent a few mins ago giving details about it.

lhazlewood:
I *think* I'm caught up on this thread, so let me summarize my current knowledge/understanding:

When the `AccountStoreResolver` interface was created, it was to provide a way for the application developer to inspect the request (in whatever way they wanted) to determine which AccountStore (if any) should be targeted during an authentication attempt.  It is not a requirement to target a specific account store, but definitely recommended in multi-tenant applications to avoid Stormpath's account store iteration logic in the API server, which could be very slow for many (dozens, hundreds) of account stores mapped to an application.

The only practical default implementations that we can provide - that I can think of at least - is a `DisabledAccountStoreResolver` which never assumes a targeted account store (and is our default configured instance), and a yet-to-be-implemented `TenantAccountStoreResolver` implementation that uses the `OrganizationNameKeyResolver` to find out which organization/tenant is currently in scope for a given request.  Once that is discovered, that can be used to return the Organization instance (which implements `AccountStore`), satisfying the `AccountStoreResolver` contract.

App developers are free of course to provide their own implementation of `AccountStoreResolver` if either of our implementations are not sufficient.

Does this make sense?  Does anyone else have a different understanding?

mrioan:
Missing:

1. Add an IT in spring-security-webmvc module (not the example) to test that the authentication object has the account store or something similar that validates that this fix is working. Tests are already available in MinimalStormpathSpringSecurityWebMvcConfigurationIT.
2. I did not thoroughly test everything. We should check that login and logout works ok in Spring and Spring Boot with and without Spring Security
3. My code is not overriding UsernamePasswordAuthenticationFilter, I simply put our new filter before that filter (so we have 2 UsernamePasswordAuthenticationFilter). I am sure this is not the best approach, we should override it somehow

dogeared:
FYI, this PR is still based on the double master merge branch. I pushed up `spring-security-accountstore-micah` a little while ago which has your changes based on a clean master branch.

https://github.com/stormpath/stormpath-sdk-java/compare/spring-security-accountstore-micah

@mraible @mrioan

mraible:
So we have three branches for this now. This one, `spring-security-accountstore`, `spring-security-accountstore-micah` and `spring-security-accountstore_dynamic`. AFAIK, the last one (dynamic) has the most up-to-date code. Based on yesterday's conversation, I believe this one should be closed and we should resume with `spring-security-accountstore_dynamic` when this feature has been re-scheduled.

mrioan:
The code available in this PR + the changes applied by @mrioan are all available in a new branch called `spring-security-accountstore-rebased`. Issue #513 must continue working on that branch

PR #548, opened 2016-03-30
394 removing jetty adding tomcat

### LINKED ISSUE #394: Update all spring and servlet examples to use Tomcat rather than Jetty

mzumbadoate:
resolves #394

### DIFF
--- a/examples/servlet/pom.xml
+++ b/examples/servlet/pom.xml
@@ -152,16 +152,17 @@
                 </executions>
             </plugin>
             <plugin>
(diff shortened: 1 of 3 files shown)


mzumbadoate:
I'm supporting Tomcat 7
There is not Tomcat 8 plugin yet, there's an alternative to support Tomcat 8 using the Cargo plugin, do we wan't this right now?

jarias:
@mzumbadoate The build is failing, but the changes looks good, tomcat7 should fine.

PR #556, opened 2016-04-04
Fixes #423

### LINKED ISSUE #423: Small modifications to the spring boot quickstart
Feedback from Ramesh on the JAVA SDK (Spring Boot Starter) :

> Links (on click) should open the resource in a new tab -> All the links in the documentations, when clicked should Ideally open a new tab on the browser, instead of taking the user away from the documentation in the same tab. Of course, I can just right click on the link and open it in a new tab instead of directly  clicking it, but mostly often than not people tend to directly click on the link, and if that takes the user away from the actual documentation on the same tab, it may cause a lot of inconvenience and we are potentially driving users away from our documentation, and they may never return.

> Get an API Key -> In Point 3 (screenshot below) , shouldn’t there must be a mention of the “Tenant” (in addition to Email and Password)?

> Get an API Key -> In Point 4 (screenshot below), the button is actually in the right side of the page, not in the middle. In either case, the positioning can change again in future, therefore it is not advisable to mention it in the documentation. Just saying "Click the Create API Key or Manage Existing Keys button" should be sufficient. Users should be able find a damn button. :-)

jarias:
- External links should open in a new window
- Add reference to the tenant when loging into the stormpath console
- Remove positional refernce of the api keys management button in the
web console since it can change.

Fixes #423

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -33,4 +33,9 @@
     _kms('//doug1izaerwt3.cloudfront.net/' + _kmk + '.1.js');
   </script>
   <script src="//cdn.optimizely.com/js/225847041.js"></script>
+  <script type="text/javascript" >
+    $(document).ready(function() {
+      $("a[href^='http']").attr('target','_blank');
+    });
+  </script>
 {% endblock %}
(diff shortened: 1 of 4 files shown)


lhazlewood:
LGTM please merge.

PR #558, opened 2016-04-04
Fixes #320

### LINKED ISSUE #320: Update the README.md to reflect new links and references
We have some links to the external spring security and spring security example repos still.

(Those repos refer back here, so nothing's broken, but the README should be updated)

jarias:
Change Spring Security links to reference the new locations for the
plugin and the example

Fixes #320

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ thoughts! support@stormpath.com
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


lhazlewood:
LGTM, please merge

PR #562, opened 2016-04-04
Ensure Stormpath RequestEventListener is invoked when using Spring Security

### LINKED ISSUE #367: Stormpath RequestEventListener does not work when SpringSecurity is enabled
For example, when SS LogoutHandlers are present in the Spring ApplicationContext, our StormpathHttpServletRequest.logout() method is not invoked - which in turn does not invoke our request event publisher.  Need to find another way of ensuring these events are triggered when SS is enabled.

lhazlewood:
Resolves #367


extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebSecurityConfiguration.java
```diff
@@ -54,10 +56,14 @@
     @Qualifier("stormpathAuthenticationProvider")
     protected AuthenticationProvider stormpathAuthenticationProvider; //provided by stormpath-spring-security
 
-    @Autowired(required = false) //required = false when stormpath.web.enabled = false
```
lhazlewood: This line was removed because you can't very well use a StormpathWebSecurityConfiguration class and disable `stormpath.web.enabled` - it wouldn't make any sense.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/CustomRequestEventListenerIT.groovy
```diff
@@ -0,0 +1,68 @@
...
+import static org.springframework.security.test.web.servlet.response.SecurityMockMvcResultMatchers.unauthenticated
+import static org.springframework.security.test.web.servlet.setup.SecurityMockMvcConfigurers.*
+import static org.testng.Assert.*;
+
+/**
+ * @since 1.0.RC9
+ */
+@WebAppConfiguration
+@ContextConfiguration(classes = [CustomRequestEventListenerConfig.class])
+class CustomRequestEventListenerIT extends AbstractTestNGSpringContextTests {
```
dogeared: very nice! We should use this as a model for other ITs...

dogeared:
The actual code LGTM. 

However, I believe all the commits you are dragging in are the result of a double-merge from master. This can happen when someone rebases commits into master and then someone else merges master back into a branch that was originally cut from master.

I don't think it will hurt anything, but it'll make the commit history look gnarly.

I suggest that you back out your commits with one-by-one with a `reset <commit hash>` and stash at each step. Then, you can re-create your branch off master, `stash pop` and commit each one.

lhazlewood:
Closing this PR in favor of #564

PR #563, opened 2016-04-05
Removed build.gradle from tutorials

### LINKED ISSUE #334: Tutorial describes steps using gradle but the root of our SDK lacks a build.gradle file
Therefore, users wanting to use grandle will also need to use maven to build the project before they can play around with the source code of our tutorial.

Solution: make sure our SDK can be built using gradle.

dogeared:
Resolves #334 

until we find a better way to keep them automatically updated.


lhazlewood:
LGTM, let's merge

PR #564, opened 2016-04-05
Ensure Stormpath RequestEventListener is called during failed authentications when using Spring Security

### LINKED ISSUE #367: Stormpath RequestEventListener does not work when SpringSecurity is enabled

lhazlewood:
Resolves #367

### DIFF
--- a/extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathAuthenticationFailureHandler.java
+++ b/extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathAuthenticationFailureHandler.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2016 Stormpath, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.stormpath.spring.config;
+
+import com.stormpath.sdk.authc.AuthenticationRequest;
+import com.stormpath.sdk.lang.Assert;
+import com.stormpath.sdk.servlet.authc.FailedAuthenticationRequestEvent;
+import com.stormpath.sdk.servlet.authc.impl.DefaultFailedAuthenticationRequestEvent;
+import com.stormpath.sdk.servlet.event.RequestEvent;
+import com.stormpath.sdk.servlet.event.impl.Publisher;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.authentication.AuthenticationFailureHandler;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * A simple {@link AuthenticationFailureHandler} implementation that delegates to an actual/delegate handler for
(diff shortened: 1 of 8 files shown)


extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathAuthenticationFailureHandler.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * A simple {@link AuthenticationFailureHandler} implementation that delegates to an actual/delegate handler for
+ * handling logic, but will then send a Stormpath {@link FailedAuthenticationRequestEvent} after delegate invocation.
+ * <p>This enables Stormpath web events to be sent in a Spring Security environment when
+ * encountering a Spring Security {@link AuthenticationException AuthenticationException}.</p>
```
mraible: I believe you can remove the second `AuthenticationException` in this sentence.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/CustomRequestEventListenerIT.groovy
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.setup.MockMvcBuilders
+import org.springframework.web.context.WebApplicationContext
+import org.testng.annotations.BeforeMethod
+import org.testng.annotations.Test
+
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestBuilders.formLogin
+import static org.springframework.security.test.web.servlet.response.SecurityMockMvcResultMatchers.unauthenticated
+import static org.springframework.security.test.web.servlet.setup.SecurityMockMvcConfigurers.*
+import static org.testng.Assert.*;
```
mraible: Do we allow `*` imports in tests? I've noticed that we don't allow them in sdk code.

mraible:
I checked out this branch and ran `examples/spring-security-webmvc` locally. I was able to debug and confirm that `StormpathLogoutHandler#logout` is called when you hit http://localhost:8080/logout. Therefore, :+1:

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathAuthenticationFailureHandler.java
lhazlewood: Not really.  While the link will be generated correctly, the text shown will be the interface fully qualified name in the generated JavaDoc, not just `AuthenticationException`.  JavaDoc links have the first string be the link to the source code location, the 2nd string is the text you want to display.

mrioan:
@lhazlewood code looks good to me. I tested Spring and Spring Boot and in both cases the failed authentication attempt triggered an event. You can considered this approved. Back to you

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/CustomRequestEventListenerIT.groovy
lhazlewood: I think we should since this code isn't run in an application.  I personally hate having to import all my assert methods and easymock methods every. single. time. I need them.  My .02.

extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/StormpathAuthenticationFailureHandler.java
lhazlewood: Ah, I think in this case it will not expand to the fully qualified name because the class is already imported into the current class?

PR #567, opened 2016-04-05
Make Application.getAccountStoreMappings consistent

### LINKED ISSUE #432: Make Application.getAccountStoreMappings consistent
Right now we have:

```
ApplicationAccountStoreMappingList getAccountStoreMappings();
ApplicationAccountStoreMappingList getAccountStoreMappings(Map<String, Object> queryParams);
ApplicationAccountStoreMappingList getApplicationAccountStoreMappings(ApplicationAccountStoreMappingCriteria criteria);
```

Seems like the third one should also be `getAccountStoreMappings`.

We should deprecate `getApplicationAccountStoreMappings` and add `getAccountStoreMappings` that takes an `ApplicationAccountStoreMappingCriteria` as a parameter.

Alternatively, we can make all 3 be `getApplicationAccountStoreMappings`

mraible:
Change getApplicationAccountStoreMappings to getAccountStoreMappings.

Fixes #432.

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/application/Application.java
+++ b/api/src/main/java/com/stormpath/sdk/application/Application.java
@@ -583,36 +583,16 @@ public interface Application extends AccountStoreHolder<Application>, Resource, 
      *
      * @param criteria the criteria to use when performing a request to the collection.
      * @return a paginated list of the application's mapped account stores that match the specified query criteria.
-     * @since 1.0.RC10
+     * @since 1.0.RC9
      */
     ApplicationAccountStoreMappingList getAccountStoreMappings(ApplicationAccountStoreMappingCriteria criteria);
 
     /**
-     * Returns a paginated list of the application's mapped Account stores that also match the specified query
-     * criteria.
-     * The {@link ApplicationAccountStoreMappings ApplicationAccountStoreMappings} utility class is available to help construct
-     * the criteria DSL.  For example:
-     * <pre>
-     * application.getApplicationAccountStoreMappings(ApplicationAccountStoreMappings.criteria()
-     *     .withAccountStore()
-     *     .orderByListIndex();
-     * </pre>
-     * or, if using static imports:
-     * <pre>
-     * import static com.stormpath.sdk.application.ApplicationAccountStoreMappings.*;
-     *
-     * ...
-     *
-     * application.getApplicationAccountStoreMappings(criteria()
-     *     .withAccountStore()
-     *     .orderByListIndex();
-     * </pre>
-     *
      * @param criteria the criteria to use when performing a request to the collection.
      * @return a paginated list of the application's mapped account stores that match the specified query criteria.
      * @since 0.9
      *
-     * @deprecated in 1.0.RC10 and will be removed before 1.0 final. Use {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria)} instead.
+     * @deprecated in 1.0.RC9 and will be removed before 1.0 final. Use {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria)} instead.
      */
     @Deprecated
     ApplicationAccountStoreMappingList getApplicationAccountStoreMappings(ApplicationAccountStoreMappingCriteria criteria);
--- a/api/src/main/java/com/stormpath/sdk/application/Application.java
+++ b/api/src/main/java/com/stormpath/sdk/application/Application.java
@@ -549,7 +549,7 @@ public interface Application extends AccountStoreHolder<Application>, Resource, 
      * </pre>
      * <p/>
      * This is a type-unsafe alternative to the
-     * {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria) getApplicationAccountStoreMappings(accountStoreMappingCriteria)}
+     * {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria) getAccountStoreMappings(accountStoreMappingCriteria)}
      * method, and might be useful when using dynamic languages like Groovy or JRuby.  Users of compiled languages,
      * or those that like IDE-completion, might favor the type-safe method instead.
      *
--- a/api/src/main/java/com/stormpath/sdk/application/Application.java
+++ b/api/src/main/java/com/stormpath/sdk/application/Application.java
@@ -572,7 +572,7 @@ public interface Application extends AccountStoreHolder<Application>, Resource, 
      * </pre>
      * or, if using static imports:
      * <pre>
-     * import static com.stormpath.sdk.account.ApplicationAccountStoreMappings.*;
+     * import static com.stormpath.sdk.application.ApplicationAccountStoreMappings.*;
      *
      * ...
      *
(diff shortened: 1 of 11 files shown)


api/src/main/java/com/stormpath/sdk/application/Application.java
```diff
@@ -566,6 +566,33 @@
      * The {@link ApplicationAccountStoreMappings ApplicationAccountStoreMappings} utility class is available to help construct
      * the criteria DSL.  For example:
      * <pre>
+     * application.getAccountStoreMappings(ApplicationAccountStoreMappings.criteria()
+     *     .withAccountStore()
+     *     .orderByListIndex();
+     * </pre>
+     * or, if using static imports:
+     * <pre>
+     * import static com.stormpath.sdk.account.ApplicationAccountStoreMappings.*;
```
lhazlewood: Is there a reason that ApplicationAccountStoreMappings is in the `com.stormpath.sdk.account` package?  Is this just where it always has been?

It would be better if it is in the `com.stormpath.sdk.applications` package since it has more to do with application configuration than anything account-specific.

mraible: You are correct @lhazlewood - this was a typo in the Javadoc example code. Fixed now.

api/src/main/java/com/stormpath/sdk/application/Application.java
```diff
@@ -584,7 +611,10 @@
      * @param criteria the criteria to use when performing a request to the collection.
      * @return a paginated list of the application's mapped account stores that match the specified query criteria.
      * @since 0.9
+     *
+     * @deprecated in 1.0.RC10 and will be removed before 1.0 final. Use {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria)} instead.
```
mraible: This seems kinda strange to deprecate something in an RC and then remove it a couple of releases later. I'm guessing it's necessary since there's been a lot of releases since 0.9?

lhazlewood: Yeah, that's the only reason.  We won't really be doing this kind of thing after 1.0 if we can help it.

lhazlewood:
LGTM, but I don't think we can merge this until after RC9 is released.

mrioan:
In Organization we have `createOrganizationAccountStoreMapping`. The analogous method in `Application` is `createAccountStoreMapping`. For the sake of homogeneity we should  also rename `createOrganizationAccountStoreMapping` to `createAccountStoreMapping` in the `Organization` interface

mraible:
@mrioan I've made the requested change to `Organization`. After doing so, I tried running `mvn install -DskipITs` in `extensions/httpclient`. It seems strange that changing a method name would cause test failures, so I'm hoping these are related to my environment, not the refactoring.

```bash
Failed tests:
  DirectoryIT.testCreateSamlDirectoryWithAttributeStatementMappingRules:237 » Resource
  DirectoryIT.testCreateSamlDirectoryWithNoAttributeStatementMappingRules:168 » Resource
  ApplicationIT.testDefaultSAMLPolicyAndProvider:1827 expected [true] but found [false]
  HttpClientRequestExecutorIT.testNoWaitOnRedirect:68 » Timeout
  TenantIT.testGetOrganizationWithCriteria:245 expected 
[https://api.stormpath.com/v1/organizations/p7BzSY10phV8HFowwNWWc] but found 
[https://api.stormpath.com/v1/organizations/4RDJygi574xFse1Z0frY74]
```

mrioan:
Hello @mraible, what is the status of this PR? Did you solve the IT issue so I can re-review it?

mraible:
@mrioan The ITs didn't seem to fail in Travis, so I believe they were only a local environment issue. It's ready for re-review.

api/src/main/java/com/stormpath/sdk/application/Application.java
```diff
@@ -549,13 +549,13 @@
      * </pre>
      * <p/>
      * This is a type-unsafe alternative to the
-     * {@link #getApplicationAccountStoreMappings(ApplicationAccountStoreMappingCriteria) getApplicationAccountStoreMappings(accountStoreMappingCriteria)}
+     * {@link #getAccountStoreMappings(ApplicationAccountStoreMappingCriteria) getApplicationAccountStoreMappings(accountStoreMappingCriteria)}
```
mrioan: at the end it still references `getApplicationAccountStoreMappings` rather than `getAccountStoreMappings`

mrioan:
This comment has not been addressed. I do not think there is a need to create a new issue for this. I think we can basically update the issue titile and work on both application and organization in order to reduce overhead

mrioan:
@mraible back to you

api/src/main/java/com/stormpath/sdk/application/Application.java
mraible: Good catch @mrioan! Do you think I should do like I did with `Organization` and remove most of the Javadoc on the deprecated method?

mraible:
@mrioan I changed `Directory#getOrganizationAccountStoreMappings` to `Directory#getAccountStoreMappings` and RC10 to RC9.

mraible:
@mrioan This PR is ready to be reviewed again.

api/src/main/java/com/stormpath/sdk/directory/Directory.java
```diff
@@ -400,12 +400,12 @@
      * {@link com.stormpath.sdk.organization.Organization} and this Directory, you would do that through the
      * {@link com.stormpath.sdk.organization.OrganizationAccountStoreMapping}. This would preserve both this Directory
      * and the {@link com.stormpath.sdk.organization.Organization}. Only the mapping between the two would be deleted.
-     * See {@link #getOrganizationAccountStoreMappings()}.
+     * See {@link #getAccountStoreMappings()}.
```
mrioan: @mraible. I am not sure about this. 

The rationale behind this issue was that concepts were duplicate. `Application#getApplicationAccountStoreMapping` was redundant in the sense that the method name was referencing the class (`Application` and `Application`). The same happened with the `Organization` class having methods duplicating the name. 

In this particular case the class is `Directory` and it has methods to return all the `Organization`s where this directory is being used as an account store. So we do not have redundant concepts here.

Additionally, we will be braking backwards compatibility. 

Therefore, because of these two reasons I think the change in `Directory` should be rolled back.

mrioan:
@mraible I added a new comment. Back to you

mraible:
@mrioan I've reverted the changes to the `Directory` interface and associated classes.

mraible:
It looks like the latest failure was caused by more than one application still being present.

```
Factory method 'stormpathApplication' threw exception; nested exception is 
java.lang.IllegalStateException: A 'stormpath.application.href' property value must be configured 
if you have more than one application registered in Stormpath.
```

mrioan:
@mraible looks good to me. Please make sure Travis pases, you can merge after that. Thanks!

mraible:
Is there a way to "retest again" with a comment? If not, how do I check for leftover applications and remove them?

mrioan:
We changed the tenant a few days ago but I do not have yet access to it. I know Micah has. I'll contact him...

dogeared:
Looks good! Please merge.

mraible:
@dogeared Should I delete the branch as well? Just want to make sure since you created it.

PR #572, opened 2016-04-06
Fix disabled refresh token

### LINKED ISSUE #547: DefaultAccessToken does not contemplate scenario for a disabled Refresh Token
Java SDK RC8.6 fails to validate a valid access token when Refresh token TTL is set to be 0, i.e. no refresh token issue is required.

`Assert.isTrue(Strings.hasText((String) claims.get("rti")));` in `DefaultAccessToken.java` fails to validate the valid token, since rti claim is null due to TTL time set to 0

mrioan:
- Token revocation listener does not check now whether the jwt corresponds to an access token or not. It just always deletes the access token. It however does check whether there is an associated refresh token to be deleted or not.
- When an access token is constructed we do not check inside the jwt any longer. We only check the href now

Resolves #547

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
@@ -17,11 +17,7 @@ package com.stormpath.sdk.impl.oauth;
 
 import com.stormpath.sdk.impl.ds.InternalDataStore;
 import com.stormpath.sdk.lang.Assert;
-import com.stormpath.sdk.lang.Strings;
 import com.stormpath.sdk.oauth.AccessToken;
-import io.jsonwebtoken.Claims;
-import io.jsonwebtoken.JwtException;
-import io.jsonwebtoken.Jwts;
 
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -44,23 +40,9 @@ public DefaultAccessToken(InternalDataStore dataStore, Map<String, Object> prope
...
-                // token *must* have an rti claim to be an access_token
-                // otherwise, it may be a refresh_token trying to be passed off as an access_token
-                Assert.isTrue(Strings.hasText((String) claims.get("rti")));
-            } catch (Exception e) {
-                throw new JwtException("JWT failed validation; it cannot be trusted.");
-            }
-        } else {
             String href = getStringProperty(HREF_PROP_NAME);
             if (href != null) {
                 Assert.isTrue(href.contains("/accessTokens/"), "href does not belong to an access token.");
```
dogeared: Maybe `/accessTokens/` should be a constant defined in this class? Not a show stopper - just a thought

dogeared:
LGTM, pending CI. One non-blocker comment above.

Please merge after CI.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
mrioan: ok!

mrioan:
@dogeared I am really out of ideas. If we need to do a local validation then we cannot contact the backend since the user is looking for speed (1). Additionally `jwts` cannot be distinguished now that we know that an access token will not necessarily have an `rti` claim (when refresh tokens are disabled). So there is no way (that I can think of) to be able to address this issue.

Do you have a workaround in mind? I don't :(

lhazlewood:
I've asked @robertjd to set up a meeting so we can work through this.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java
```diff
@@ -44,23 +42,9 @@ public DefaultAccessToken(InternalDataStore dataStore, Map<String, Object> prope
...
-
-                // token *must* have an rti claim to be an access_token
-                // otherwise, it may be a refresh_token trying to be passed off as an access_token
-                Assert.isTrue(Strings.hasText((String) claims.get("rti")));
-            } catch (Exception e) {
-                throw new JwtException("JWT failed validation; it cannot be trusted.");
-            }
-        } else {
             String href = getStringProperty(HREF_PROP_NAME);
             if (href != null) {
```
dogeared: The indentation is off here.

mrioan:
This is a known issue. Users should not set Refresh Token's TTL to 0. It should be set to 1 ms (`PT0.001S`) so they are not actually disabled but the refresh token will be actually unusable.

PR #574, opened 2016-04-06
Refactored to use new test environments for travis ci.

### LINKED ISSUE #566: Create a new Tenant to avoid conflicts with pure-sage
sometimes there are more active apps apparently  express apps running and this causes the build to crash, so we need to use a different tenant, when that is created remember to update travis.yml api key, secret and any other key

dogeared:
resolves #566 

For Internal reference: https://stormpath.atlassian.net/wiki/display/SDKS/JSDK+Testing+Environments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ jdk:
 - oraclejdk8
 env:
   global:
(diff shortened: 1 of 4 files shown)


mrioan:
@dogeared LGTM. I approved this. Since you are in PTO I will merge it myself. Thanks

PR #593, opened 2016-04-12
Add YAML as a configuration option to servlet plugin

### LINKED ISSUE #571: Support YAML config
Per framework spec [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/configuration.md) and [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/web-config.yaml).

We need to support YAML config in:

* servlet plugin
* Spring (all integrations)
* Spring Boot (all integrations)

mraible:
Add YAML as a configuration option to servlet plugin with SnakeYAML

This implementation logs a warning if a `.yml` file is found and SnakeYAML doesn't exist on the classpath.

NOTE: SnakeYAML doesn't need to be added to stormpath-sdk-servlet's `pom.xml` as a dependency since it's included as a dependency of TestNG.

Fixes #571.


mraible:
This PR is not ready for review yet. I still need to write a test to verify YAML can be used with `stormpath-spring`.

mraible:
It looks like `@PropertySource` does not support YAML, so it's not possible to override properties with YAML in a Spring Test environment. https://jira.spring.io/browse/SPR-13912

I believe it might be possible if we import Spring Boot classes for testing, but I'm not sure that's what we want.

lhazlewood:
A quick note: this config needs to work in servlet-only (no spring) and traditional spring environments (not just spring boot).

In fact, I think the biggest driver is servlet-only: spring and spring boot already have configuration mechanisms.

impl/src/main/java/com/stormpath/sdk/impl/config/YAMLPropertiesSource.java
```diff
@@ -0,0 +1,102 @@
...
+        Assert.notNull(resource, "resource argument cannot be null.");
+        this.resource = resource;
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Map<String, String> getProperties() {
+        try (InputStream in = resource.getInputStream()) {
+            // check to see if file exists
+            if (in != null) {
```
dogeared: What's the scenario in which `in` could be null at this point? if the file doesn't exist, `getInputStream` will throw a `FileNotFoundException` per `FileResource.java`. 

Maybe add a catch with `FileNotFoundException`? Then, you could get rid of the null check, yes?

impl/src/test/groovy/com/stormpath/sdk/impl/config/YAMLPropertiesSourceTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+import org.testng.annotations.Test
+
+import static org.testng.Assert.assertEquals
+import static org.testng.Assert.fail
+import static org.testng.AssertJUnit.assertTrue
+/**
+ * @since 1.0
+ */
+class YAMLPropertiesSourceTest {
```
dogeared: Can you add another test (or tests) that exercise the other cases in `buildFlattenedMap`? Like, when key starts with `[`?

Check out: https://afternoon-oasis-83667.herokuapp.com/com/stormpath/sdk/impl/config/YAMLPropertiesSource.html

mraible: Do we have configuration values that can be maps or collections? If not, maybe it's better to remove the `else if` statements in buildFlattenedMap?

impl/src/main/java/com/stormpath/sdk/impl/config/YAMLPropertiesSource.java
mraible: `Resource#getInputStream` throws `IOException` - see [Resource.java](https://github.com/stormpath/stormpath-sdk-java/blob/master/impl/src/main/java/com/stormpath/sdk/impl/io/Resource.java). I can check for instanceof `FileNotFoundException`, but I still have to catch it.

impl/src/test/groovy/com/stormpath/sdk/impl/config/YAMLPropertiesSourceTest.groovy
dogeared: That's a good point. Check the framework spec - I will too.

impl/src/main/java/com/stormpath/sdk/impl/config/YAMLPropertiesSource.java
dogeared: Yeah, I see your point - can't assume it's a FileResource (even though we know it is). I just hate extra branching if it's not absolutely necessary. But, the interface doesn't guarantee a non-null value, so I guess we need it.

impl/src/main/java/com/stormpath/sdk/impl/config/YAMLPropertiesSource.java
```diff
@@ -0,0 +1,102 @@
...
+                try {
+                    // test to see if Yaml is on the classpath
+                    Class.forName("org.yaml.snakeyaml.Yaml");
```
dogeared: nice! ;)

dogeared:
With test coverage above 95% (whether that's removing unused code or providing more coverage), this PR is 👍

mraible:
@dogeared I added tests for maps and collections. How long does https://afternoon-oasis-83667.herokuapp.com/com/stormpath/sdk/impl/config/YAMLPropertiesSource.html take to update?

dogeared:
This PR is good to go. Test coverage is at 98% now on the YAMLPropertiesSource 👍

PR #601, opened 2016-04-15
Ensure that log messages in catch blocks pass in the exception for complete logging.

### LINKED ISSUE #600: Log more information about the Throwable in HttpClientRequestExecutor
Currently, when there is an exception in executing a request, we only log the message. This results in log messages like:

```
15:15:10.717 [http-nio-8080-exec-27] WARN com.stormpath.sdk.impl.http.httpclient.HttpClientRequestExecutor - Unable to execute HTTP request: blarg.stormpath.io: unknown error
```

In this case, the exception was a `SocketException`, but we lose that information since we are only logging the message from the Throwable.

The fix is to also log the Throwable itself to get the full stacktrace:

```
...
} catch (Throwable t) {
    log.warn("Unable to execute HTTP request: " + t.toString(), t);
}
...
```

dogeared:
resolves #600

### DIFF
--- a/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -155,7 +155,7 @@ public class HttpClientRequestExecutor implements RequestExecutor {
         this.httpClientRequestFactory = new HttpClientRequestFactory();
 
         PoolingClientConnectionManager connMgr = new PoolingClientConnectionManager();
-        if (MAX_CONNECTIONS_PER_ROUTE >= MAX_CONNECTIONS_TOTAL) {
+        if (MAX_CONNECTIONS_TOTAL >= MAX_CONNECTIONS_PER_ROUTE) {
             connMgr.setDefaultMaxPerRoute(MAX_CONNECTIONS_PER_ROUTE);
             connMgr.setMaxTotal(MAX_CONNECTIONS_TOTAL);
         } else {
@@ -163,8 +163,9 @@ public class HttpClientRequestExecutor implements RequestExecutor {
             connMgr.setMaxTotal(DEFAULT_MAX_CONNECTIONS_TOTAL);
 
             log.warn(
-                "connectionMaxTotal ({}) is less than connectionMaxPerRoute ({}). Reverting to defaults: connectionMaxTotal ({}) and connectionMaxPerRoute ({}).",
-                MAX_CONNECTIONS_TOTAL, MAX_CONNECTIONS_PER_ROUTE, DEFAULT_MAX_CONNECTIONS_TOTAL, DEFAULT_MAX_CONNECTIONS_TOTAL
+                "{} ({}) is less than {} ({}). Reverting to defaults: connectionMaxTotal ({}) and connectionMaxPerRoute ({}).",
+                MAX_CONNECTIONS_TOTAL_PROPERTY_KEY, MAX_CONNECTIONS_PER_ROUTE_PROPERTY_KEY,
+                MAX_CONNECTIONS_TOTAL, MAX_CONNECTIONS_PER_ROUTE, DEFAULT_MAX_CONNECTIONS_TOTAL, DEFAULT_MAX_CONNECTIONS_PER_ROUTE
             );
         }
 
--- a/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -83,9 +83,11 @@ public class HttpClientRequestExecutor implements RequestExecutor {
 
     private static final int DEFAULT_MAX_CONNECTIONS_PER_ROUTE = 10;
     private static final String MAX_CONNECTIONS_PER_ROUTE_PROPERTY_KEY = "com.stormpath.sdk.impl.http.httpclient.HttpClientRequestExecutor.connPoolControl.maxPerRoute";
+    private static final String MAX_CONNECTIONS_PER_ROUTE_VALUE;
 
     private static final int DEFAULT_MAX_CONNECTIONS_TOTAL = 20;
     private static final String MAX_CONNECTIONS_TOTAL_PROPERTY_KEY = "com.stormpath.sdk.impl.http.httpclient.HttpClientRequestExecutor.connPoolControl.maxTotal";
+    private static final String MAX_CONNECTIONS_TOTAL_VALUE;
 
     private int numRetries = DEFAULT_MAX_RETRIES;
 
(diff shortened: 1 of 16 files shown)


extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -130,7 +130,8 @@ public HttpClientRequestExecutor(ApiKey apiKey, Proxy proxy, AuthenticationSchem
         } catch (NumberFormatException nfe) {
             log.error(
```
lhazlewood: Unless I'm missing something, this hould be `log.warn`, not error.  An error is better reserved for things that we can't auto-resolve (and since we're re-trying, we are auto-resolving).  We *should* `log.error` when all re-tries fail and we have to throw an exception.

Also, we should avoid String concatenation when forming log messages.  This incurs unnecessary memory overhead and garbage collection even if the log level is disabled.  We should use log message placeholders:

http://www.slf4j.org/faq.html#logging_performance

lhazlewood: Also `System.getProperty(MAX_CONNECTIONS_PER_ROUTE_PROPERTY_KEY)` can be made a class constant so that we don't need to invoke that method at runtime during logging.

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -142,7 +143,8 @@ public HttpClientRequestExecutor(ApiKey apiKey, Proxy proxy, AuthenticationSchem
...
```
lhazlewood: warn / placeholder / constant

dogeared:
Great call on the log message parameterization @lhazlewood. Not only is it more efficient, it makes the code more readable.

impl/src/main/java/com/stormpath/sdk/impl/api/ClientApiKeyBuilder.java
```diff
@@ -104,9 +104,10 @@ protected Properties getDefaultApiKeyFileProperties() {
             Reader reader = createFileReader(DEFAULT_API_KEY_PROPERTIES_FILE_LOCATION);
             props = toProperties(reader);
         } catch (IOException ignored) {
-            log.debug("Unable to find or load default api key properties file [" +
-                      DEFAULT_API_KEY_PROPERTIES_FILE_LOCATION + "].  This can be safely ignored as this is a " +
-                      "fallback location - other more specific locations will be checked.", ignored);
+            log.debug(
+                "Unable to find or load default api key properties file [{}]. This can be safely ignored as this is a fallback location - other more specific locations will be checked.",
```
mrioan: Not a deal breaker but line breaks should be present in long lines. For future reference

dogeared: The idea was to avoid string concatenation in a log message. But, I think that the compiler will optimize two string literals joined with a `+`. Is that true?

lhazlewood: When in doubt, you can wrap log message construction logic in a `if (log.isDebugEnalbed()) { ... }` check.

lhazlewood: And yeah, if you're concatenating two immutable strings that do not depend on a variable, the compiler will inline that call to avoid the runtime overhead.

mrioan:
@dogeared I see in `DefaultAccessTokenRequestAuthenticator#authenticate` and `ResourceRequestAuthenticator#authenticate` that the original Exception is not being passed along. Is it possible to improve that?

mrioan:
@dogeared just one comment. Back to you. Thanks

impl/src/main/java/com/stormpath/sdk/impl/api/ClientApiKeyBuilder.java
dogeared: Cool, thanks. I prefer to use the string concatenation knowing it will optimize to avoid the double "if" check that occurs when using 

```
if (log.isDebugEnabled() { 
   log.debug(...)
}
```

dogeared:
@mrioan - I think the suggestions for `DefaultAccessTokenRequestAuthenticator#authenticate` and `ResourceRequestAuthenticator#authenticate` may be out of scope for this fix. In both cases, they rely on factories that build an exception. Passing along the original exception would require a change in signature to the factories. Perhaps we should create an issue? In the interim, we could certainly log the original exception if you think that would be helpful.

mrioan:
Hey @dogeared, if you think it is within the scope of this issue then please append the original exception message in the log as you suggested (since we cannot easily attach the exception instance). Thanks

dogeared:
@mrioan - I added the logging of original exceptions in `DefaultAccessTokenRequestAuthenticator#authenticate` and `ResourceRequestAuthenticator#authenticate`

extensions/oauth/src/main/java/com/stormpath/sdk/impl/oauth/authc/ResourceRequestAuthenticator.java
```diff
@@ -60,13 +64,18 @@ public OauthAuthenticationResult authenticate(Application application, ResourceA
 
         JwtWrapper jwtWrapper;
 
+        // TODO: When we get rid of Java 6, we can consolidate the two catch blocks below into one
         try {
             jwtWrapper = new JwtWrapper(request.getAccessToken());
             jwtSignatureValidator.validate(jwtWrapper);
         } catch (OAuthSystemException e) {
+            log.warn("Caught Exception: {}. Rethrowing as AccessTokenOauthException", e.getMessage(), e);
```
mrioan: Just for the sake of allowing the user to easily understand when this exception was thrown, could you please say something like `Caught Exception while validating JWT signature: {}. Rethrowing as AccessTokenOauthException` ?

extensions/oauth/src/main/java/com/stormpath/sdk/impl/oauth/authc/ResourceRequestAuthenticator.java
```diff
@@ -60,13 +64,18 @@ public OauthAuthenticationResult authenticate(Application application, ResourceA
...
+
             throw ApiAuthenticationExceptionFactory
                 .newOauthException(AccessTokenOauthException.class, INVALID_ACCESS_TOKEN);
         } catch (InvalidJwtException e) {
+            log.warn("Caught Exception: {}. Rethrowing as AccessTokenOauthException", e.getMessage(), e);
```
mrioan: Same here but since this is an `InvalidJwtException` I think it is better to log it as `info` or `debug` rather than `warn` as this exception is expected to be thrown often when invalid jwts are received

mrioan:
@dogeared just two comments. If you agree with them you can consider this PR approved (assuming Travis passes). Thanks!

dogeared:
@mrioan - I added the verbiage, but left it as `log.warn` so the Exception blocks could be consolidated when we get rid of Java 6. Do you think they should both be `debug` or `info`?

PR #604, opened 2016-04-25
Add JSON as a configuration option to servlet plugin

### LINKED ISSUE #596: Support JSON config
Per framework spec [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/configuration.md) and [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/web-config.yaml).

We need to support JSON config in:

* core sdk
* servlet plugin

Spring and Spring Boot do not need to support JSON as users can use their [default configuration mechanisms](https://github.com/stormpath/stormpath-sdk-java/issues/571#issuecomment-207492992).

mraible:
resolved #596

Add JSON as a configuration option to servlet plugin.

Fixes the servlet part of 596, but does not solve the whole issue.


dogeared:
updated scope of #596 to only be servlet and not core. Core will be addressed in #607.

PR #606, opened 2016-04-26
Check Spring Security authentication on cookie request

### LINKED ISSUE #605: Spring Security is creating a new Authentication object for every pre-authenticated request
Our `SpringSecurityResolvedAccountFilter` is always triggering an Spring Security authentication without first checking if the `SecurityContext` has an `Authentication` object already available. This causes  the `Authentication` object to be re-created and populated on every cookie-authenticated request. Since we do not cache collections then all the `Account`'s `Groups` are re-retrieved from the server during this process. Therefore, many API calls are unnecessary made and pre-authenticated (i.e. cookie-based) requests are taking longer than they should.

The fix is to simply check if `SecurityContextHolder.getContext().getAuthentication() == null` before actually executing a new authentication

mrioan:
Resolves #605


lhazlewood:
LGTM, please merge after a successful build.

Issue #364, opened 2015-11-22
Javascript / Angular Rich Client SPA Support


lhazlewood:
Ensure Servlet/Spring/Spring Boot/Spring Security endpoints are rich-client-compatible per framework spec [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/single-page-apps.md):

* html5 push state mode that will always render a static html file on GETs
* POSTs need to support both form-url-encoded content (server-rendered forms) and JSON (ria/spa)
* 3rd party login providers should display if in the Application's account store mappings

Note: This issue depends on implementing the `application/json` model responses for `/login` and `/register` in [this](https://github.com/stormpath/stormpath-sdk-java/issues/577) issue and [this](https://github.com/stormpath/stormpath-sdk-java/issues/583) issue, respectively.

dogeared:
*Note*: After the `364-spa-support` branch is merged to master, we need to merge `reviewed-blocked_on_364` to master.  `reviewed-blocked_on_364` has a number of PRs merged into it that are dependent on `364-spa-support`.

PR #608, opened 2016-04-26
585 /me endpoint

### LINKED ISSUE #585: Add support for /me endpoint per framework spec
per the framework spec [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/user-context.md).

- add built in support for `/me`
- add property to enable or disable per framework spec
- always respond with `application/json` per framework spec
- support group expansion per framework spec

mzumbadoate:
resolves #585


api/src/main/java/com/stormpath/sdk/account/AccountInfo.java
```diff
@@ -0,0 +1,139 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * Created by mzumbado on 4/18/16.
+ */
+public class AccountInfo {
```
lhazlewood: Why the *Info classes?  Our SDK instances are already backed by JSON - there shouldn't be a need to create intermediate classes.  We can just use the JSON we already have.  Maybe clone the map and remove the properties we don't want to serialize?

mrioan:
Hello @mzumbadoate . Could you please update the title of this issue so it better descibres what is the issue actually about? Thanks

api/src/main/java/com/stormpath/sdk/account/AccountInfo.java
mzumbadoate: @lhazlewood done removing the Info classes

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
```diff
@@ -1,33 +1,64 @@
...
-import java.util.Map;
+import java.util.*;
 
 /**
  * 1.0.RC8
  */
 public class MeController extends AbstractController {
 
+    private boolean expandGroups;
+    private Set<String> accountProperties = new HashSet<String>(Arrays.asList("href", "username", "email", "givenName",
```
lhazlewood: should be static final, no?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
```diff
@@ -1,33 +1,64 @@
...
+            "middleName", "surname", "fullName", "status", "createdAt", "modifiedAt", "emailVerificationToken"));
```
lhazlewood: Actually, after chatting this morning, it would be better to not have this list at all.

Instead, every non Map, array, or collection property of an account should always be included in the rendered output.  I'll show an example below.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
```diff
@@ -1,33 +1,64 @@
...
-        return new DefaultViewModel("stormpathJsonView", model);
+        if (UserAgents.get(request).isJsonPreferred()) {
+            return new DefaultViewModel("stormpath/me", model).setRedirect(false);
+        }
+
+        //otherwise HTML view:
+        return new DefaultViewModel(getNextUri()).setRedirect(true);
+    }
+
+    private Map<String, Object> getModel(AbstractResource abstractResource, Set<String> resourceProperties) {
```
lhazlewood: Based on the comments above, this should be something like:

```java
private Map<String, Object> getModel(AbstractResource abstractResource) {
    Map<String, Object> result = new HashMap<String, Object>();
    for(String propertyName: abstractResource.getPropertyNames()) {
    Object value = abstractResource.getProperty(propertyName);
    if (!(value instanceof Map.class || value instanceof Collection.class)) {
        result.put(propertyName, value);
    }
    return result;
}
```

mzumbadoate: @lhazlewood done, can you check again please

dogeared:
Depends on #364

dogeared:
replaced by #615

PR #627, opened 2016-05-12
Upgrade dependencies to the latest versions

### LINKED ISSUE #559: Ensure that all library dependencies are at latest stable version

mraible:
Remove JUnit references and replace with TestNG. Fix Groovy issues after upgrading to latest version.

Fixes #559.

### DIFF
--- a/examples/quickstart/pom.xml
+++ b/examples/quickstart/pom.xml
@@ -11,11 +11,11 @@
     <packaging>jar</packaging>
 
     <properties>
-        <groovy.version>1.8.8</groovy.version>
+        <groovy.version>2.4.6</groovy.version>
         <jdk.version>1.6</jdk.version>
-        <logback.version>1.1.2</logback.version>
+        <logback.version>1.1.7</logback.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <testng.version>6.8</testng.version>
+        <testng.version>6.9.10</testng.version>
     </properties>
 
     <dependencies>
@@ -84,11 +84,6 @@
                         </goals>
                     </execution>
                 </executions>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.0</version>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/examples/servlet/pom.xml
+++ b/examples/servlet/pom.xml
@@ -34,9 +34,9 @@
     <properties>
         <jdk.version>1.6</jdk.version>
         <jstl.version>1.2</jstl.version>
-        <logback.version>1.1.2</logback.version>
+        <logback.version>1.1.7</logback.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <servlet.version>3.0.1</servlet.version>
+        <servlet.version>3.1.0</servlet.version>
         <skipITs>true</skipITs>
     </properties>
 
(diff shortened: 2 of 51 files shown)


examples/quickstart/pom.xml
```diff
@@ -87,11 +87,6 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
```
dogeared: Why remove this? Is it unnecessary?

examples/servlet/pom.xml
```diff
@@ -34,9 +34,9 @@
     <properties>
         <jdk.version>1.6</jdk.version>
         <jstl.version>1.2</jstl.version>
-        <logback.version>1.1.2</logback.version>
+        <logback.version>1.1.7</logback.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <servlet.version>3.0.1</servlet.version>
+        <servlet.version>3.1.0</servlet.version>
```
dogeared: Hooray!

examples/quickstart/pom.xml
mraible: It's a duplicate. It's already defined on line 56 in this file. https://github.com/stormpath/stormpath-sdk-java/blob/887051575aa31c3922192766451e9394bbe19cee/examples/quickstart/pom.xml#L56

dogeared: gotcha. Thanks.

extensions/spring/stormpath-spring-security/pom.xml
```diff
@@ -64,8 +64,10 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-test</artifactId>
         </dependency>
-
+        <dependency>
```
dogeared: per Les' concern the other day, does this have any impact on end-developers? Would they need to have hamcrest themselves to use Spring Security?

mraible: No, it's just for compiling our tests. Apparently it used to be included in one of the dependencies (maybe TestNG?) and now it isn't.

pom.xml
```diff
@@ -306,6 +295,16 @@
                 <version>${jackson.version}</version>
             </dependency>
             <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
```
dogeared: Is this for servlet? Spring/Spring Boot already has this built-in, right? Is there any impact to Spring and Spring Boot by including it here? For instance, what if we have a later version of this library specified here than Spring uses?

dogeared:
How is the update of `OauthHttpServletRequest` in this PR related to updating the versions?

... back to you @mraible

pom.xml
mraible: This is added in dependencyManagement so the versions are locked down for any projects that include it as a dependency. You'll notice I had to include `jackson-annotations` and `jackson-core` below. Without this, it was pulling in older versions of those dependencies. This section does not specifically include Jackson as a dependency.

mraible:
@dogeared New methods were added to `HttpServletRequest` in Servlet 3.1, so they needed to be implemented in `OAuthHttpServletRequest`.

examples/servlet/pom.xml
jarias: So this means no more Java 6?

dogeared: Yup. We already took that out of what's merged into master.

extensions/spring/stormpath-spring-security/pom.xml
dogeared: 👍

pom.xml
dogeared: got it. Thanks.

dogeared:
LGTM! Please merge.

PR #631, opened 2016-05-17
Add support for /register per the framework spec

### LINKED ISSUE #583: Add support for /register per the framework spec
per framework spec [here](https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md).

- We need to verify that the current HTML responses conform to the spec
- We need to add in the `stormpath.web.register.enabled` property
- We need to add support for `application/json`, including being able return the register model.
- We need to expand on the currently supported properties for specifying form fields. These properties need to be able to be in yaml format as specified in the spec
- We need to support custom fields and have those custom fields populate customData on submit, as specified in the spec

jarias:
Fixes #583

It also includes a refactor of the me controller cause the Account model transformation was shared between register, login and me controllers.


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/impl/DefaultConfig.java
```diff
@@ -137,6 +144,11 @@ public ControllerConfigResolver getChangePasswordControllerConfig() {
     }
 
     @Override
+    public boolean isRegisterAutoLoginEnabled() {
+        return CFG.getBoolean("stormpath.web.register.autoLogin");
```
mraible: This property is not documented in https://github.com/stormpath/stormpath-sdk-java/wiki/1.0-Configuration-Changes-&-Additions-Guide#not-in-specification. Should it be?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/MeFilter.java
```diff
@@ -6,16 +6,14 @@
 import javax.servlet.ServletException;
 
 /**
- * Created by mzumbado on 4/18/16.
+ * @since 1.0
```
mraible: Should be `1.0.0`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultAccountModelFactory.java
```diff
@@ -2,33 +2,87 @@
...
+                            accountMap.put(property, account.getCustomData());
+                        } else {
+                            accountMap.put(property, getResourceProperties((AbstractResource) propertyValue));
+                        }
+                    }
+                } catch (NoSuchMethodException e) {
+                    log.error("Error expanding account property", e);
+                } catch (InvocationTargetException e) {
+                    log.error("Error expanding account property", e);
+                } catch (IllegalAccessException e) {
```
mraible: You could use Java 7's multi-catch feature to simplify this. https://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ValidationException.java
```diff
@@ -0,0 +1,22 @@
+package com.stormpath.sdk.servlet.mvc;
+
+/**
+ * @since 1.0
```
mraible: Should be `1.0.0`

mraible:
I'm not sure why this branch is running against Java 6. https://travis-ci.org/stormpath/stormpath-sdk-java/builds/130928832

Any ideas?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/MeFilter.java
mrioan: 1.0.0

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
```diff
@@ -47,75 +48,48 @@ public FormController() {
...
         this.csrfTokenManager = configResolver.getCsrfTokenManager();
+        this.formFields = configResolver.getFormFields();
+
         this.fieldValueResolver = new ContentNegotiatingFieldValueResolver();
 
         Assert.notNull(this.csrfTokenManager, "csrfTokenManager cannot be null.");
         Assert.notNull(this.fieldValueResolver, "fieldValueResolver cannot be null.");
     }
 
-    public CsrfTokenManager getCsrfTokenManager() {
```
mrioan: Why where this getters/setters removed? Not questioning, just asking in order to understand what was the rationale behind this.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -97,89 +99,16 @@ public boolean isNotAllowedIfAuthenticated() {
         return true;
     }
 
-    public String getForgotLoginUri() {
```
mrioan: Why where this getters/setters removed? Not questioning, just asking in order to understand what was the rationale behind this.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/RegisterErrorModelFactory.java
```diff
@@ -0,0 +1,30 @@
+package com.stormpath.sdk.servlet.mvc;
+
+import com.stormpath.sdk.servlet.i18n.MessageSource;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * @since 1.0
```
mrioan: 1.0.0

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultAccountModelFactory.java
jarias: Yeah this branch is still JDK 6 source level, I guess I could change and let travis fail for Java 6

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
jarias: There where no need, nobody is accessing the controller fields outside the controller. They could probably be even final, once the controller is configured that's it.

jarias:
@mraible I think this was cut out of master a long time ago that's why Java 6 is still in the travis build and the language level still at 6 also.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/impl/DefaultConfig.java
jarias: Its on the spec https://github.com/stormpath/stormpath-framework-spec/blob/master/registration.md

```
stormpath:
  web:
    register:
      enabled: true
      uri: "/register"
      nextUri: "/"
      autoLogin: false
...
```

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/form/DefaultField.java
```diff
@@ -15,6 +15,8 @@
  */
 package com.stormpath.sdk.servlet.form;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
```
mrioan: I do not think you really need a builder here. Just use the fluid interface: change each method so they do not return void but the current `Field` instance instead.

You will then do: 
```
Field field = new DefaultField();
field.setName(name).setValue(value).setPlaceholder(placeholder);
```

impl/src/main/java/com/stormpath/sdk/impl/account/DefaultAccount.java
```diff
@@ -97,7 +97,7 @@
     static final CollectionReference<RefreshTokenList, RefreshToken> REFRESH_TOKENS =
             new CollectionReference<RefreshTokenList, RefreshToken>("refreshTokens", RefreshTokenList.class, RefreshToken.class);
 
-    static final Map<String, Property> PROPERTY_DESCRIPTORS = createPropertyDescriptorMap(
+    public static final Map<String, Property> PROPERTY_DESCRIPTORS = createPropertyDescriptorMap(
```
mrioan: None of our resources provide public access to this Map. I do not think we should be changing this. What are you trying to achieve with this? Can you think of a different way of doing what you need without exposing this map?

mrioan:
# UAT

- spring-security-webmvc: I added this:

This custom field is displayed as `!CustomField!`.

- Servlet Plugin

Idem, this custom field is displayed as `!CustomField!`.

- spring-boot-webmvc

- This custom field is displayed as `??CustomField_en_US??`
- After logout, the message displayed in the login screen says: !stormpath.web.login.form.status.logout!

mrioan:
@jarias first round of review is ready. Back to you. Thanks!

impl/src/main/java/com/stormpath/sdk/impl/account/DefaultAccount.java
jarias: Its use here I need a way to get all the default account properties without an actual account instance.

```java
private Map<String, Object> getCustomData(HttpServletRequest request, Form form) {
        //Custom fields are either declared as form fields which shouldn't not be account fields or through a customField attribute
        Map<String, Object> result = new LinkedHashMap<String, Object>();

        for (Field field : form.getFields()) {
            //Field is not part of the default account properties then is a custom field
            if (DefaultAccount.PROPERTY_DESCRIPTORS.get(field.getName()) == null && !field.getName().equals(csrfTokenManager.getTokenName())) {
                result.put(field.getName(), field.getValue());
            }
        }

        Object customData = fieldValueResolver.getAllFields(request).get("customData");
        if (customData instanceof Map) {
            //noinspection unchecked
            result.putAll((Map<? extends String, ?>) customData);
        } //If not a map ignore, the spec doesn't cover this case

        return result;
    }
```

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/form/DefaultField.java
jarias: That breaks the Java Bean pattern, I don't really like it, is there a reason for doing it this way?

jarias:
@mrioan regarding the messages both the label and the placeholder should be i18n properties, at least is what we do now, i18n is not part of the spec is something the Java SDK does since is very common in Java web development.

One solution to this would be to use the raw text if the property is not found, but that means resolving the i18n keys in the backend instead of the JSP or thymeleaf template what do you think?

jarias:
@mrioan ready to re-review

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/form/DefaultField.java
mrioan: What do you mean by "breaks the Java Bean pattern"?. I mentioned about removing the Builder because I do not see any benefit for adding it. It is not really constructing/building anything in particular which requires a particular logic that the Builder can enforce. I simply see a builder being used as a setter. That's why I proposed to simply use a fluent interface solely. I am basically thinking towards avoiding "killing a mosquito with a tank".

impl/src/main/java/com/stormpath/sdk/impl/account/DefaultAccount.java
mrioan: Before calling `account.getCustomData().putAll(getCustomData(req, form));` you are going one property at a time in order to set the (fixed) `Account` properties. So that piece of code already knows the "set" of properties belonging to the account. Anything outside that set goes into `CustomData`. Therefore you should basically put those into custom data. That way you can:1

1. Rollback `PROPERTY_DESCRIPTORS` to non-public
2. Remove the import and use of `DefaultAccount` (we should try to avoid importing things from the `impl` package; it should always be the last option, not the first one)

mrioan:
@jarias Can you please create a new issue to work on both `??CustomField_en_US??` and `!CustomField!.`? Put milestone `1.0.0` and reference this issue please.

Back to you. Thanks

PR #642, opened 2016-05-23
Remove the ROLE_ prefix for HttpServletRequest#isUserInRole(String)

### LINKED ISSUE #325: Stormpath groups fail in multiple places as roles but work fine as authorities
An expected prefix is set for all role names - by default this is set to `ROLE_`, see:

    org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter.rolePrefix

Stormpath role names are always of the form `https://api.stormpath.com/v1/groups/...` so if you call:

    HttpServletRequest request = ...
    request.isUserInRole("https://api.stormpath.com/v1/groups/...") 

You'll eventually hit:

    org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestWrapper.isGranted(String)`.

This will tack the role prefix onto whatever you passed into `isUserInRole(...)` and it ends up trying to match `ROLE_https://...` against the Stormpath roles assigned to the user.

You can fix this in your `StormpathWebSecurityConfigurerAdapter` subclass by setting the prefix to the empty string (null isn't allowed):

    http.servletApi().rolePrefix("");

**Question:** should Stormpath be handling this for you in its setup logic rather than requiring anyone who wants to use `isUserInRole(...)` to set this up?

Note: if you Google setting role prefix to the empty string you see various people advising that this is a "bad idea" - but in the Stormpath context maybe the reasoning behind this doesn't apply.

It might actually be nice to set the role prefix to `https://api.stormpath.com/v1/groups/` then you'd just have to use the last component of a group's path when specifying a group - but I guess that would have knock on consequences in other situations like using e.g. `@PreAuthorize("hasRole(@roles.USER)")` etc.

mrioan:
Resolves #325

### DIFF
--- a/examples/spring-security-webmvc/pom.xml
+++ b/examples/spring-security-webmvc/pom.xml
@@ -33,6 +33,7 @@
         <servlet.version>3.1.0</servlet.version>
         <slf4j.version>1.7.21</slf4j.version>
         <spring.version>4.2.3.RELEASE</spring.version>
+        <spring.security.version>4.1.0.RELEASE</spring.security.version>
         <tomcat.version>7.0.59</tomcat.version>
     </properties>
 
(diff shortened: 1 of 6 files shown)


mraible:
LGTM 👍

dogeared:
please merge!

PR #653, opened 2016-05-25
Change YamlPropertiesSource to set lists as a comma-delimited value

### LINKED ISSUE #650: Array type properties not working correctly
the produces property and the fieldOrder property for login and register are describe as lists in the spec.

In the servlet extension the defaults are written as:

```
stormpath.web.produces[0] = application/json
stormpath.web.produces[1] = text/html
```

But the ExpressionConfigReader lacks the ability to read this and return a proper list

Also in Spring we asume "," separated values to describe lists which might break users custom configuration

```java
@Value("#{ @environment['stormpath.web.produces'] ?: 'application/json, text/html' }")
protected String produces;
```
In this case we asume is a single string which is later parse using `MediaType.parseMediaTypes(..)`

For the fieldOrder we delegate the work to Spring to get a `String[]` but I don't know if it supports the prop[INDEX] notation.

see AbstractSpringControllerConfigResolver.java

```java
Arrays.asList(env.getProperty("stormpath.web." + getControllerKey() + ".form.fields.fieldOrder", String[].class, getDefaultFieldOrder()))
```

mraible:
Change YAML processing to create a single comma-delimited property rather than several index-based properties.

Fixes #650.

### DIFF
--- a/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
+++ b/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
@@ -36,8 +36,7 @@ stormpath.web.refreshTokenCookie.httpOnly = true
 stormpath.web.refreshTokenCookie.secure =
 stormpath.web.refreshTokenCookie.path =
 stormpath.web.refreshTokenCookie.domain =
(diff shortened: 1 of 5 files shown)


dogeared:
👍  LGTM - please merge

PR #660, opened 2016-05-26
Allow custom fields' label and placeholder values to be plain text

### LINKED ISSUE #645: Figure out how to handle or document i18n message in form field config
Currently the Java web integration provides i18n support, so in order to configure extra fields for /login or /register the values for label and placeholder should be 18n keys otherwise the value won't render correctly in the view

Example:

```
```
Would render in servlet `!CustomField!` and `??CustomField_en_US??` on Spring

Options:

1. Clearly document that the values should be i18n keys in the i18n.properties file.
2. Do a fallback behavior if the key is not there use the value as it is so the example above would work without any extra effort by the user.

mraible:
This detects if an i18n key exists for the label and placeholder values. If a translated key is not found, the key is rendered to the screen.

NOTE: This is a branch from `364-spa-support-merged`. That branch should probably be merged to `master` first so you can see the changes to fix #645 in this one.


dogeared:
Note: Closing this one. I created #666 to `target 364-spa-support-merged` to make it clear what was worked on.

Issue #652, opened 2016-05-25
JSON Form model displays label and placeHolder fields as i18n keys


jarias:
Need to resolve the property if the response is JSON, for HTML this is done in the actual view.

This code is not yet in master only in 364-spa-support-merged, that's why the ticket is **blocked**

```
⇒  curl -v -s -H "Accept: application/json" http://localhost:8080/register | jq
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> GET /register HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.48.0
> Accept: application/json
> 
< HTTP/1.1 200 OK
< Server: Apache-Coyote/1.1
< Content-Type: application/json
< Transfer-Encoding: chunked
< Date: Wed, 25 May 2016 16:48:28 GMT
< 
{ [1786 bytes data]
* Connection #0 to host localhost left intact
{
  "form": {
    "fields": [
      {
        "name": "givenName",
        "label": "stormpath.web.register.form.fields.givenName.label",
        "placeholder": "stormpath.web.register.form.fields.givenName.placeholder",
        "required": true,
        "type": "text"
      },
      {
        "name": "surname",
        "label": "stormpath.web.register.form.fields.surname.label",
        "placeholder": "stormpath.web.register.form.fields.surname.placeholder",
        "required": true,
        "type": "text"
      },
      {
        "name": "email",
        "label": "stormpath.web.register.form.fields.email.label",
        "placeholder": "stormpath.web.register.form.fields.email.placeholder",
        "required": true,
        "type": "email"
      },
      {
        "name": "password",
        "label": "stormpath.web.register.form.fields.password.label",
        "placeholder": "stormpath.web.register.form.fields.password.placeholder",
        "required": true,
        "type": "password"
      }
    ]
  },
  "accountStores": [
    {
      "href": "https://api.stormpath.com/v1/directories/c3Rwcf17S95IMsG6MTvGq",
      "provider": {
        "href": "https://api.stormpath.com/v1/directories/c3Rwcf17S95IMsG6MTvGq/provider",
        "providerId": "google",
        "clientId": "584731295650-uh34m1td8ck396q3rebtlb5tspevrcea.apps.googleusercontent.com"
      },
      "name": "google"
    },
    {
      "href": "https://api.stormpath.com/v1/directories/1rdrV2qwHnHrxhIuPwmTaa",
      "provider": {
[...]

PR #666, opened 2016-05-31
645 customfields i18n passthru

### LINKED ISSUE #645: Figure out how to handle or document i18n message in form field config

dogeared:
Resolves #645 

This detects if an i18n key exists for the label and placeholder values. If a translated key is not found, the 
key is rendered to the screen.

NOTE: This is a branch from 364-spa-support-merged. That branch should probably be merged to master first so you can see the changes to fix #645 in this one.


mraible:
To UAT this change for servlet, add the following to `examples/servlet/src/main/resources/stormpath.properties`:

```
stormpath.web.login.form.fields.CustomField.type = text
stormpath.web.login.form.fields.CustomField.label = Custom Field Login
stormpath.web.login.form.fields.CustomField.enabled = true
stormpath.web.login.form.fields.CustomField.visible = true
stormpath.web.login.form.fields.CustomField.required = true
stormpath.web.login.form.fields.CustomField.placeholder = Custom Field Login

stormpath.web.register.form.fields.CustomField.label = Custom Registration Field
stormpath.web.register.form.fields.CustomField.placeholder = Custom Registration Field
```

Then start the example using `mvn tomcat7:run` and confirm these fields show up on the login and registration screens. Rather than showing `!Custom Login Field!` and `!Custom Registration Field!`, it should show `Custom Login Field` and `Custom Registration Field`.

To UAT for Spring, add the above keys to `examples/spring-webmvc/src/main/resources/application.properties` and run the same tests.

dogeared:
On Register, the custom field is created as custom data.

On Login, is entering the custom field with the correct vaue supposed to be enforced? Currently, you must enter something in the field, but it can be anything - not necessarily the value stored in custom data.

mraible:
I didn't write the custom field logic, I believe @jarias did. This PR is only for the i18n of the custom field keys.

lhazlewood:
I don't understand this change.

1.  Where do I specify the i18n key and its values across different languages?
2.  How do I ensure that my custom fields' `label` and `placeholder` values reflect i18n keys (not values)?

mraible:
@lhazlewood If you specify a key for the label or placeholder, it will be used. If the keys don't exist in your i18n bundle, the raw values will be passed through rather than showing that they weren't found.

lhazlewood:
Ah, ok thanks - that wasn't obvious to me in the example shown in your [previous comment](https://github.com/stormpath/stormpath-sdk-java/pull/666#issuecomment-223020271).  Thanks!

extensions/spring/boot/stormpath-thymeleaf-spring-boot-starter/src/main/resources/templates/stormpath/login.html
```diff
@@ -51,13 +51,16 @@
...
+                            <th:block th:if="${field.type != 'hidden'}"
+                                      th:with="label=${#messages.msg(field.label)},placeholder=${#messages.msg(field.placeholder)}">
                                 <label th:class="${#lists.isEmpty(accountStores) ? 'col-sm-4' : 'col-sm-12'}"
-                                       th:text="#{${field.label}}">|Field label|</label>
+                                       th:text="${#strings.startsWith(label, '??') ? field.label : label}">|Field label|</label>
 
                                 <div th:class="${#lists.isEmpty(accountStores) ? 'col-sm-8' : 'col-sm-12'}">
                                     <input th:name="${field.name}" th:value="${field.value}" th:type="${field.type}"
-                                           th:placeholder="#{${field.placeholder}}"
+                                           th:placeholder="${#strings.startsWith(placeholder, '??') ? field.placeholder : placeholder}"
```
lhazlewood: This logic shouldn't be in a view - it should be in the controller.

extensions/spring/boot/stormpath-thymeleaf-spring-boot-starter/src/main/resources/templates/stormpath/register.html
```diff
@@ -44,12 +44,14 @@
...
-
+                                <!--/* Check to see if key is found for custom fields. If not, just display key. */-->
+                                <!--/* https://github.com/stormpath/stormpath-sdk-java/issues/645 */-->
+                                <th:block th:if="${field.type != 'hidden'}"
+                                          th:with="label=${#messages.msg(field.label)},placeholder=${#messages.msg(field.placeholder)}">
+                                    <label class="col-sm-4" th:text="${#strings.startsWith(label, '??') ? field.label : label}">|Field label|</label>
                                     <div class="col-sm-8">
                                         <input th:name="${field.name}" th:value="${field.value}" th:type="${field.type}"
-                                               th:placeholder="#{${field.placeholder}}"
+                                               th:placeholder="${#strings.startsWith(placeholder, '??') ? field.placeholder : placeholder}"
```
lhazlewood: Logic should be in the controller - a view shouldn't have to do discovery to see if i18n translation is valid or not.

extensions/spring/boot/stormpath-thymeleaf-spring-boot-starter/src/main/resources/templates/stormpath/login.html
mraible: I thought about this, but figured this was the easiest solution since the translation is happening in the view, not the controller.

mraible:
@lhazlewood @dogeared I updated this PR to do the translation logic in `FormController`. This also fixes #652. To UAT 652, you can run the following curl commands (`brew install jq` to get pretty JSON output):

`curl -v -s -H "Accept: application/json" http://localhost:8080/login | jq`
`curl -v -s -H "Accept: application/json" http://localhost:8080/register | jq`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
```diff
@@ -149,18 +149,28 @@ protected Form createForm(HttpServletRequest request, HttpServletResponse respon
...
                 String val = fieldValueResolver.getValue(request, clone.getName());
-                if (clone.getName() == "password" && retainPassword) {
+                if (clone.getName().equals("password") && retainPassword) {
                     clone.setValue(val);
                 } else {
                     clone.setValue(val);
                 }
+                // #645: Allow unresolved i18n keys to pass through for labels and placeholders
+                String label = i18n(request, clone.getLabel());
+                if (!label.startsWith("!")) {
```
jarias: Why not change the message source to accept a default and hide this logic away.

mraible: That's a good idea. I'll make this change tomorrow.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
```diff
@@ -149,18 +149,28 @@ protected Form createForm(HttpServletRequest request, HttpServletResponse respon
...
-        List<Field> fields = new ArrayList<Field>();
+        List<Field> fields = new ArrayList<>();
 
         for (Field templateField : formFields) {
             Field clone = templateField.copy();
 
             if (clone.isEnabled()) {
                 String val = fieldValueResolver.getValue(request, clone.getName());
-                if (clone.getName() == "password" && retainPassword) {
+                if (clone.getName().equals("password") && retainPassword) {
```
dogeared: maybe yoda syntax here? 

`if (retainPassword && "password".equals(clone.getName()))`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
mraible: @jarias I added a new method to `MessageSource` and `DefaultMessageSource` that allows you to pass in a defaultMessage. It wasn't as easy as Spring's `MessageSource` because we're using varargs. Spring uses [Object[] for arguments](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/MessageSource.html) in its implementation. 

Because of this, I had to do some duplication of code. You can see my commit at https://github.com/stormpath/stormpath-sdk-java/pull/666/commits/29bbd97a2181342c5fa0e71469e8d27d824cc3c5. Please let me know if you have any suggestions for improvement.

jarias:
@mraible LGTM

PR #623, opened 2016-05-10
Fixes #577

dogeared:
Add JSON and produces support for /login

PR #672, opened 2016-06-01
Updated LogoutController as well as the classes that use it to support the produces property for content negotiation.

### LINKED ISSUE #671: Logout is busted in the presence of Content Negotiation
produces property is null when constructing the `LogoutController` and thus causes an NPE in `DefaultContentNegotiationResolver.getContentType`

dogeared:
resolves #671

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/LogoutFilter.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/LogoutFilter.java
@@ -27,7 +27,7 @@ public class LogoutFilter extends ControllerFilter {
 
     @Override
     protected void onInit() throws ServletException {
-        LogoutController c = new LogoutController();
+        LogoutController c = new LogoutController(getConfig());
         c.setLogoutInvalidateHttpSession(getConfig().isLogoutInvalidateHttpSession())
             .setLogoutNextUri(getConfig().getLogoutControllerConfig().getNextUri());
         setController(c);
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/LogoutFilter.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/LogoutFilter.java
@@ -27,7 +27,7 @@ public class LogoutFilter extends ControllerFilter {
 
     @Override
     protected void onInit() throws ServletException {
-        LogoutController c = new LogoutController(getConfig());
+        LogoutController c = new LogoutController(getConfig().getLogoutControllerConfig(), getConfig().getProducesMediaTypes());
         c.setLogoutInvalidateHttpSession(getConfig().isLogoutInvalidateHttpSession())
             .setLogoutNextUri(getConfig().getLogoutControllerConfig().getNextUri());
         setController(c);
--- a/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/mvc/ControllerTest.groovy
+++ b/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/mvc/ControllerTest.groovy
@@ -1,7 +1,25 @@
+/*
+ * Copyright 2016 Stormpath, Inc.
+ *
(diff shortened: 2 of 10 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/LogoutFilter.java
```diff
@@ -27,7 +27,7 @@
 
     @Override
     protected void onInit() throws ServletException {
-        LogoutController c = new LogoutController();
+        LogoutController c = new LogoutController(getConfig());
```
mrioan: In general Les does not like the idea of passing configuration objects around. Configuration should only be used to get data when initializing things. Passing the config object can cause a runtime issue if an object tries to get some value at some point which is not present or is invalid. We would want things to break during boot time rather than runtime. So I agree with Les that passing configuration objects should be discouraged.

lhazlewood: Yeah, the idea is that dependency injection is the best (non-coupling) way to configure things.  For example, Spring configuration is totally different than our Servlet config stuff - ideally a `Config` object shouldn't exist in a Spring environment at all.

This allows configuration to occur via whatever mechanism is best for a given environment, without 'forcing' the `Config` concept on all environments.

dogeared: We may need to do a larger refactor after we get over this hump. Here, I am following the same pattern used for the `LoginController`. Without this, the `LogoutController` is broken because it doesn't pick up on the `produces` property without the config.

dogeared: Just confirming that at commit `9d46c52`, @mrioan established the pattern of passing in passing the `Config` into `LoginController` and calling `super` with it. If we want to refactor, that's fine, but currently `LogoutController` is broken and this PR fixes it using the same pattern. I think we should let this PR get merged and then refactor the pattern across the controllers for servlet and spring if we want to.

mrioan: @dogeared I agree that having that the LoginController should also be changed. I am working on that as well.

BTW, I did nor really add that into the `LoginController` myself. [This is the real commit]( (https://github.com/stormpath/stormpath-sdk-java/pull/623/commits/7bc2611e74e12ef76d74e1d1254fa263b1ed1ea6#r63070186)

dogeared:
@mrioan - do you want to incorporate fixing the (currently broken) LogoutController as well? Not sure how to proceed. We can't UAT with the Logout function simply not working. Thoughts?

mrioan:
@dogeared yes, I am working on that. I think it is already working for me. I need to UAT a little bit more to be sure

dogeared:
👍  Great! Let me know when it's done and we can just kill this PR.

extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/mvc/ControllerTest.groovy
```diff
@@ -142,13 +160,39 @@ class ControllerTest {
...
+        expect(controllerConfigResolver.getLocaleResolver()).andReturn localeResolver
+        expect(controllerConfigResolver.getControllerKey()).andReturn "someKey"
+
+        expect(controllerConfigResolver.getNextUri()).andReturn "/"
+        expect(controllerConfigResolver.getUri()).andReturn "/"
+        expect(controllerConfigResolver.getMessageSource()).andReturn messageSource
+        expect(controllerConfigResolver.getLocaleResolver()).andReturn localeResolver
+        expect(controllerConfigResolver.getControllerKey()).andReturn "someKey"
+
+        replay controllerConfigResolver
```
dogeared: nice

dogeared:
LGTM. Please merge after CI. 👍

PR #686, opened 2016-06-07
401 fix

### LINKED ISSUE #681: Returning `400` when should be returning `401`
In cases where there is a failure in bearer token authorization, the status code returned should be `401`.

Steps to reproduce:

1) get an access and refresh token

```
http -f POST localhost:8080/oauth/token \
grant_type=password username=<valid email> password=<valid password>
```

2) save the returned `refresh_token` to a local variable for later use

```
REFRESH_TOKEN=<refresh_token returned from step 1>
```

3) hit the `/me` endpoint with the refresh token as a bearer token

```
http localhost:8080/me Authorization:"Bearer $REFRESH_TOKEN"
```

This should fail and it does. However, the status returned currently is `400` and it should be `401`.

dogeared:
#resolves #681


dogeared:
closing for wrong target branch

Issue #674, opened 2016-06-02
refresh token can be used as access_token cookie


dogeared:
Steps to reproduce:

1) create a token pair in the usual way:

```
http -f POST \
http://localhost:8080/oauth/token \
Origin:http://localhost:8080 \
grant_type=password username=<valid email> password=<valid password>
```

2) save the tokens in a local variables for ease of use

```
REFRESH_TOKEN=<refresh_token from previous step>
ACCESS_TOKEN=<access_token from previous step>
```

3) Sanity check:

**fails as expected**
```
➥ http localhost:8080/me Authorization:"Bearer $REFRESH_TOKEN"
HTTP/1.1 400 Bad Request
Cache-Control: no-store
Connection: close
Content-Type: application/json;charset=UTF-8
Date: Thu, 02 Jun 2016 21:39:40 GMT
Pragma: no-cache
Server: Apache-Coyote/1.1
Transfer-Encoding: chunked

{
    "error": "invalid_client",
    "error_description": "access_token is invalid."
}
```

**succeeds as expected**
```
➥ http localhost:8080/me Authorization:"Bearer $ACCESS_TOKEN"
HTTP/1.1 200 OK
Cache-Control: no-store, no-cache
Cache-Control: no-store
Content-Language: en-US
Content-Type: application/json;charset=UTF-8
Date: Thu, 02 Jun 2016 21:42:23 GMT
Pragma: no-cache
Server: Apache-Coyote/1.1
Transfer-Encoding: chunked

{
    "account": {
        "createdAt": "2016-03-04T06:29:48.506Z",
        "email": "mt@me.com",
        "fullName": "m t",
        "givenName": "m",
        "href": "https://api.stormpath.com/v1/accounts/3qYGmIzUhxPKYO28kN0IbR",
        "middleName": null,
        "modifiedAt": "2016-05-24T02:14:01.780Z",
        "status": "ENABLED",
        "surname": "t",
        "username": "mt@me.com"
    }
}
```

4) Here's the bug (This should fail just like the first part of step 3):

```
➥ http localhost:8080/me Cookie:"access_token=$REFRESH_TOKEN"
HTTP/1.1 200 OK
Cache-Control: no-store, no-cache
Cache-Control: no-store
Content-Language: en-US
Content-Type: application/json;charset=UTF-8
Date: Thu, 02 Jun 2016 21:40:52 GMT
Pragma: no-cache
Server: Apache-Coyote/1.1
[...]

Issue #675, opened 2016-06-02
/me endpoint results in NPE for certain integrations


dogeared:
using this to test:

```
http localhost:8080/me
```

**PASS**: `Servlet` (returns: `HTTP/1.1 401 Unauthorized`)

**FAIL**: `Spring WebMVC` (NPE, DefaultAccountModelFactory:75; MeController:50)

**FAIL**: `Spring Security Spring WebMVC` (NPE, DefaultAccountModelFactory:75; MeController:50)

**FAIL**: `Spring Boot WebMVC` (NPE, DefaultAccountModelFactory:75; MeController:50)

**FAIL**: `Spring Security Spring Boot WebMVC` (NPE, DefaultAccountModelFactory:75; MeController:50)

mrioan:
Mimicking what the Servlet is doing we should be returning `401` when not authenticated for all the integrations

PR #687, opened 2016-06-07
674 Refresh token shall not be used as an access token in cookie

mrioan:
Resolves https://github.com/stormpath/stormpath-sdk-java/issues/674 https://github.com/stormpath/stormpath-sdk-java/issues/675

UAT is simple, just follow the comments in #674 and #675

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
@@ -23,7 +23,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.util.List;
 
 /**
- * 1.0.RC8
+ * @since 1.0.RC8
  */
 public class MeController extends AbstractController {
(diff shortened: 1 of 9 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
```diff
@@ -23,7 +23,7 @@
 import java.util.List;
 
 /**
- * 1.0.RC8
+ * @since 1.0.RC8
```
dogeared: This should really be: `@since 1.0.0.`

We haven't yet supported the `/me` endpoint.

dogeared:
LGTM. Please merge!

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/MeController.java
mrioan: good

PR #702, opened 2016-06-13
If X-Stormpath-Agent header is set, its value should be prepended to the User Agent string

### LINKED ISSUE #478: Ensure that User Agent string prepends client substring if header is present
related to #287 

Per: https://github.com/stormpath/stormpath-framework-spec/pull/33

If the `X-Stormpath-Agent` header is set, its value should be *prepended* to the User Agent string.

The "sweet spot" for this is: `DefaultDataStore.applyDefaultRequestHeaders`. We can check for the `X-Stormpath-Agent` header on the `request` and, if present, prepend to this call:

```
request.getHeaders().set("User-Agent", USER_AGENT_STRING);
```

mraible:
Fixes #478.

I'd love to write a test for this, but I'm not sure what the best way to do that is.


mraible:
Closing to re-branch of master after big merge.

PR #704, opened 2016-06-14
Ensure that User Agent string prepends X-Stormpath-Agent header if it exists

### LINKED ISSUE #478: Ensure that User Agent string prepends client substring if header is present

mraible:
Fixes #478.

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
@@ -28,6 +28,7 @@ import com.stormpath.sdk.impl.ds.cache.WriteCacheFilter;
 import com.stormpath.sdk.impl.error.DefaultError;
 import com.stormpath.sdk.impl.http.CanonicalUri;
 import com.stormpath.sdk.impl.http.HttpHeaders;
+import com.stormpath.sdk.impl.http.HttpHeadersHolder;
 import com.stormpath.sdk.impl.http.MediaType;
 import com.stormpath.sdk.impl.http.QueryString;
 import com.stormpath.sdk.impl.http.QueryStringFactory;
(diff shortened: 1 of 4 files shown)


impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -571,8 +573,14 @@ private Response execute(Request request) throws ResourceException {
 
     protected void applyDefaultRequestHeaders(Request request) {
         request.getHeaders().setAccept(java.util.Collections.singletonList(MediaType.APPLICATION_JSON));
-        request.getHeaders().set("User-Agent", USER_AGENT_STRING);
-        if (request.getHeaders().getContentType() == null){
+        List<String> stormpathAgents = request.getHeaders().get(STORMPATH_AGENT);
```
jarias: I don't think this is the right implementation, the X-Stormpath-Agent according to spec comes form our client side integrations like Angular, so when they do a request to the backend web integration (express or spring for example) they pass that header so we can added if present to the request that would be send to stormpath, here you are looking at the request we build to talk to the stormpath API not the one that comes from a web integration.

Se for example the express impl https://github.com/stormpath/express-stormpath/blob/10252505b7c4602e52771fb65656b9e73937f598/lib/stormpath.js#L89 their middleware checks a request from client and adds it to what express user agent string.

Or the Go impl again it captures client requests and looks for that header so it can attach it to the user agent string when an API call is made https://github.com/jarias/stormpath-sdk-go/blob/38f5a99365b337ec1e7a2ae944df2195f2a7fdae/web/web.go#L167

@lhazlewood, @mrioan or @dogeared correct me if I'm wrong.

mrioan:
hello @mraible I tried with `Spring Boot WebMVC` but I do not see the `X-Stormpath-Agent` ever reaching the `DefaultDataStore#applyDefaultRequestHeaders`. I do see the request coming to that method, but the header is not there. Did you try it yourself in some particular way?

I used both `httpie` and `curl` like this:

```
http -f POST http://localhost:8080/oauth/token Origin:http://localhost:8080 X-Stormpath-Agent:XXXXX grant_type=password username=EMAIL password=PASSWD
```
```
curl -L -X POST -H "Content-Type: application/x-www-form-urlencoded" -H "X-Stormpath-Agent: XXXXX" -H "Accept: application/json" -u "APIKEY_ID:APIKEY_SECRET" -d "grant_type=client_credentials" http://localhost:8080
```

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
mraible: I see what you're saying @jarias. I implemented this here based on @dogeared's advice in https://github.com/stormpath/stormpath-sdk-java/issues/478#issue-133070087. If `DefaultDataStore.applyDefaultRequestHeaders` is not the spot for this, where should should it go?

jarias: Maybe a new web filter is needed for this.

mraible:
I implemented a `ThreadLocal` to store the headers. Please review again @jarias and @mrioan.

jarias:
@mraible LGTM

Issue #711, opened 2016-06-16
examples/spring-security-spring-boot-webmvc-bare-bones does not load stormpath assets


mraible:
![screen shot 2016-06-16 at 10 46 41 am](https://cloud.githubusercontent.com/assets/17892/16125198/c96f1e4c-33af-11e6-9868-e976664f6646.png)

lhazlewood:
I'm pretty sure this is a handlermapping order issue.  I'm encountering this same exact issue today in another project.

I believe it will be solved by modifying the `AbstractStormpathWebMvcConfiguration` implementation's `stormpathHandlerMapping()` method by adding a couple of mappings:

```java
mappings.put(`/assets/js/stormpath.js`, aResourceHttpRequestHandlerInstance);
mappings.put(`/assets/cs/stormpath.css`, aResourceHttpRequestHandlerInstance);
```

lhazlewood:
I had to do this as a workaround:

```java
@Configuration
public class WebConfig extends WebMvcConfigurerAdapter {

    @Value("${stormpath.web.handlerMapping.order}")
    private int stormpathHandlerMappingOrder = -300;

    @Override
    public void addResourceHandlers(ResourceHandlerRegistry registry) {
        registry.setOrder(stormpathHandlerMappingOrder);
        registry.addResourceHandler("/assets/js/stormpath.js").addResourceLocations("classpath:/META-INF/resources/");
        registry.addResourceHandler("/assets/css/stormpath.css").addResourceLocations("classpath:/META-INF/resources/");
    }
}
```

mraible:
This only seems to happen when SAML is enabled in `application.properties`:

```
stormpath.web.callback.enabled=true
```

The CSS and JS is loaded fine otherwise. 

![image](https://cloud.githubusercontent.com/assets/17892/16402501/8d851dbe-3c88-11e6-8826-9efa6f7530e6.png)

PR #713, opened 2016-06-16
Make local SAML integration work like ID site

### LINKED ISSUE #438: Make local SAML integration behave more like ID Site
Currently, `stormpath.web.idSite.enabled = true` enables SAML globally for the application.

However, the Stormpath application may be configured with a mix of SAML, social and other types of account stores.

Rather than have the setting, we should auto-discover what directory types (both directly and through organizations) are mapped to the Stormpath Application and render the view accordingly.

Here's an example from an ID Site app that has two different SAML directories mapped  in the same Application:

image

mraible:
Fixed #438.

# To UAT:

1. Get access to remote-ridge tenant on api.stormpath.com. 
2. Generate an API Key and download to ~/.stormpath/apiKey.remote-ridge.properties
3. Configure Maven or Tomcat to start with the following environment variables:
```
STORMPATH_APPLICATION_HREF=https://api.stormpath.com/v1/applications/6vFT4ARfWCmuHVV8VktjTo
STORMPATH_API_KEY_FILE=/Users/<username>/.stormpath/apiKey.remote-ridge.properties
```
  
## For Spring MVC:
1. In examples/spring-webmvc/src/main/resources/application.properties, add properties to disable idSite and enable SAML:

    ```
    stormpath.web.idSite.enabled = false
    stormpath.web.callback.enabled = true
    ```

2. Start examples/spring-webmvc and go to http://localhost:8080. Click the "Login" button and verify that 1) the SAML buttons show up and 2) clicking on each SAML buttons takes you to its respective SAML provider. Login if you have credentials.
3. Change `stormpath.web.callback.enabled` to `false` and restart. Verify that there are no Facebook or SAML buttons are displayed.

## For Spring Boot:
Same steps as Spring MVC, but use spring-boot-webmvc.

## For Spring Security:
Same steps as Spring MVC, but use spring-security-spring-boot-webmvc, spring-security-spring-boot-webmvc-bare-bones and spring-security-webmvc.

NOTE: I found #711 when testing the spring-security-spring-boot-webmvc-bare-bones example.

## For Servlet:
1. Create examples/servlet/src/main/resources/stormpath.properties, and add the following properties to disable idSite and enable SAML:

    ```
    stormpath.web.idSite.enabled = false
    stormpath.web.callback.enabled = true
    ```

2. Start examples/servlet and go to http://localhost:8080. Click the "Login" button and verify that 1) the SAML buttons show up and 2) clicking on each SAML buttons takes you to its respective SAML provider. Login if you have credentials.
3. Change `stormpath.web.callback.enabled` to `false` and restart. Verify that there are no Facebook or SAML buttons are displayed.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/DefaultFilterChainResolverFactory.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/DefaultFilterChainResolverFactory.java
@@ -129,6 +129,10 @@ public class DefaultFilterChainResolverFactory implements Factory<FilterChainRes
         String unauthorizedUrlPattern = cleanUri(unauthorizedUrl);
         boolean unauthorizedChainSpecified = false;
 
+        String samlUrl = "/saml";
+        String samlUrlPattern = cleanUri(samlUrl);
+        boolean samlChainSpecified = false;
+
         String meUrl = config.getMeUrl();
         String meUrlPattern = cleanUri(meUrl);
         boolean meChainSpecified = false;
@@ -259,6 +263,12 @@ public class DefaultFilterChainResolverFactory implements Factory<FilterChainRes
                     if (!chainDefinition.contains(filterName)) {
                         chainDefinition += Strings.DEFAULT_DELIMITER_CHAR + filterName;
                     }
+                } else if (uriPattern.startsWith(samlUrlPattern)) {
+                    samlChainSpecified = true;
+                    String filterName = DefaultFilter.saml.name();
+                    if (!chainDefinition.contains(filterName)) {
+                        chainDefinition += Strings.DEFAULT_DELIMITER_CHAR + filterName;
+                    }
                 } else if (uriPattern.startsWith(meUrlPattern)) {
                     meChainSpecified = true;
 
@@ -301,6 +311,9 @@ public class DefaultFilterChainResolverFactory implements Factory<FilterChainRes
         }
         if (!accessTokenChainSpecified) {
             fcManager.createChain(accessTokenUrlPattern, DefaultFilter.accessToken.name());
+        }
+        if (!samlChainSpecified) {
+            fcManager.createChain(samlUrlPattern, DefaultFilter.saml.name());
         }
         if (!meChainSpecified) {
             fcManager.createChain(meUrlPattern, "authc," + DefaultFilter.me.name());
--- a/impl/src/main/java/com/stormpath/sdk/impl/idsite/IdSiteClaims.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/idsite/IdSiteClaims.java
@@ -32,6 +32,7 @@ public class IdSiteClaims extends DefaultClaims {
     public static final String SHOW_ORGANIZATION_FIELD = "sof";
     public static final String USE_SUBDOMAIN = "usd";
     public static final String SP_TOKEN = "sp_token";
+    public static final String ACCOUNT_STORE_HREF = "ash";
 
     public static final String ACCESS_TOKEN = "accessToken";
     public static final String JWT_REQUEST = "jwtRequest";
@@ -109,6 +110,15 @@ public class IdSiteClaims extends DefaultClaims {
 
     public IdSiteClaims setSpToken(String spToken) {
         setValue(SP_TOKEN, spToken);
+        return this;
+    }
+
+    public String getAccountStoreHref() {
+        return getString(ACCOUNT_STORE_HREF);
+    }
+
+    public IdSiteClaims setAccountStoreHref(String accountStoreHref) {
+        setValue(ACCOUNT_STORE_HREF, accountStoreHref);
         return this;
     }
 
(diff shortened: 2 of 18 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/DefaultFilterChainResolverFactory.java
```diff
@@ -129,6 +129,10 @@ protected FilterChainResolver createInstance() throws ServletException {
         String unauthorizedUrlPattern = cleanUri(unauthorizedUrl);
         boolean unauthorizedChainSpecified = false;
 
+        String samlUrl = "/saml";
```
mraible: I don't think that this needs to be configurable since this same path is hard-coded in [stormpath.js](https://github.com/stormpath/stormpath-sdk-java/blob/438-local-saml/extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js#L65).

impl/src/main/java/com/stormpath/sdk/impl/idsite/IdSiteClaims.java
```diff
@@ -32,6 +32,7 @@
     public static final String SHOW_ORGANIZATION_FIELD = "sof";
     public static final String USE_SUBDOMAIN = "usd";
     public static final String SP_TOKEN = "sp_token";
+    public static final String ACCOUNT_STORE_HREF = "ash";
```
dogeared: Wondering if we should have a SAMLClaims class that extends DefaultClaims or extends IdSiteClaims if there's a lot of overlap?

If ash is *not* used in the context of IdSite, then I think we should for sure.

dogeared: Also, the `@JsonSerialize` annotation on line 26 is deprecated. We should update it to: `@JsonInclude(JsonInclude.Include.NON_NULL)`

impl/src/main/java/com/stormpath/sdk/impl/provider/ProviderClaims.java
```diff
@@ -0,0 +1,77 @@
...
+package com.stormpath.sdk.impl.provider;
+
+import io.jsonwebtoken.impl.DefaultClaims;
+
+/**
+ * @since 1.0.0
+ */
+public class ProviderClaims extends DefaultClaims {
+
+    public static final String CALLBACK_URI = "cb_uri";
```
dogeared: @mrioan @mraible  - I pulled the common claims up into `ProviderClaims`. However, I am not sure this is entirely correct. For instance, in the context of `Organizations`, there is code in `IdSiteClaims` that may still be relevant to `SamlClaims`. Would you please take a look?

dogeared:
@mraible @mrioan - I abstracted out the common code for `IdSiteClaims` and created `SamlClaims`. Check my question above.

All UAT is passing.

One thing to note is that `logout` is not working properly. However, it is not working properly for `IdSite` either. I will create a separate issue to track that. It should not hold up this PR.

impl/src/main/java/com/stormpath/sdk/impl/saml/SamlClaims.java
```diff
@@ -0,0 +1,19 @@
+package com.stormpath.sdk.impl.saml;
```
mraible: Missing copyright?

impl/src/main/java/com/stormpath/sdk/impl/saml/SamlClaims.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.stormpath.sdk.impl.provider.ProviderClaims;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class SamlClaims extends ProviderClaims {
```
mraible: @since 1.0.0

impl/src/main/java/com/stormpath/sdk/impl/provider/ProviderClaims.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2015 Stormpath, Inc.
```
mraible: 2016?

impl/src/test/groovy/com/stormpath/sdk/impl/saml/SamlClaimsTest.groovy
```diff
@@ -0,0 +1,43 @@
+package com.stormpath.sdk.impl.saml
```
mraible: Is copyright needed for tests?

impl/src/main/java/com/stormpath/sdk/impl/saml/SamlClaims.java
dogeared: Added copyright.

impl/src/test/groovy/com/stormpath/sdk/impl/saml/SamlClaimsTest.groovy
dogeared: yup. I've added it. Thanks.

mraible:
LGTM 👍

dogeared:
In my UAT I found that the SAML buttons do not behave as on might expect in the presence of Organizations. However, this is not well defined, so it should not hold up this issue.

PR #715, opened 2016-06-17
714 spring redirect login

### LINKED ISSUE #714: Spring Security is not redirecting to login when an endpoint requires authentication
It seems we introduced an error when we switched from Spring Security's own form login to ours here https://github.com/stormpath/stormpath-sdk-java/issues/682.

This bug is only present in master

Will be tested via TCK: https://github.com/stormpath/stormpath-framework-tck/issues/273

mrioan:
Resolves #714


lhazlewood:
@mrioan I think this is good to go to be merged - thanks!

PR #722, opened 2016-06-20
Fixes #719

### LINKED ISSUE #719: Spring Security WebMVC login is broken
Currently the Spring Security WebMVC example doesn't load the login page, no exceptions are thrown it just show a blank screen with a 404 status code.

spring-security-webmvc is the failing example

jarias:
Missing context path in some URLs inside JSPs


mraible:
LGTM 👍

PR #723, opened 2016-06-20
Fixes #684

### LINKED ISSUE #684: isJsonPreferred() must be replaced by ContentNegotationResolver
In `FormController#onErrorSubmit` there is a line doing 

```
if (UserAgents.get(request).isJsonPreferred()) {
``` 

This does not seem right, we should be using the `ContentNegotationResolver` in order to know the kind of expected response.

There might be other places where `isJsonPreferred ` or `isHtmlPreferred()` is used. Please search and improve them all.

jarias:
isJsonPreferred() must be replaced by ContentNegotationResolver


mraible:
@jarias Should `DefaultUserAgent#isJsonPreferred` be deprecated, along with `DefaultUserAgent#isHtmlPreferred`?

jarias:
@mraible no cause its use by DefaultContentNegotiationResolver which is the proper class to use for resolving the content type

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
```diff
@@ -206,7 +205,7 @@ protected ViewModel onErrorSubmit(HttpServletRequest request, HttpServletRespons
 
         List<ErrorModel> errors = toErrors(request, form, e);
 
-        if (UserAgents.get(request).isJsonPreferred()) {
+        if (isJsonPreferred(request, response)) {
```
mrioan: Inside `isJsonPreferred(...)` I see that `ContentNegotiationResolver.INSTANCE` is executed every time. We could simply have a `ContentNegotiation` instance as a property of this class which is used there. Could you please change that? Thanks

mrioan:
@jarias. I just added a comment. Back to you. Thanks

jarias:
@mrioan PR updated

mraible:
LGTM 👍

PR #725, opened 2016-06-21
Updated example and tutorial poms to be included in deployment

### LINKED ISSUE #724: Publish example and tutorial jars

dogeared:
Resolves #724

### DIFF
--- a/examples/pom.xml
+++ b/examples/pom.xml
@@ -141,9 +141,6 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
                     <version>2.7</version>
-                    <configuration>
-                        <skip>true</skip>
-                    </configuration>
                 </plugin>
             </plugins>
         </pluginManagement>
--- a/examples/quickstart/pom.xml
+++ b/examples/quickstart/pom.xml
@@ -12,7 +12,7 @@
 
     <properties>
         <groovy.version>2.4.6</groovy.version>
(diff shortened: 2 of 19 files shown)


mrioan:
LGTM. Merge please. Thanks! @dogeared

dogeared:
Thanks @mrioan. One question: The root of `examples` and `tutorials` still has `<skip>true</skip>` will this cause the sub-projects to be skipped? If so, I'll need to update the parent poms as well.

mraible:
@dogeared Yes, this should cause the sub-projects will be skipped. However, even if you change `examples/pom.xml` to remove `<skip>true</skip>`, most of the sub-projects override it, so just changing the parent won't work.

* examples/spring
* examples/spring-boot
* examples/spring-boot-default
* examples/spring-boot-webmvc
* examples/spring-security-spring-boot-webmvc
* examples/spring-security-spring-boot-webmvc-bare-bones
* examples/spring-security-webmvc
* examples/spring-webmvc

Also, we should change `<jdk.version>1.6</jdk.version>` to `<jdk.version>1.7</jdk.version>` in all the examples.

dogeared:
Right, @mraible - this PR updates all the sub-projects. I will update the root for each of them as well as the java version. Thanks.

dogeared:
@mrioan and/or @mraible - please take another look. I've update jdk version and parent poms for examples and tutorials.

mraible:
LGTM 👍

PR #727, opened 2016-06-22
Change SpringCache to use put instead of putIfAbsent

### LINKED ISSUE #636: SpringCache using putIfAbsent for the put operation
com.stormpath.spring.cacheSpringCache is using org.springframework.cache.Cache.putIfAbsent in the implementation of the put method. This causes problems with values not being updated in cache at all (because they already exist in the cache).

On the contrary, the DefaultCache implementation (which uses a simple Map), uses a simple Map.put operation.

Shouldn't the SpringCache implementation be using simple put as well ?

mraible:
Fixes #636.


jarias:
LGTM @mraible

PR #736, opened 2016-06-28
Refactor access token to use "stt" claim to verify token type

### LINKED ISSUE #696: Refactor access token and refresh token code to take advantage of new type header
JWT's from the backend now include `stt` (stormpath token type) claim in the header.

We are currently checking for the presence of the `rti` claim in the body to determine if a token is an access token.

We now have explicit types in the JWT header to determine type.

`stt` will be `access` for an access token and `refresh` for a refresh token

mraible:
Fixes #696.


dogeared:
Hey @mraible - This looks good to me.

One question:

Do you think we need to do any internal checking in:

```
OAuthRefreshTokenRequestAuthentication refreshGrantRequest = 
OAuthRequests.OAUTH_REFRESH_TOKEN_REQUEST.builder()
  .setRefreshToken(val)
  .build();
```

and the corresponding `OAUTH_ACCESS_TOKEN_REQUEST` to ensure that the JWT passed in is a valid refresh token (or access token - as the case may be)?

mraible:
@dogeared I think it makes sense for Authenticators that do local validation (e.g. `DefaultOAuthBearerRequestAuthenticator`). However, this seems to be the only authenticator that does local validation, so not sure.

dogeared:
ok. @mraible - LGTM. I am going to merge since you are off today.

PR #746, opened 2016-06-29
Ensured that callback is is true by default. 

### LINKED ISSUE #738: Social login should be enabled by default
According to [spec](https://github.com/stormpath/stormpath-framework-spec/blob/master/example-config.yaml) the callback property should be enabled by default.

That callback property is for both Social Login and SAML so we need 2 changes:

1. In `AbstractStormpathWebMvcConfiguration` the `stormpath.web.callback.enabled` should be `true`
2. Additionally, since that property defines behavior not only for SAML but also for Social Buttons then the variable name should be changed from `samlEnabled` to `callbackEnabled`

dogeared:
Resolves #738 

Rename samlEnabled to callbackEnabled.


jarias:
@dogeared LGTM lets just wait for CI

PR #750, opened 2016-07-01
Use hasAuthority instead of hasRole for expressions

### LINKED ISSUE #365: Support role based authorization in WebSecurityConfigurerAdapter
This should work and currently does not:

```
@Configuration
public class SpringSecurityWebAppConfig extends WebSecurityConfigurerAdapter {
    @Override
    protected void configure(HttpSecurity http) throws Exception {
        http
            .apply(stormpath()).and()
            .authorizeRequests()
                .antMatchers("/admin").hasRole("@roles.GROUP_ADMIN")
                .antMatchers("/new_book").hasRole("@roles.GROUP_USER")
                .antMatchers("/upvote").hasRole("@roles.GROUP_USER")
                .antMatchers("/").permitAll();
    }
}
```

NOTE: I tried plugging in the fully qualified group URL directly and that does not work either.

mraible:
resolves #365

Use hasAuthority instead of hasRole for expressions. Also removed unnecessary configuration and renamed `Roles` class to `Groups`.


mraible:
Updated tutorial to match in PR https://github.com/stormpath/stormpath.github.io/pull/268.

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/SpringSecurityWebAppConfig.java
```diff
@@ -36,8 +36,6 @@ protected void configure(HttpSecurity http) throws Exception {
         // We want to restrict access to one path and leave all other paths open.
         http
             .apply(stormpath()).and()
-            .authorizeRequests()
-            .antMatchers("/restricted").fullyAuthenticated()
-            .antMatchers("/**").permitAll();
+            .authorizeRequests();
```
dogeared: The original code was there to demonstrate the syntax, even though all paths are locked down by default.

I think  you still need to permit everything but `/restricted` for this example to work properly. I will double check as a UAT step.

examples/spring-security-webmvc/src/main/java/com/stormpath/spring/examples/SpringSecurityWebAppConfig.java
```diff
@@ -94,8 +94,6 @@ protected void configure(HttpSecurity http) throws Exception {
 
         http
             .apply(stormpath()).and()
-            .authorizeRequests()
-            .antMatchers("/restricted").fullyAuthenticated()
-            .antMatchers("/**").permitAll();
+            .authorizeRequests();
```
dogeared: same comment as above.

tutorials/spring-boot/03-spring-security-refined/src/main/java/com/stormpath/tutorial/config/Groups.java
```diff
@@ -23,11 +23,11 @@
  * @since 1.0.RC5
  */
 @Component
-public class Roles {
+public class Groups {
     public final String USER;
 
     @Autowired
-    public Roles(Environment env) {
+    public Groups(Environment env) {
```
dogeared: There may be an impact on the tutorial docs. I will check on this...

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/SpringSecurityWebAppConfig.java
mraible: As far as functionality goes, I found that there was no difference between this and what was previously there.

tutorials/spring-boot/03-spring-security-refined/src/main/java/com/stormpath/tutorial/config/Groups.java
mraible: I updated the tutorial docs as well. See https://github.com/stormpath/stormpath.github.io/pull/268

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/SpringSecurityWebAppConfig.java
dogeared: @mraible:

For the spring default example (and the others below), you should be able to get to `/` *without* having to authenticate. With the changes you've made, going to the front door immediately sends you to the `/login` view.

This is the expected behavior for the `bare bones` example, but not for these.

At a minimum, we need:

```
http
    .apply(stormpath()).and()
    .authorizeRequests()
    .antMatchers("/").permitAll();
```

The way it was before was just to demonstrate the syntax.

Here's what you should see when you hit the front door of this example:

image

mraible: OK, should I just restore the previous config so there's less changes?

dogeared: I think so. We don't *technically* need the `.antMatchers("/restricted").fullyAuthenticated()` line (since everything is locked down by default), but I think it makes it more clear what's going on.

mraible:
@dogeared I reverted the security config back to what it was. Please review again.

dogeared:
LGTM - please merge after CI completes.

dogeared:
You can update the tutorial doc here: `extensions/spring/boot/docs/source/tutorial.rst`

dogeared:
LGTM - please merge after CI completes. 👍

PR #754, opened 2016-07-06
753: renamed 'stormpath.web.oauth2.validationStrategy' and replaced...

lhazlewood:
…it with spec-compliant 'stormpath.web.oauth2.password.validationStrategy'.  Also updated the default value of this property to be 'local' (web config was correct, spring config was not).


extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
```diff
@@ -169,7 +169,6 @@ stormpath.web.sendVerificationEmail.view = /sendVerificationEmail
 # The context-relative path where an authenticated (already logged in) user will be redirected when attempting to
 # access a url they are not allowed to access.
 stormpath.web.unauthorized.uri = /unauthorized
-stormpath.web.oauth2.validationStrategy = stormpath
```
lhazlewood: This diff doesn't tell the whole story - this file already has an existing `stormpath.web.oauth2.password.validationStrategy` property.  This PR just removes this extraneous/incorrect property.

jarias:
LGTM @lhazlewood

PR #761, opened 2016-07-12
Make SAML and ID Site properties and handlers spec compliant

### LINKED ISSUE #748: make SAML and ID Site properties and handlers spec compliant
ID Site is clearly spelled out here: https://github.com/stormpath/stormpath-framework-spec/blob/master/idsite.md

Dealing with SAML is less clear in the framework spec, but here's what came out of the discussion with the team:

1) There should be a single callback uri defined that handles the result responses from ID Site and from direct SAML interactions. This is the `stormpath.web.callback.uri` property and it defaults to `/stormpathCallback`.

2) If the `stormpath.web.idSite.enabled` property is `true` and the  `stormpath.web.callback.enabled` is `false`, this is a config error that should be caught on startup.

3) If stormpath.web.idSite.enabled` is false and `stormpath.web.callback.enabled` is false AND there are SAML directories mapped to the application, that is also a configuration error. This does not have to be caught on startup. The SAML buttons should not be rendered, BUT there should be an error message in the logs about the config error.

4) The handler for `/StormpathCallback` will determine if it's a callback from ID Site or from a direct SAML interaction and will process accordingly. This information should be clear from the incoming JWT.

mraible:
Fixes #748.


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -96,8 +104,24 @@ public boolean isNotAllowedIfAuthenticated() {
     @Override
     protected void appendModel(HttpServletRequest request, HttpServletResponse response, Form form, List<ErrorModel> errors,
                                Map<String, Object> model) {
-        if (samlLoginEnabled) {
-            model.put("accountStores", accountStoreModelFactory.getAccountStores(request));
+
+        List<AccountStoreModel> accountStores = accountStoreModelFactory.getAccountStores(request);
+        // 748: If stormpath.web.idSite.enabled is false and stormpath.web.callback.enabled is false AND
+        // there are SAML directories mapped to the application, that is a configuration error.
+        if (!idSiteEnabled && !callbackEnabled && accountStores.size() > 0) {
```
dogeared: UAT fails here.

`accountStoreModelFactory.getAccountStores(request)` returns *all* account stores, not just SAML account stores.

Steps to reproduce:

1) put the following in `application.properties` in the Spring Boot Default example

    ```
    stormpath.web.idSite.enabled = false
    stormpath.web.callback.enabled = false
    ```

2) Connect to this application in `remote-ridge`: https://api.stormpath.com/v1/applications/6vFT4ARfWCmuHVV8VktjTo (It has a number of directories mapped to it, but NO SAML directories)

3) fire up the app and attempt to login

You will see the WARM message:

`ID Site is disabled and callbacks are disabled, yet this application has SAML directories. Please enable callbacks or remove SAML directories.`

I think the solution is to see if any of the account stores are SAML and set a boolean to true if so.

mraible: @dogeared How can I tell if an account store is SAML?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -96,10 +107,27 @@ public boolean isNotAllowedIfAuthenticated() {
     @Override
     protected void appendModel(HttpServletRequest request, HttpServletResponse response, Form form, List<ErrorModel> errors,
                                Map<String, Object> model) {
-        if (samlLoginEnabled) {
-            model.put("accountStores", accountStoreModelFactory.getAccountStores(request));
+
+        // 748: If stormpath.web.idSite.enabled is false and stormpath.web.callback.enabled is false AND
+        // there are SAML directories mapped to the application, that is a configuration error.
+        if (!idSiteEnabled && !callbackEnabled) {
+            List<AccountStoreModel> accountStores = samlAccountStoreModelFactory.getAccountStores(request);
```
dogeared: Moved this inside the if since it's an expensive operation.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -96,10 +107,27 @@ public boolean isNotAllowedIfAuthenticated() {
...
+            if (accountStores.size() > 0) {
+                String errorMsg = "ID Site is disabled and callbacks are disabled, yet this application has SAML directories. Please enable callbacks or remove SAML directories.";
+                log.warn(errorMsg);
+                if (errors == null) {
+                    errors = new ArrayList<>();
+                }
+                // only add to errors on GET, not POST
+                if (request.getMethod().equals(HttpMethod.GET.name())) {
```
dogeared: Is the expectation that an error would show on the page? Because it doesn't in my UAT.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/provider/DefaultSamlProviderModelFactory.java
```diff
@@ -33,7 +33,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class DefaultProviderModelFactory implements AccountStoreModelFactory {
+public class DefaultSamlProviderModelFactory implements AccountStoreModelFactory {
```
dogeared: I renamed this class as it *only* returns SAML account stores.

extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/impl/DefaultConfigFactoryTest.groovy
```diff
@@ -124,6 +124,17 @@ class DefaultConfigFactoryTest {
         assertTrue config.isMeEnabled()
     }
 
+    @Test(expectedExceptions = IllegalArgumentException, enabled = false)
```
dogeared: @mraible - this should be a test or be removed. Any reason to keep it here?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -96,10 +107,27 @@ public boolean isNotAllowedIfAuthenticated() {
...
+                    errors.add(ErrorModel.builder().setStatus(HttpServletResponse.SC_OK).setMessage(errorMsg).build());
+                }
+            }
         }
 
+        List<AccountStoreModel> accountStores = accountStoreModelFactory.getAccountStores(request);
```
dogeared: Took out the `if` test here because we should *always* show social buttons if social providers are mapped to the application, whether or not `callbackEnabled` is true or false.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
mraible: No, you said you didn't want to show the user an error. This is for testing mostly.

extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/impl/DefaultConfigFactoryTest.groovy
mraible: @dogeared Good catch, I'll enable it and fix any issues I find.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
dogeared: 👍

PR #768, opened 2016-07-15
Don't perform costly authentication on every request

### LINKED ISSUE #529: SpringSecurityResolvedAccountFilter performs a costly authentication on every request
While the purpose of SpringSecurityResolvedAccountFilter is clear to me, I have a feeling it should not produce a new Authentication every time it runs. Each authentication means reading the Groups the user is member of, and this is very costly as collections are not (yet) cached.

I think the SpringSecurityResolvedAccountFilter should check if the current Authentication (the user href inside the UserDetails) matches the account resolved from request (probably from cookie param). Only if the authentication in the context does not match, or does not exist, a new authentication should be performed.

I've modified the filter to meet those needs and it works quite good, eliminating the communication to Stormpath on every possible request. The performance gains are very noticable, esp. where there are a lot of resources to load (css, js, images).

mraible:
SpringSecurityResolvedAccountFilter should check if the current Authentication (the user href inside the UserDetails) matches the account resolved from request. Only if the authentication in the context does not match, or does not exist, a new authentication should be performed.

Fixes #529.

### DIFF
--- a/extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/filter/SpringSecurityResolvedAccountFilter.java
+++ b/extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/filter/SpringSecurityResolvedAccountFilter.java
@@ -15,7 +15,6 @@ import org.springframework.util.Assert;
 import javax.servlet.FilterChain;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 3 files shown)


lhazlewood:
no tests (for regression purposes) 😢 .  Back over to you @mraible

mraible:
@lhazlewood I added unit tests and adjusted the integration tests. Please review again once CI passes.

lhazlewood:
@mraible looks good - thank you!  Please merge after the build succeeds.

PR #770, opened 2016-07-18
Made DefaultError serializable and added default constructor

### LINKED ISSUE #628: DefaultError is not serializable
User reported:

We are using Hazelcast with spring-session. So everything that is put in session is visible in all nodes.

Session contains key named SPRING_SECURITY_LAST_EXCEPTION of type `org.springframework.security.authentication.AuthenticationServiceException` that contains cause of type `ResourceException` with field error of type `DefaultError`. The latter cannot be serialized.

Can you fix that by making the class DefaultError serializable and providing default constructor ?

mraible:
Fixes #628.


dogeared:
Hey @mraible - I just wanted to confirm that we wouldn't run into anything in calling `super(null, null)` and to keep the coverage numbers up, so I added in a unit test.

If that's ok with you, then you can merged this PR.

mraible:
Thanks @dogeared. I appreciate the help.

PR #726, opened 2016-06-22
Spring MVC: Add support for returning 401 and 404 errors as JSON

mraible:
Fixes #706.

## To UAT Spring MVC

Start `examples/spring-webmvc` and make the following requests:

1. `curl -v -H "Accept: text/html" http://localhost:8080/missing`
2. `curl -v -H "Accept: application/json" http://localhost:8080/missing`

The first one should result in an HTML response, the second should contain JSON. The message from JSON is a key since i18n is often done client-side when talking to a JSON API.

## To UAT Spring Security

Start `examples/spring-security-webmvc` and make the following requests:

1. `curl -v -H "Accept: text/html" http://localhost:8080/missing`
2. `curl -v -H "Accept: application/json" http://localhost:8080/missing`
3. `curl -v -H "Accept: text/html" http://localhost:8080/restricted`
4. `curl -v -H "Accept: application/json" http://localhost:8080/restricted`

Requests with `text/html` should result in an HTML response. Requests with `application/json` should result in a JSON response.

PR #779, opened 2016-07-22
Graceful degradation when stormpath.web.register.enabled is true but the app does not have a default account store

### LINKED ISSUE #333: Make the login views for servlet and spring suppress "Create Account" if there's no default Account Store
ID Site does this already:

```
<span ng-show="canRegister" wd-can-register>
    <span> or </span>
    <a href="#register">Create an Account</a>
 </span>
```

`canRegister` is set to false if ID Site is unable to read a default password policy (indicating that there's no default account store).

lhazlewood:
Fixes #333


lhazlewood:
## UAT

Setup:

- ensure `stormpath.web.register.enabled = true` (this is the default value)
- ensure the application record in stormpath does *not* have a mapped default account store.

Tests:

A) Start `examples/servlet` via `mvn tomcat7:run` and you should see that:

1. There is a WARN message in the logs that looks like this: 

    > stormpath.web.register.enabled = true, but the application does not have anywhere to save new accounts: the application has not been assigned a default account store (either a Directory, Group or Organization).  You must specify the default account store where your application's newly created accounts will be reside.  As a result, registration will be disabled until an account store is assigned as the application's default account store (you can do this via the REST API or in the Stormpath Administration console).
1. When visiting `/login` you should not see a 'Create Account' link.
1. When manually visiting `/register` you should *not* see the registration page

B) Same results when running `java -jar examples/spring-boot-webmvc/target/*.jar`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/RegisterEnabledResolver.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Override
+    public Boolean get(HttpServletRequest request, HttpServletResponse response) {
+
+        if (!this.registerEnabled) {
+            return false;
+        }
+
+        final Application application = applicationResolver.getApplication(request);
+        return predicate.test(true, application);
```
dogeared: Your predicate uses the boolean to determine if it should return false.

You could do: `return predicate.test(this.registerEnabled, application);` and get rid of the `if` above.

Do you have the `if` so that it doesn't call `applicationResolver.getApplication(request)` unnecessarily? If that's a performance issue, then never mind!

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/impl/DefaultConfig.java
```diff
@@ -261,6 +271,33 @@ public WebHandler getRegisterPostHandler() {
...
+        if (containsKey(key)) {
+            try {
+                return getInstance(key);
+            } catch (ServletException e) {
+                throw new RuntimeException("Couldn't instantiate stormpath.web.register.enabled.resolver", e);
+            }
+        } else {
+            boolean enabled = CFG.getBoolean("stormpath.web.register.enabled");
+            RegisterEnabledResolver resolver =
+                new RegisterEnabledResolver(enabled, ApplicationResolver.INSTANCE, getRegisterEnabledPredicate());
```
dogeared: You could replace `ApplicationResolver.INSTANCE` with `getApplicationResolver()` if you want to isolate the static usage to one place.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
```diff
@@ -62,6 +65,8 @@
     private WebHandler postLoginHandler;
     private boolean idSiteEnabled;
     private boolean callbackEnabled;
+    private Resolver<Boolean> registerEnabledResolver =
+        new RegisterEnabledResolver(true, ApplicationResolver.INSTANCE, new RegisterEnabledPredicate());
```
dogeared: Why is `registerEnabledResolver` assigned here and then again in the constructor that takes a `Config` and an `ErrorModelFactory`? Is it so that it is assigned when the default constructor is used?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/RegisterEnabledResolver.java
lhazlewood: You got it - no need to invoke the predicate unless `this.registerEnabled` is true ;)

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/impl/DefaultConfig.java
lhazlewood: Nice!

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/LoginController.java
lhazlewood: Yep!  To ensure it exists when using the default constructor :)

lhazlewood:
@dogeared changes per review made.  Please confirm so we can merge :)

dogeared:
UAT Results: This breaks the expected json responses. Right now, when register is enabled, but there is no default account store for the application, the following:

`http localhost:8080/register Accept:application/json` is *not* returning a json response. 

For Spring WebMVC, it's returning a 404 response with HTML. For Servlet, it's returning the (HTML) home page.

I think it should be a 404 with an empty body.

lhazlewood:
@dogeared the issue you're seeing is due to #726.  Once that is done, it will 404 correctly.  That being the case, any objections to merging this one?

lhazlewood:
Per convo in slack, merge is ok since it's covered in #726:

image

PR #780, opened 2016-07-22
Add i18n translations for Stormpath error codes.

### LINKED ISSUE #763: Add i18n translations to stormpath error codes
The framework spec specifies that errors from the API should passthough as the error messages in the web integrations see https://github.com/stormpath/stormpath-framework-spec/blob/master/error-handling.md

Currently the Java SDK has i18n support and the implementation of the error translation from the API to the web errors tries to find i18n keys with the following format

`stormpath.web.errors.STORMPATH_ERROR_CODE`

If the key is not found then the message as it is is pass through and the exception of the message not found is log as a warning 

```
WARN  [http-bio-8080-exec-6] c.s.s.c.AbstractStormpathWebMvcConfiguration i18n key not found
org.springframework.context.NoSuchMessageException: No message found under code 'stormpath.web.errors.7104' for locale 'en_US'.
    at org.springframework.context.support.AbstractMessageSource.getMessage(AbstractMessageSource.java:159)
    at com.stormpath.spring.config.AbstractStormpathWebMvcConfiguration$3.getMessage(AbstractStormpathWebMvcConfiguration.java:1043)
```

We should add all the translations so this warning won't confuse users of what is happening and because we should support translation of this error messages.

mraible:
Fixes #763.

### DIFF
--- a/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/i18n.properties
+++ b/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/i18n.properties
@@ -158,3 +158,101 @@ stormpath.web.sendVerificationEmail.form.loginLink.text = Back to Login
 #TODO the spec doesn't provide the error message for posting unknown fields
 stormpath.web.form.fields.unknown=Unknown field [{0}]
 
+# ==================================================
+# Stormpath API Errors: #763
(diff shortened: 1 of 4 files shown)


mraible:
@mrioan @lhazlewood @dogeared @mzumbadoate @jarias Please review and vote.

lhazlewood:
LGTM, please merge!

PR #787, opened 2016-07-25
Don't add cache-control headers for JSON Views since they're set in controllers. Also, change to be consistent with value when setting cache-control header.

### LINKED ISSUE #786: Duplicate Cache-Control headers
Steps to Reproduce:

1) Launch the Spring WebMVC example
2) Get a valid access token

    http -f POST localhost:8080/oauth/token Origin:http://localhost:8080 grant_type=password username=<username> password=<password>

3) Hit the `/me` endpoint

    http localhost:8080/me Cookie:"access_token=<access token from previous step>"

4) Notice the response contains two Cache-Control headers (This is making some of the `MeIT` tests in the TCK fail)

    HTTP/1.1 200 OK
    Cache-Control: no-store, no-cache
    Cache-Control: no-store
    Content-Language: en-US
    Content-Type: application/json;charset=UTF-8
    Date: Sun, 24 Jul 2016 20:52:43 GMT
    Pragma: no-cache
    Server: Apache-Coyote/1.1
    Transfer-Encoding: chunked

mraible:
Fixes #786.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/DefaultUnauthenticatedHandler.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/DefaultUnauthenticatedHandler.java
@@ -38,7 +38,7 @@ public class DefaultUnauthenticatedHandler implements UnauthenticatedHandler {
     @Override
     public boolean onAuthenticationRequired(HttpServletRequest request, HttpServletResponse response) throws Exception {
 
-        response.setHeader("Cache-Control", "no-store");
+        response.setHeader("Cache-Control", "no-store, no-cache");
         response.setHeader("Pragma", "no-cache");
 
         if (isHtmlPreferred(request)) {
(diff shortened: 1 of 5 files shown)


dogeared:
There's one test in the TCK that's failing. This test is looking specifically for `Cache-Control: no-store` in `Oauth2IT.oauthClientCredentialsGrantSucceeds`:

```
        given()
            .param("grant_type", "client_credentials")
            .auth()
                .preemptive().basic(apiKeyId, apiKeySecret)
            .contentType(ContentType.URLENC)
        .when()
            .post(OauthRoute)
        .then()
            .statusCode(200)
            .contentType(ContentType.JSON)
            .body("access_token", not(isEmptyOrNullString()))
            .body("token_type", equalToIgnoringCase("Bearer"))
            .body("expires_in", is(3600))
            .body("refresh_token", nullValue())
            .header("Cache-Control", is("no-store"))
            .header("Pragma", is("no-cache"))
```

I am thinking that the second-to-last line of the test should be changed to `contains` rather than `is`.

Do  you concur?

edjiang:
I'm down, but is there a reason why you're sending it twice? 

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.32

>HTTP/1.1 caches SHOULD treat "Pragma: no-cache" as if the client had sent "Cache-Control: no-cache"

https://tools.ietf.org/html/rfc6749#section-5.1

```   The authorization server MUST include the HTTP "Cache-Control"
   response header field [RFC2616] with a value of "no-store" in any
   response containing tokens, credentials, or other sensitive
   information, as well as the "Pragma" response header field [RFC2616]
   with a value of "no-cache".```

PR #797, opened 2016-07-26
Allow for adding / removing properties from DefaultConfig

bdemers:
Currently `put`, `putAll`, `remove`, and `clear` cannot be used as they are backed by an unmodifiable map.

If `props` is truly not to be modified, we should instead throw an exception directly from the corresponding method (to make it more obvious)

### DIFF
--- a/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/impl/DefaultConfigFactoryTest.groovy
+++ b/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/impl/DefaultConfigFactoryTest.groovy
@@ -28,6 +28,7 @@ import java.lang.reflect.Field
 
 import static org.testng.Assert.assertEquals
 import static org.testng.Assert.assertFalse
+import static org.testng.Assert.assertNull
 import static org.testng.Assert.assertTrue
 import static org.testng.Assert.assertNotNull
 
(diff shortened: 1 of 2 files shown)


mraible:
LGTM 👍 @lhazlewood @mrioan @dogeared What do you think?

extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/impl/DefaultConfigFactoryTest.groovy
```diff
@@ -142,6 +143,73 @@ class DefaultConfigFactoryTest {
         config = new ConfigLoader().createConfig(new MockServletContext())
     }
 
+    @Test
+    public void putProperty() {
```
mrioan: No show stopper but the name of the test should a) start with the word `test` and b) help understand what it is testing; like `testPutPropertyIsAbleToModifyTheMap()`. Just for future reference

mrioan:
Just a simple educational comment from, no need to really change the code this time. LGTM. Thanks @bdemers

bdemers:
@mrioan easy enough to fix.

Do you guys usually squash and rebase before merging PR's?

lhazlewood:
I prefer a squash, but we don't usually do that.

bdemers:
updated @since tags to next release and squashed

mrioan:
@bdemers Feel free to merge. Thanks

PR #806, opened 2016-07-28
Issue 800 customize authentication scheme

iancorcoran:
Hi folks...

Apologies for the size of the PR, it's actually a few little changes that impacted many files vs large functional changes.

The main thrust of the change is removing the AuthenticationScheme enum (now the String is just passed along to the ClientBuilder etc.) and adding the concept of an AuthenticatorResolver (which can also be set on ClientBuilder). This will allow more flexibility to implement custom authenticators in the future, should the need arise. 

Jose had the idea to extract a more general ClientCredentials interface, in support of the above, and a lot of the changes are a result of having ApiKey implement this and changing various references to now point to CC. He also moved a couple of things into api from impl to support HttpAuthenticator etc. like HttpHeaders, MediaType.

Ian


api/src/main/java/com/stormpath/sdk/http/HttpAuthenticator.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Genericized interface to be implemented by HTTP authentication schemes. Such scheme defines the way the communication with
+ * the Stormpath API server will be authenticated.
+ *
+ * @see ClientCredentials
+ * @see AuthenticationScheme
+ * @since 1.0.0
+ */
+public interface HttpAuthenticator<M extends HttpMessage, C extends ClientCredentials> {
```
lhazlewood: This shouldn't be in the public `api` module - this is more about configuring the HttpClient implementation and should be limited to the `impl` module.

impl/src/main/java/com/stormpath/sdk/impl/config/ClientConfiguration.java
```diff
@@ -56,14 +57,22 @@ public void setApiKeySecret(String apiKeySecret) {
         this.apiKeySecret = apiKeySecret;
     }
 
-    public AuthenticationScheme getAuthenticationScheme() {
+    public String getAuthenticationScheme() {
         return authenticationScheme;
     }
 
-    public void setAuthenticationScheme(AuthenticationScheme authenticationScheme) {
+    public void setAuthenticationScheme(String authenticationScheme) {
```
lhazlewood: Why is this a String?  It should be an `AuthenticationScheme` instance (where `AuthenticationScheme` is an interface and not an enum).

api/src/main/java/com/stormpath/sdk/authc/AuthenticatorResolver.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Interface to implemented by resolvers capable of creating an http authenticator for a given authentication scheme.
+ * Such schemes define the way the communication with the Stormpath API server will be authenticated.
+ * @see AuthenticationScheme
+ * @see HttpAuthenticator
+ * @param <T> Any implementation of {@link HttpAuthenticator}
+ * @since 1.0.0
+ */
+public interface AuthenticatorResolver <T extends HttpAuthenticator> {
```
lhazlewood: Yeah, we should not add this concept.  The existing `AuthenticationScheme` concept should be turned in to an interface, and we add a new `AuthenticationSchemes` utility class in the `api` module that contains only `BASIC` and `SAUTHC`.  The custom scheme would be external the SDK codebase I think.  Thoughts?

lhazlewood:
@iancorcoran can you please merge from `master`?  We had a big config change merge this morning, and I'd want to ensure this branch/PR works with it.

api/src/main/java/com/stormpath/sdk/authc/AuthenticatorResolver.java
iancorcoran: Jose had started off with this... I thought having AuthenticationScheme and AuthenticationSchemes read a little confusingly, and in that implementation all the AuthenticationScheme did was return an Authenticator (which per your above comment, you're not a fan of having in api module either).

api/src/main/java/com/stormpath/sdk/api/ApiKey.java
```diff
@@ -37,7 +38,7 @@
  * @see #getAccount()
  * @since 1.0.RC
  */
-public interface ApiKey extends Resource, Saveable, Deletable, com.stormpath.sdk.client.ApiKey {
+public interface ApiKey extends Resource, Saveable, Deletable, ClientCredentials, com.stormpath.sdk.client.ApiKey {
```
lhazlewood: The public api `ApiKey` is for application users - I don't think a custom scheme's concepts should bubble-up/be coupled with the notion of our REST API's ApiKey that customers depend on.

lhazlewood: P.S. I think the idea was to remove the `com.stormpath.sdk.client.ApiKey` interface entirely because of this.  This is slated right before the 1.0.0 release since it is backwards incompatible.

api/src/main/java/com/stormpath/sdk/authc/AuthenticatorResolver.java
lhazlewood: An `AuthenticationScheme` _is_ an authenticator.  It is an algorithm that authenticates the request.  In other words, here's the default list: http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml.  If we supported them, each one would implement the `AuthenticationScheme` interface.  Adding a new or custom one is done just implementing the interface.

The `Authenticator` interface is for SDK end-users, specifically for manually authenticating HTTP requests into their own application (like for OAuth).  It isn't the same concept we use for authenticating requests made by the SDK to the Stormpath API server (I suppose they can be merged, but it's a little late for that now - maybe in a 2.0 release :).

lhazlewood: Also we use the plural name convention when there are static utility helpers or instances that people may choose to use.

For example  `Collections`, `Strings`, `Classes`, `Accounts`, etc.  Think of them as more readable `*Utils` classes.  That is, this:

```java
Accounts.where(Accounts.surname()</b>.containsIgnoreCase("Smith")<b>)</b>
     .and(<b>Accounts.givenName()</b>.eqIgnoreCase("John"))
     .orderBySurname().descending()
     .withGroups(10, 10)
     .offsetBy(50)
     .limitTo(25));
```

is easier to read than:

```java
AccountUtils.where(AccountUtils.surname()</b>.containsIgnoreCase("Smith")<b>)</b>
     .and(<b>AccountUtils.givenName()</b>.eqIgnoreCase("John"))
     .orderBySurname().descending()
     .withGroups(10, 10)
     .offsetBy(50)
     .limitTo(25));
```

So if `AuthenticationScheme` is an interface, then `AuthenticationSchemes` is a utility class that might have pre-defined instances, e.g.

`AuthenticationSchemes.SAUTHC1`
`AuthenticationSchemes.BASIC`

iancorcoran:
Closing as these changes and suggestions have been rolled into PR 812

PR #807, opened 2016-07-28
Add support for posting providerData to /login

### LINKED ISSUE #796: Support providerData POST to /login
Per framework spec: https://github.com/stormpath/stormpath-framework-spec/blob/master/login.md#-post-body-format

The `/login` endpoint should support POSTing valid providerData to do social logins.

The Framework TCK test: `FacebookSocialLoginIT.loginWithValidFacebookAccessTokenSucceeds` exercises this.

mraible:
Fixes #796.

@dogeared Rather than creating a `SocialUtils` class, I just pulled methods up into `AbstractController`.

TCK Results with this change:
```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running TestSuite
Tests run: 110, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 150.077 sec - in TestSuite

Results :

Tests run: 110, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
@@ -15,10 +15,20 @@
  */
 package com.stormpath.sdk.servlet.mvc;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.stormpath.sdk.application.Application;
+import com.stormpath.sdk.application.ApplicationAccountStoreMapping;
+import com.stormpath.sdk.directory.AccountStore;
+import com.stormpath.sdk.directory.AccountStoreVisitor;
+import com.stormpath.sdk.directory.Directory;
+import com.stormpath.sdk.group.Group;
 import com.stormpath.sdk.http.HttpMethod;
+import com.stormpath.sdk.impl.provider.DefaultGithubProvider;
 import com.stormpath.sdk.lang.Assert;
 import com.stormpath.sdk.lang.Strings;
+import com.stormpath.sdk.organization.Organization;
 import com.stormpath.sdk.servlet.account.AccountResolver;
+import com.stormpath.sdk.servlet.application.ApplicationResolver;
 import com.stormpath.sdk.servlet.event.RequestEvent;
 import com.stormpath.sdk.servlet.event.impl.Publisher;
 import com.stormpath.sdk.servlet.filter.ContentNegotiationResolver;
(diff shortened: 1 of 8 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
```diff
@@ -254,4 +280,64 @@ protected void publishRequestEvent(RequestEvent e) throws ServletException {
...
+                public void visit(Directory directory) {
+                    if ("github".equals(directory.getProvider().getProviderId())) {
+                        githubProvider[0] = (DefaultGithubProvider) directory.getProvider();
+                    }
+                }
+
+                @Override
+                public void visit(Organization organization) {}
+            };
+            accountStore.accept(accountStoreVisitor);
```
dogeared: There can only be one github directory mapped to an application. Should we check to see if `githubProvider[0]` has been set and bust out of the loop if so?

Also, I know we have this pattern elsewhere, but is there any way to do something different than hacking a `final` by making it an array?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
```diff
@@ -254,4 +280,64 @@ protected void publishRequestEvent(RequestEvent e) throws ServletException {
...
+     * This method is for exchanging a code for an access token with GitHub. Needed by LoginController
+     * and GithubCallbackController, hence the reason it's in a common parent class.
+     * @param code The code from GitHub
+     * @param request The current request
+     * @return an access token
+     */
+    protected String exchangeGithubCodeForAccessToken(String code, HttpServletRequest request) {
+        final DefaultGithubProvider[] githubProvider = new DefaultGithubProvider[1];
+
+        for (ApplicationAccountStoreMapping mapping : getApplication(request).getAccountStoreMappings()) {
```
dogeared: `LoginController` has an `accountModelFactory` that is a collection *only* of external providers - SAML and social. While there can be any number of SAML providers, there can only be one of each of the social providers. Perhaps it would be more efficient to move this into `LoginController` and make use of the `accountModelFactory` that's already setup anyway?

mraible: This method is needed by `GithubCallbackController`, so I don't think it can be pushed down. It could be extracted into a utility class like I had before with `SocialUtils`, but I didn't get that class to work with buttons (GET callback requests).

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AbstractController.java
mraible: Can't break out of an enhanced `for` loop, but I can change it to use the classic syntax if you prefer.

I'm not sure what you mean by the 2nd question.

dogeared: Hmmm... maybe not worth it. I am just concerned that with a large number of directories, this could take a long time. For instance, if the github directory were smack in the middle of an application that had 1000 directories (maybe a far fetched example), it would still iterate over an additional 500 directories.

As for my second question, I am just saying that  this:

`final DefaultGithubProvider[] githubProvider = new DefaultGithubProvider[1];`

seems like a hack because you need a final inside the interface implementation, but it's not really final as we are adding an element to the array.

Not a show stopper - we have this pattern elsewhere. It just bugs me. ;)

dogeared: We can improve this later, if we want. Not a blocker.

dogeared:
LGTM - please merge after CI

Issue #800, opened 2016-07-27
Add ability to provide an authentication scheme different than basic and/or sauthc1


josebarrueta:

jarias:
Hi @josebarrueta , what other auth schemes does the core support?

josebarrueta:
Julio, I'll close this later, we already fixed it.

Jose Luis

PR #812, opened 2016-07-29
Issue 800 enable custom authentication schemes

iancorcoran:
Fixes https://github.com/stormpath/stormpath-sdk-java/issues/800


api/src/main/java/com/stormpath/sdk/api/ApiKey.java
```diff
@@ -37,7 +38,7 @@
  * @see #getAccount()
  * @since 1.0.RC
  */
-public interface ApiKey extends Resource, Saveable, Deletable {
+public interface ApiKey extends Resource, Saveable, Deletable, ClientCredentials {
```
mrioan: What is the rationale for this new Interface?

api/src/main/java/com/stormpath/sdk/client/AuthenticationSchemes.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2016 Stormpath, Inc.  All rights reserved.
+ */
+package com.stormpath.sdk.client;
+
+import com.stormpath.sdk.lang.Classes;
+
+public class AuthenticationSchemes {
```
mrioan: Missing javadoc. Classes in the `api` module are accessible to be used by end users, they should always be properly documented so it explains what this class is for and it must have the `@since` tag: 
```
/**
 * @since 1.0.0
*/

File: api/src/main/java/com/stormpath/sdk/client/ClientCredentials.java [hunk 3]
```diff
@@ -0,0 +1,6 @@
+package com.stormpath.sdk.client;
+
+public interface ClientCredentials {
```
mrioan: Missing javadoc

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/client/DefaultClientTest.groovy
```diff
@@ -62,7 +63,7 @@ class DefaultClientTest {
         String baseUrl = "http://localhost:8080/v1"
         def proxy = createStrictMock(com.stormpath.sdk.client.Proxy)
         def cacheManager = createStrictMock(CacheManager)
-        def authcScheme = AuthenticationScheme.SAUTHC1
+        def authcScheme = new SAuthc1RequestAuthenticator()
```
mrioan: We should try to avoid creating classes ourselves when possible in order to mimic what end users will do:  `AuthenticationScheme.SAUTHC1`

api/src/main/java/com/stormpath/sdk/api/ApiKey.java
iancorcoran: We want the ability to implement custom authorization schemes. So an ApiKey would just be one implementation of a Client Credentials abstraction. Jose had some more thoughts around this,  will be submitting a further commit for this PR in a bit.

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutorIT.groovy
```diff
@@ -4,7 +4,7 @@ import com.stormpath.sdk.account.Account
 import com.stormpath.sdk.client.AuthenticationScheme
 import com.stormpath.sdk.client.ClientIT
 import com.stormpath.sdk.http.HttpMethod
-import com.stormpath.sdk.impl.http.Response
+import com.stormpath.sdk.http.Response
```
mrioan: This IT test is not working.

mrioan:
I would need to see a few ITs showing how the schemes are changed and requests are actually working with them.

api/src/main/java/com/stormpath/sdk/client/AuthenticationSchemes.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static final String SAUTHC1 = "SAUTHC1";
```
lhazlewood: These shouldn't be Strings.  They should be actual instances.  For example:

```java
public static final AuthenticationScheme SAUTHC1 = Classes.newInstance("com.stormpath.sdk.impl.http.authc.SAuthc1RequestAuthenticator");
//etc...
```

PR #841, opened 2016-08-02
Fixed base url for servlet to be consistent with framework spec.

### LINKED ISSUE #840: update servlet to process base url according to sdk spec
As outlined here, https://github.com/stormpath/stormpath-sdk-spec/blob/master/specifications/config.md

base url should be set from `stormpath.client.baseUrl`

For java, it should also respond to: `STORMPATH_CLIENT_BASEURL` environment variable

dogeared:
fixes #840


mraible:
LGTM 👍

dogeared:
just waiting for CI before merging. Jobs are in the queue, but it's not reflected here for some reasone.

PR #852, opened 2016-08-02
AuthenticationScheme converted to interface, AuthenticationSchemes he…

josebarrueta:
…lper class added, Http Request and dependencies moved from impl to api, ClientCredentials abstraction from ApiKey introduced


api/src/main/java/com/stormpath/sdk/client/ClientCredentials.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.stormpath.sdk.client;
+
+/**
+ * Credentials to be used when authenticating requests to the Stormpath API server.
+ * @since 1.0.0
+ */
+public interface ClientCredentials {
```
mrioan: Why not using ApyKey class? Why is this new interface needed?

josebarrueta: Because, we want to be able to have a different type of credentials to authenticate against Stormpath Rest API, ```ApiKeyCredentials``` will be one implementation of ```ClientCredentials```, but other can exists.

josebarrueta: Because, we want to be able to have a different type of credentials to authenticate against Stormpath Rest API, ApiKeyCredentials will be one implementation of ClientCredentials, but others can exist.

dogeared:
Closing as need to defer to 2.0

josebarrueta:
@dogeared what you mean to defer it to 2.0? We might need it sooner than that.

PR #854, opened 2016-08-02
CORSFilter support for Servlet

### LINKED ISSUE #699: Support CORS preflight
for internal, per: https://stormpath.zendesk.com/agent/tickets/7806

Right now, as a workaround, you can create filter to properly handle a browser OPTIONS preflight request:

```
@Component
public class MyCORSFilter implements Filter {

    @Autowired
    RequestAuthorizer stormpathOriginAccessTokenRequestAuthorizer;

    private final Logger log = LoggerFactory.getLogger(MyCORSFilter.class);

    public MyCORSFilter() {
        log.info("SimpleCORSFilter init");
    }

    @Override
    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {

        HttpServletRequest request = (HttpServletRequest) req;
        HttpServletResponse response = (HttpServletResponse) res;

        try {
            stormpathOriginAccessTokenRequestAuthorizer.assertAuthorized(request, response);

            response.setHeader("Access-Control-Allow-Origin", request.getHeader("Origin"));
            response.setHeader("Access-Control-Allow-Credentials", "true");
            response.setHeader("Access-Control-Allow-Methods", "POST, GET, OPTIONS, DELETE");
            response.setHeader("Access-Control-Max-Age", "3600");
            response.setHeader("Access-Control-Allow-Headers", "Content-Type, Accept, X-Requested-With, remember-me");
        } catch (OauthException e) {
            log.error("Specified Origin '{}' is not allowed: {}", request.getHeader("Origin"), e.getMessage(), e);
        }

        chain.doFilter(req, res);
    }

    @Override
    public void init(FilterConfig filterConfig) {
    }

    @Override
    public void destroy() {
    }

}
```

This is something we should build in to the SDK.

mraible:
fixes #699 

This is just the beginning of this implementation, but it does work. I based the servlet example off of https://spring.io/guides/gs/rest-service-cors/.

Here's the steps to reproduce:

1. Start `examples/servlet` using `mvn tomcat7:run`. Go to http://localhost:8080/hello.html and you should see an HTML page that resembles the following.

```
The ID is 1

The content is Hello, null!
```

This request does not use CORS, so this is just to prove the `GreetingController` and XHR request to it works.

2. With Tomcat still running, open a new terminal and cd into `examples/servlet/src/main/webapp`, then run `python -m SimpleHTTPServer $1`. This starts a web server for the current directory, running on port 8000. Open http://localhost:8000/hello.html in your browser and you should see a similar response to above.

3. Shut down Tomcat, run `export STORMPATH_WEB_CORS_ENABLED=false`, then restart Tomcat with `mvn tomcat7:run`. Now when you hit http://localhost:8000/hello.html, the ID and content should be blank and an error will be in your browser console.


mraible:
Fixes #699.

examples/servlet/src/main/java/com/stormpath/sdk/examples/servlet/Greeting.java
```diff
@@ -0,0 +1,20 @@
+package com.stormpath.sdk.examples.servlet;
```
jarias: Missing license header

examples/servlet/src/main/java/com/stormpath/sdk/examples/servlet/GreetingController.java
```diff
@@ -0,0 +1,29 @@
+package com.stormpath.sdk.examples.servlet;
```
jarias: Missing license header

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/filter/CORSFilterFactory.java
```diff
@@ -0,0 +1,40 @@
...
+import com.stormpath.sdk.servlet.authz.RequestAuthorizer;
+import com.stormpath.sdk.servlet.config.Config;
+import com.stormpath.sdk.servlet.filter.cors.CORSFilter;
+
+import javax.servlet.ServletContext;
+
+import static com.stormpath.sdk.servlet.config.impl.DefaultConfig.REQUEST_AUTHORIZER;
+
+/**
+ * @since 1.0.0
```
jarias: I think this should target 1.1.0

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/cors/CORSFilter.java
```diff
@@ -0,0 +1,44 @@
+package com.stormpath.sdk.servlet.filter.cors;
```
jarias: license header

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/cors/CORSFilter.java
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.servlet.FilterChain;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public class CORSFilter extends HttpFilter {
+
+    private static final Logger log = LoggerFactory.getLogger(CORSFilter.class);
+    private RequestAuthorizer originRequestAuthorizer;
```
jarias: This is never used

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/cors/CORSFilter.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public void setOriginRequestAuthorizer(RequestAuthorizer originRequestAuthorizer) {
+        this.originRequestAuthorizer = originRequestAuthorizer;
+    }
+
+    @Override
+    protected void filter(HttpServletRequest request, HttpServletResponse response, FilterChain chain)
+            throws Exception {
+
+        try {
```
jarias: Origin is never checked, also the original ticket talks about implementing the CORS preflight which is used by SPA frameworks like angular, where they do an OPTIONS request in that case the request should not continue the filter chain since it only needs to respond with the CORS headers.

PR #858, opened 2016-08-04
Added social login documentation.

### LINKED ISSUE #646: Add social login instructions to Spring Boot docs
Even though social login is only supported via ID Site today, we should still have information to that effect in the docs. Today, a user would conclude by the docs that social login is not available in any form.

jarias:
Added both servlet and spring boot docs.

Fixes #646


dogeared:
Hey @jarias - please remove the changes to .travis.yml and the HREF changes to the properties files. Those have all been updated in master.

Also, do those tests still need to be disabled?

PR #880, opened 2016-08-08
Added social login documentation

### LINKED ISSUE #646: Add social login instructions to Spring Boot docs

jarias:
Fixes #646


docs/source/social.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. contents::
+   :local:
+   :depth: 2
+
+Overview
+--------
+
+Besides username and password :ref:`login <login>`, users can leverage Stormpath
+social directory support to single sign on with their favorite social providers
```
mrioan: Single Sign On is is a different thing. Please rephrase this sentence so this is not mentioned here.

mrioan: Please make the supported Social providers a bullet list so they can clearly "advertise" them

docs/source/social.rst
```diff
@@ -0,0 +1,65 @@
...
+(Facebook, GitHub, Google and LinkedIn).
+
+.. image:: /_static/social-login.png
+
+After clicking on the desire provider the user should follow the usual OAuth2
```
mrioan: Please do not use "should". "Will" is more appropriate as it means that things will happen. Should implies that things could or could not happen.

mrioan:
@jarias this documentation does not explain how to get social login actually working. Please see `https://docs.stormpath.com/php/laravel/latest/social_login.html` and `https://docs.stormpath.com/nodejs/express/latest/social_login.html` and mimic what they documented. The more you can copy the better, we should not try to author new text. 

Most product guides tailor just SDK-specific parts but take all of the documentation from the Rest API guide. We must try to do the same for the web integration, just replicate/copy whatever has been documented for another SDK while tailoring only what is Java-specific.

back to you @jarias

mrioan:
@jarias  I just reviewed social login documentation. Did a few (minor) changes myself so we avoid comings and goings…
LGTM. Thanks

dogeared:
I rebased master. Once CI passes, I will merge.

PR #909, opened 2016-08-16
Nullcheck to avoid NPE

### LINKED ISSUE #908: ?error causes NPE
Launch a NON-Spring Security example. Browse to:

`localhost:8080/login?error`

an NPE is thrown. This is due to code looking for a Spring Security attribute in a session that may not exist.

dogeared:
fixes #908


mraible:
👍 Code looks good. Did not test locally.

PR #916, opened 2016-08-17
Ignore invalid status parameter on login.

### LINKED ISSUE #907: unknown status codes should be ignored
If you launch one of our examples and go to:

`localhost:8080/login?status=blarg`

You will see an message source placeholder since the key cannot be found. 

We should simply ignore unknown status codes

mraible:
Fixes #907.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultLoginFormStatusResolver.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultLoginFormStatusResolver.java
@@ -19,6 +19,8 @@ import com.stormpath.sdk.lang.Assert;
 import com.stormpath.sdk.servlet.i18n.MessageSource;
 
 import javax.servlet.http.HttpServletRequest;
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 /**
  * @since 1.0.0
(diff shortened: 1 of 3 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultLoginFormStatusResolver.java
```diff
@@ -37,6 +39,17 @@ public DefaultLoginFormStatusResolver(MessageSource messageSource, String verify
...
+        validStatus.add("unverified");
+        validStatus.add("verified");
+        validStatus.add("created");
+        validStatus.add("forgot");
+        validStatus.add("reset");
+
+        if (validStatus.contains(status)) {
+            return messageSource.getMessage("stormpath.web.login.form.status." + status, request.getLocale(), verifyEmailUri);
+        } else {
+            throw new IllegalArgumentException("Status '" + status + "' is not valid.");
```
dogeared: Perhaps the `else` could just set the empty string rather than throw an exception? You could put in a comment that it's invalid. Then, you don't need the try/catch in LoginController.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultLoginFormStatusResolver.java
```diff
@@ -37,6 +39,17 @@ public DefaultLoginFormStatusResolver(MessageSource messageSource, String verify
 
     @Override
     public String getStatusMessage(HttpServletRequest request, String status) {
-        return messageSource.getMessage("stormpath.web.login.form.status." + status, request.getLocale(), verifyEmailUri);
+        Set<String> validStatus = new LinkedHashSet<>();
```
dogeared: Thinking this should be externalized so it's not buried here. Maybe as an enum? The approved solution [here](http://codereview.stackexchange.com/questions/42620/validating-an-input-string-against-some-enum-fields) speaks to checking if an enum contains a given string value version.

mraible:
@dogeared I changed to use an Enum and return an empty String. Thanks for the feedback!

dogeared:
LGTM - please merge after CI passes 👍

PR #923, opened 2016-08-18
703: enabled CompositeMessageSource bean definition when necessary

### LINKED ISSUE #703: i18n message translation fails in some Spring Boot web apps
`AbstractStormpathWebMvcConfiguration#stormpathSpringMessageSource()` checks to see if the specified message source is null or a placeholder, where 'placeholder' is indicated if it is a DelegatingMessageSource that does not have a parent.

I encountered a scenario today where the message source was a DelegatingMessageSource and it *did* have a parent.  However, the parent was itself a DelegatingMessageSource (which did not have a parent).  I suspect this occurs when nesting application contexts or other 'frameworky' scenarios.

I believe the best/safest thing to do is to keep walking the parent hierarchy until we find one that is 1) not a DelegatingMessageSource and 2) not null.

If these two conditions are met, the CompositeMessageSource instance should be created using the Stormpath i18nPropertiesMessageSource and the discovered non-DelegatingMessageSource.

If these two conditions are not met, then just create/return the Stormpath i18nPropertiesMessageSource.

The quick workaround for this scenario until this bug is fixed is for the app developer to explicitly declare a MessageSource bean in their Spring configuration:

```java
@Bean
public MessageSource messageSource() {
    ResourceBundleMessageSource src = new ResourceBundleMessageSource();
    src.setBasename("com.stormpath.sdk.servlet.i18n");
    src.setDefaultEncoding("UTF-8");
    return src;
}
```

And the `stormpathSpringMessageSource` won't be created (preferring the user default).

lhazlewood:
Fixes #703.

Targeted at the 1.1.x branch due to new classes


mraible:
@lhazlewood I merged the latest master into this branch to eliminate conflicts. Not sure if that's the best approach.

PR #928, opened 2016-08-18
903 adding debug msg when i18n key not found

### LINKED ISSUE #903: Why are many Stormpath error codes not covered by i18n.properties?
I just tried logging in with a deliberately invalid password.

This results in an error with the message key `stormpath.web.errors.7100`.

This does not have an entry in [`i18n.properties`](https://github.com/stormpath/stormpath-sdk-java/blob/master/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/i18n.properties).

Providing at least the English versions of these errors was covered by #763 - which is now marked as closed.

If one looks at `i18n.properties` one sees many codes are covered, but some like 7100 are not. An invalid password must be amongst one of the most common errors seen so its omission seems odd.

7100 is covered by [`api-errors.properties`](https://github.com/stormpath/stormpath-sdk-java/blob/master/extensions/servlet/src/test/resources/com/stormpath/sdk/servlet/api-errors.properties) under the test resources.

Here it has an overly specific value (which is fair enough for test) which gives away the fact that the problem is with the password:

    status.7100.dev = Login attempt failed because the specified password is incorrect.

So why despite all this does one still see "Invalid username or password." in the web frontend? The answer is that you hit the `NoSuchMessageException` catch block of `getMessage(String key, String defaultMessage, Locale locale, Object... args)` in [`AbstractStormpathWebMvcConfiguration`](https://github.com/stormpath/stormpath-sdk-java/blob/master/extensions/spring/stormpath-spring-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebMvcConfiguration.java) which logs a warning and then returns a default value - in this case the value retrieved for `com.stormpath.sdk.servlet.mvc.LoginErrorModelFactory.INVALID_LOGIN_MESSAGE`.

So the end user sees something OK - but the logs get spammed with warnings and stack traces.

mzumbadoate:
fixes #903

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
@@ -15,6 +15,9 @@
  */
 package com.stormpath.sdk.servlet.i18n;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.UnsupportedEncodingException;
 import java.text.MessageFormat;
 import java.util.Locale;
@@ -26,6 +29,7 @@ import java.util.ResourceBundle;
  */
 public class DefaultMessageSource implements MessageSource {
 
+    private static final Logger log = LoggerFactory.getLogger(DefaultMessageSource.class);
     public static final String BUNDLE_BASE_NAME = "com.stormpath.sdk.servlet.i18n";
 
     @Override
@@ -41,6 +45,7 @@ public class DefaultMessageSource implements MessageSource {
             String msg = new String(bundle.getString(key).getBytes("ISO-8859-1"), "UTF-8");
             return MessageFormat.format(msg, new Object[]{});
         } catch (MissingResourceException e) {
+            log.debug("Couldn't load the property " + key);
             return defaultMessage;
         } catch (UnsupportedEncodingException e) {
             throw new IllegalStateException("Couldn't load property from resource bundle", e);
@@ -54,6 +59,7 @@ public class DefaultMessageSource implements MessageSource {
             String msg = new String(bundle.getString(key).getBytes("ISO-8859-1"), "UTF-8");
             return MessageFormat.format(msg, args);
         } catch (MissingResourceException e) {
+            log.debug("Couldn't load the property " + key);
             return defaultMessage;
         } catch (UnsupportedEncodingException e) {
             throw new IllegalStateException("Couldn't load property from resource bundle", e);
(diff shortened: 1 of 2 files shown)


dogeared:
LGTM. Do we already have test coverage around the exception path? If not, maybe a unit test?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
```diff
@@ -41,6 +45,7 @@ public String getMessage(String key, String defaultMessage, Locale locale) {
             String msg = new String(bundle.getString(key).getBytes("ISO-8859-1"), "UTF-8");
             return MessageFormat.format(msg, new Object[]{});
         } catch (MissingResourceException e) {
+            log.debug("Couldn't load the property " + key);
```
lhazlewood: You should not use String concatenation in log messages, as they create GC garbage/overhead even if the log level is disabled.

Please use placeholders, e.g.:

```java
log.debug("Couldn't load the property for key {}", key);
```

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
```diff
@@ -54,6 +59,7 @@ public String getMessage(String key, String defaultMessage, Locale locale, Objec
             String msg = new String(bundle.getString(key).getBytes("ISO-8859-1"), "UTF-8");
             return MessageFormat.format(msg, args);
         } catch (MissingResourceException e) {
+            log.debug("Couldn't load the property " + key);
```
lhazlewood: same

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/i18n/DefaultMessageSource.java
```diff
@@ -70,6 +76,7 @@ public String getMessage(String key, Locale locale, Object... args) {
...
```
lhazlewood: same

extensions/spring/stormpath-spring-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebMvcConfiguration.java
```diff
@@ -1174,6 +1174,7 @@ public String getMessage(String key, String defaultMessage, Locale locale, Objec
                     return springMessageSource.getMessage(key, args, locale);
                 } catch (NoSuchMessageException e) {
                     //Same behavior as com.stormpath.sdk.servlet.i18n.DefaultMessageSource
+                    log.debug("Couldn't load the i18n property " + key);
```
lhazlewood: same

dogeared:
@mzumbadoate - I updated the log messages per @lhazlewood so that CI would be done when you come back online in the morning.

Pending an answer to the question above, this is ready to go.

PR #931, opened 2016-08-19
SAML and Social Login does not redirect the user back to the intercepted pre-login uri

### LINKED ISSUE #849: SAML and Social Login does not redirect the user back to the intercepted pre-login uri
*SAML and Social Login support relay state in Stormpath but we are not using it*

During Login, this is the expected behavior with SAML and Social:

1. Open `localhost:8080`
2. Click /restricted
3. You are redirected to the login page
4. You login with SAML (or Social)

**Actual Outcome**: You are redirected to `/`
**Expected Outcome**: You are redirected to `/restricted`

Stormpath supports `relay state` for SAML and Social. We need to use that to retain the `next` value. For us, that state is a JWT which allows us to put the next URI in it.

jarias:
Fixes #849

### DIFF
--- a/extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js
+++ b/extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js
@@ -43,26 +43,39 @@ function baseUrl() {
 }
 
 function linkedinLogin(clientId) {
-    window.location.replace('https://www.linkedin.com/uas/oauth2/authorization' +
-        '?client_id=' + clientId +
-        '&response_type=code' +
-        '&scope=' + encodeURIComponent('r_emailaddress r_basicprofile') +
-        '&redirect_uri=' + encodeURIComponent(baseUrl() + '/callbacks/linkedin') +
-        '&state=${oauthStateToken}');
+    window.location.replace(
+        buildUrl('https://www.linkedin.com/uas/oauth2/authorization',
+            {
+                client_id: clientId,
+                response_type: 'code',
+                scope: 'r_emailaddress r_basicprofile',
+                redirect_uri: baseUrl() + '/callbacks/linkedin',
+                state: 'oauthState' //linkedin requires state to be pass all the time.
+            }
+        )
+    );
 }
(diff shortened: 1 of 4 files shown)


extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js
```diff
@@ -94,4 +107,22 @@ function facebookLogin(appId) {
...
+
+function buildUrl(url, params) {
+    var next = getParameterByName('next');
+
+    if (next !== undefined && next !== '') {
+        params.state = next;
+    }
+
+    if (url.includes('?')) {
+        return url + '&' + $.param(params);
```
mraible: Do we require jQuery to be present for Stormpath to work? `$.param(params)` is a jQuery function.

mraible:
Code looks good, but curious about requiring jQuery. Do you have a tenant I can UAT this against?

extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js
lhazlewood: We should not require jQuery if possible - our default pages should be as lightweight as possible.  Users can add whatever they want in their own templates.

dogeared: There are a few things we require for our default templates - both JSP and thymeleaf - including bootstrap and jquery.

`stormpath.js` is primarily for the javascript clients for social providers.

jarias:
@mraible jQuery was already included so I just used it for UAT, I can add you to the SAML tenant, do you have an stormpath email address?

mraible:
@jarias Yes - matt.raible@stormpath.com.

jarias:
Just added  you, there are 2 SAML directories use SAML Julio Test when login, the other one doesn't work.

Tenant name fine-trader

For social use tenant tender-sword also invited

dogeared:
UAT complete. Behaves as expected.

PR #932, opened 2016-08-19
Improve timeout settings so setting in seconds is expected and documented.

### LINKED ISSUE #920: Timeout in properties file is expected in secs but milliseconds when programmatically
At least this is what @josebarrueta and me discovered while being confused about documentation and code.

We need to validate that this is actually a problem and if true we need to use seconds everywhere (milliseconds seems to much granular with no real reason)

`DefaultClientBuilder` has this (note `*1000`):
```
if (props.get(DEFAULT_CLIENT_CONNECTION_TIMEOUT_PROPERTY_NAME) != null) {
            clientConfig.setConnectionTimeout(Integer.valueOf(props.get(DEFAULT_CLIENT_CONNECTION_TIMEOUT_PROPERTY_NAME)) * 1000);
        }
```

On the other hand `ClientBuilder#setConnectionTimeout` says this in its javadoc:

```
* @param timeout connection and socket timeout in milliseconds
```

Additionally be sure that documentation is also updated if required

mraible:
Fixes #920.

### DIFF
--- a/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
+++ b/extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
@@ -171,10 +171,14 @@ public class HttpClientRequestExecutor implements RequestExecutor {
             );
         }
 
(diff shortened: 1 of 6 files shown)


dogeared:
LGTM

PR #940, opened 2016-08-24
Issue 877 set request auth factory clientbuilder

### LINKED ISSUE #877: add ability to set a Request Authentication Factory in ClientBuilder
```
    // implementation needs to ensure that factory implements RequestAuthenticator
    ClientBuilder setRequestAuthenticatorFactory(Factory<?> factory);
```

iancorcoran:
resolves #877


api/src/main/java/com/stormpath/sdk/client/ClientBuilder.java
```diff
@@ -351,6 +351,14 @@
     ClientBuilder setBaseUrl(String baseUrl);
 
     /**
+     * Sets the request authentication factory for the client to use. It is not recommended to use this setting.
+     * @param factory the request authenticator factory to use.
+     * @return the ClientBuilder instance for method chaining
+     * @since 1.0.4
+     */
+    ClientBuilder setRequestAuthenticatorFactory(Object factory);
```
mrioan: If this method is not "recommended" (as documented in its Javadoc) then I think it should be removed from the interface and kept it in the implementation class. Additionally adding this method to the interface causes a lot new dependencies to be added to the API module. I think this change is doing far more noise that it should...

Experienced users can do something like:

```
ClientBuilder builder = Clients.builder();
((DefaultClientBuilder) builder). setRequestAuthenticatorFactory(factory);
Client client = builder.build();
```

iancorcoran: Had done this based on Les's recommendation... but given that we ended up needing to rely on DefaultClientBuilder anyway (to set ClientCredentials) I think this can be removed.

impl/src/main/java/com/stormpath/sdk/impl/authc/credentials/ApiKeyFileCredentialsProvider.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.stormpath.sdk.lang.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.util.Properties;
+
+public class ApiKeyFileCredentialsProvider extends AbstractApiKeyCredentialsProvider {
```
mrioan: What is this class for? 

Additionally, missing `@since <VERSION_HERE>`

impl/src/main/java/com/stormpath/sdk/impl/authc/credentials/ClientCredentials.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.stormpath.sdk.impl.authc.credentials;
+
+/**
+ * Credentials to be used when authenticating requests to the Stormpath API server.
+ * @since 1.0.4
+ */
+public interface ClientCredentials {
```
mrioan: Why was this new class introduced? What is its purpose?

impl/src/main/java/com/stormpath/sdk/impl/authc/credentials/DefaultClientCredentialsProviderChain.java
```diff
@@ -0,0 +1,38 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.stormpath.sdk.impl.authc.credentials;
+
+import com.stormpath.sdk.impl.config.ClientConfiguration;
+import com.stormpath.sdk.lang.Assert;
+
+public class DefaultClientCredentialsProviderChain extends ClientCredentialsProviderChain {
```
mrioan: I understand the purpose of this. But this is a good example of what I was trying to say at the very beginning. Our properties handling implementations works just fine, we never had an issue with it.

Just for the sake of adding a new method to the ClientBuilder this whole implementation has been modified (?). I think that you are trying to kill a mosquito with a tank. If there was no other alternative I could understand this but I do think there is a way to kill the mosquito with a simple slap: branch `850_AuthenticationScheme_to_interface`

impl/src/main/java/com/stormpath/sdk/impl/authc/credentials/ClientCredentials.java
iancorcoran: Right now the SDK is reliant on having an ApiKey to talk to Stormpath. Abstracting out the idea of credentials will allow other types of credentials to be used (ping me on hipchat if you want to talk more of the business scenario for that). Pretty much anywhere ApiKeys are used today can be replaced by the equivalent ApiKeyCredentials.

mrioan:
I do not think it is necessary for me to continue reviewing this PR. 

It is causing too many changes for the sake of adding a new method to the `ClientBuilder`. I already provided a much simpler implementation in branch `850_AuthenticationScheme_to_interface`. 

If I am overlooking something and this PR is indeed doing something that `850_AuthenticationScheme_to_interface` is not doing then please provide more information so I can better understand what this PR is trying to achieve.

This PR is not approved.

impl/src/main/java/com/stormpath/sdk/impl/authc/credentials/ApiKeyFileCredentialsProvider.java
iancorcoran: There are a number of rules in ClientApiKeyBuilder around where to look (system properties, environment variables, file locations etc) for api key details. In order to build ClientCredentials rather than just an ApiKey, I have created the concept of a CredentialsProvider (in this case ApiKeyFileCredentialsProvider builds ApiKeyCredentials based on looking at a configured or default api key properties file). These CredentialsProviders can be chained together (see DefaultClientCredentialsProviderChain which follows the order of preference previously defined in ClientApiKeyBuilder). It's largely similar to what Amazon are doing with their AWS SDK. 

It also gives us the future flexibility to specify our own CredentialsProvider chains, if setting the credentials on DefaultClientBuilder directly isn't sufficient.

iancorcoran:
It is not just for the sake of 1 method that these changes are necessary. `850_AuthenticationScheme_to_interface` has `RequestAuthenticator` depending on `ApiKey`. Since we will not be leveraging an api key for our scenario, the abstraction of `ClientCredentials` makes sense. So changing the `RequestAuthenticator` interface and the corresponding factory is also necessary. 

We could in theory just build an `ApiKeyCredentials` out of whatever `ClientApiKeyBuilder` gives us back (assuming `ClientCredentials` wasn't explicitly set on `DefaultClientBuilder`), but that is less extensible. 

FWIW @josebarrueta @lhazlewood and I reviewed and agreed with building out the `ClientCredentialsProviderChain` on Friday afternoon.

lhazlewood:
sorry @mrioan that I didn't have a chance to talk with you and @dogeared about this last week.  I'm fine with these changes assuming all integration tests work, especially because the `api` module wasn't impacted at all - this is not customer-facing.

PR #942, opened 2016-08-26
Use CookieCsrfTokenRepository instead of HttpSessionCsrfTokenRepository so Spring Security can be configured stateless

### LINKED ISSUE #918: StormpathWebSecurityConfigurer having the csrfTokenEnabled (default) adds JSESSIONID cookie in response.
In my case, I just wanted to be able to use the spring-boot (with spring-security) module to protect my REST endpoints, as it's REST application I didn't want to have JSESSIONID cookie in my responses, since my service is stateless. 

Going through the code, to find what was creating a session in the server during the http calls, found that the session is created by the default implementation of the CSRF token reposity.

In ```StormpathWebSecurityConfigurer.java``` this is the Bean definition:

```
    @Autowired
    @Qualifier("stormpathCsrfTokenRepository")
    private CsrfTokenRepository csrfTokenRepository;

    ...

    @Value("#{ @environment['stormpath.web.csrf.token.enabled'] ?: true }")
    protected boolean csrfTokenEnabled;
    ...
```

Which is defined by ```AbstractStormpathWebSecurityConfiguration.java```:

```
import org.springframework.security.web.csrf.HttpSessionCsrfTokenRepository;
...    
    public CsrfTokenRepository stormpathCsrfTokenRepository() {
        HttpSessionCsrfTokenRepository csrfTokenRepository = new HttpSessionCsrfTokenRepository();
        csrfTokenRepository.setParameterName(csrfTokenName);
        return csrfTokenRepository;
    }
```

The HttpSessionCsrfTokenRepository (implemented in spring project) creates a session (if none exists).

```
...

public void saveToken(CsrfToken token, HttpServletRequest request,
			HttpServletResponse response) {
		if (token == null) {
			HttpSession session = request.getSession(false);
			if (session != null) {
				session.removeAttribute(this.sessionAttributeName);
			}
		}
		else {
			HttpSession session = request.getSession();
			session.setAttribute(this.sessionAttributeName, token);
		}
	}
...
```

Maybe, the default bean definition is OK, but we should document this better for people not to spend a lot of time trying to figure this out. Or default to an implementation that doesn't require an HTTP session being created.

Either way, as of today it's easy to remove this behavior by disabling the csrf token (which I did) or by overriding the implementation of the CsrfTokenRepository for one with no session creation.

mraible:
Fixes #918.


jarias:
LGTM

jarias:
@mraible I wonder if this should target 1.1.0 instead of 1.0.4 since it changes how CSRF works for Spring Sec, for the end user is transparent, but for devs it could be a breaking change if they where using the session stored csrf information.

What do you think @lhazlewood @dogeared

mrioan:
Additionally, our implicit philosophy is to keep Spring Security with default behaviors. By default Spring Security is using sessions instead of cookies

What I would suggest is to provide this new behavior as an additional bean. Users can then decide which behavior to use via configuration. By default be continue using Sessions as always.

mraible:
@mrioan You are correct, when CSRF is enabled with Spring Security, session is used by default. The Spring documentation shows you that you can change this using the following:

```java
@EnableWebSecurity
public class WebSecurityConfig extends WebSecurityConfigurerAdapter {

	@Override
	protected void configure(HttpSecurity http) throws Exception {
		http
			.csrf()
				.csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse());
	}
}
```

What do you think the best way to resolve this issue is? Should we add a link to this documentation or recommend users override the default `CsrfTokenRepository` bean as follows:

```java
@Value("#{ @environment['stormpath.web.csrf.token.name'] ?: '_csrf'}")
protected String csrfTokenName;

@Bean(name = "stormpathCsrfTokenRepository")
public CsrfTokenRepository stormpathCsrfTokenRepository() {
    CookieCsrfTokenRepository csrfTokenRepository = new CookieCsrfTokenRepository();
    csrfTokenRepository.setParameterName(csrfTokenName);
    return csrfTokenRepository;
}
````

If we're only updating documentation, where would you recommend we document this at?

**Comment by mrioan** [comment]
@mraible if we already have the code written and working then I think we can keep that code available and then users can decide which approach to use. Therefore we need both code and documentation explaining how to switch to the cookie-based csrf strategy. Any thoughts?

**Comment by jarias** [comment]
@mrioan Why not make it configurable? Add a new property that lets the user decide the impl, leaving the default to what we have already.

**Comment by mraible** [comment]
@mrioan @jarias I'll work on making it configurable. What should the property name be? Here's a suggestion:

```
stormpath.web.csrf.token.repository=session
```

The other option would be `cookie`. Or we could use `stateful` and `stateless`.

**Comment by mrioan** [comment]
@mraible `stormpath.web.csrf.token.repository = session` and `stormpath.web.csrf.token.repository = cookie` makes sense to me.

Keep in mind that this configuration is only for SpringSecurity so remember to add a comment in right above the new property in `web.stormpath.properties`

Thanks!

**Comment by mraible** [comment]
@mrioan @jarias I made these suggested changes. Please review again.

File: extensions/spring/stormpath-spring-security-webmvc/src/main/java/com/stormpath/spring/config/AbstractStormpathWebSecurityConfiguration.java [hunk 1]
```diff
@@ -128,9 +132,23 @@ public SamlResultListener springSecuritySamlResultListener() {
     }
 
     public CsrfTokenRepository stormpathCsrfTokenRepository() {
-        HttpSessionCsrfTokenRepository csrfTokenRepository = new HttpSessionCsrfTokenRepository();
-        csrfTokenRepository.setParameterName(csrfTokenName);
-        return csrfTokenRepository;
+        // 918: Use session-based CSRF token repository by default, since that's what Spring Security uses by default
+        if ("session".equals(csrfTokenRepository)) {
```
mrioan: Can you please move this if logic to a `CsrgTokenRepositoryFactory`?

mrioan: @mraible I think it would make sense to move this if logic to a `CsrgTokenRepositoryFactory`.

mraible: Why? I don't know if a simple if statement really needs to be pulled into a whole new class.

dogeared:
@mrioan @mraible - Is this good to go? Looks like it. If so, please merge!

PR #945, opened 2016-08-30
Adding a test and updating doc for servlet init param config

bdemers:
The previous doc mentioned any `stormpath.*` init params would be used, however, only
properties under a single init parameter of `stormpath.properties` are used.


mraible:
LGTM 👍

PR #947, opened 2016-08-31
Allow the servlet plugin to be disabled by setting `stormpath.enabled = false`

### LINKED ISSUE #802: Disable Stormpath in Web App
We need a way to disable Stormpath for good in a Web Application that depends on us. 

`stormpath.web.enabled = false` doesn't totally disable Stormpath as it is still asking for the `ApiKey`.

For reference see ticket 8668

### LINKED ISSUE #871: Disable Servlet Plugin
Our Servlet plugin cannot currently being totally disabled. Endpoints can but not the Servlet. 

For example, an application that has the servlet plugin integrated will never be able to boot if the `apiKey.properties` file is not present. 

We would need to provide a way (programmatically?) to avoid the Servlet plugin to even start.

We want to support this in Spring and Spring Boot as well (Spring Boot should already be working, so this needs to be actually validated)

For reference see ticket 8668

mraible:
Fixes #871 
Fixes #802

To UAT, run `export STORMPATH_ENABLED=false` and then start examples/servlet. Verify that /login and /registration result in a 404.

### DIFF
--- a/extensions/servlet/pom.xml
+++ b/extensions/servlet/pom.xml
@@ -85,6 +85,12 @@
             <artifactId>spring-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- Pulls in HttpClientRequestExecutor implementation for testing in DefaultClientLoaderListenerTest -->
+        <dependency>
+            <groupId>com.stormpath.sdk</groupId>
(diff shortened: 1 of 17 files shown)


jarias:
UAT pass, code LTGM

PR #950, opened 2016-08-31
CookieAuthenticationResultSaver.isCookieSecure(...) should complain if the current request is not secure

### LINKED ISSUE #937: CookieAuthenticationResultSaver.isCookieSecure(...) should complain if the current request is not secure
With the move to 1.X it seems that `CookieAuthenticationResultSaver.isCookieSecure(...)` by default now sets the `secure` flag of the cookies that it creates.

I.e. the generated cookies will be submitted with subsequent requests only if those requests come via `https`.

Fair enough - but shouldn't `CookieAuthenticationResultSaver.isCookieSecure(...)` first check and complain if the _current_ request is itself not via `https`?

We have an internal service that we run over `http` - when we upgraded its Stormpath Java SDK dependency version to 1.0.2. it stopped working.

Why? Because our login attempts come in via `http` - everything goes fine but the resulting `access_token` and `refresh_token` cookies have `secure` set - so when the redirect that occurs after a successful login happens the browser does not include these cookies in the redirect request (because like the login request it goes via `http`) and we just end up being redirected back to the login page again.

It actually seems a little odd that the browser itself doesn't complain (in the console at least), that a secure cookie should be set, i.e. returned from the server, via `http` is certainly weird.

There's already a dire _warning_ in `CookieAuthenticationResultSaver.isCookieSecure(...)` if things are configured such that it will produce cookies _without_ the `secure` flag.

But shouldn't there also be a dire _error_ if it's configured to set the `secure` flag but the current request is not over `https`, i.e. a setup where the combination of protocol and security settings just won't work?

jarias:
Fixes #937


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/InsecureCookieException.java
```diff
@@ -0,0 +1,15 @@
...
+ * @since 1.1.0
+ */
+public class InsecureCookieException extends RuntimeException {
+    public InsecureCookieException(String s) {
+        super(s);
+    }
+
+    @Override
+    public String getLocalizedMessage() {
+        return "stormpath.cookie.insecure.error";
```
mrioan: Please create an static constant for this hardcoded text.

extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/i18n.properties
```diff
@@ -252,3 +252,5 @@ stormpath.web.errors.13000 = Unable to connect to an SMTP server located at the
 stormpath.web.errors.13001 = Unable to authenticate to an SMTP server with the username and password provided.
 stormpath.web.errors.13002 = Unable to connect to an SMTP server using the security protocol provided.
 stormpath.web.errors.13003 = You cannot create a new email server because one already exists for your Tenant.
+
+stormpath.cookie.insecure.error = Authentication error
```
mrioan: I think `Internal configuration error during authentication attempt.` is more accurate

jarias:
Since @mrioan marked this as 1.1.0 I'm closing this PR a creating a new one that targets 1.1.x branch

PR #954, opened 2016-09-01
Default to application/json when no accept header present

### LINKED ISSUE #865: Registering in JSON without an accept header gives a weird error
We should still perform registration and pick an acceptable response even without an accept header. 

```HTTP
POST /register HTTP/1.1
Host: localhost:8080
Content-Type: application/json

{
    "givenName": "Edward1",
    "surname": "Jiang",
    "email": "edw12311@stormpath.com",
    "password": "TestTest1"
}
```

Returns me:

```HTTP
{
  "status": 400,
  "message": "Oops! Unable to register you! Please contact support."
}
```

mraible:
Fixes #865.


jarias:
Updated @mraible original impl even though it solve the problem it wasn't the correct approach, the real issue was that the ContentNegotiatingFieldValueResolver was using the UserAgent class to determine the desire content type, but it actually should be using ContentNegotiationResolver which takes into consideration the produces property to determine the correct  content type.

PR #969, opened 2016-09-09
Add support for stormpath_token grant type

### LINKED ISSUE #957: Support stormpath_token grant type
We need to make sure that we support stormpath_token grant type

https://stormpath.atlassian.net/wiki/display/AM/ID+Site+OAuth+Access+Token+Grant+Type

And add the proper authenticator according to the spec, we currently have an ID_SITE_AUTHENTICATOR that should actually be StormpathAssertionAuthenticator and we also need to adds OAuthStormpathTokenAuthenticator

https://github.com/stormpath/stormpath-sdk-spec/blob/master/specifications/authenticators.md

mraible:
Fixes #957.


mraible:
Opened against `master` instead of `1.1.x`. Closing to create a new one.

PR #926, opened 2016-08-18
Add stormpath_social grant type

mraible:
Fixes #895

To UAT, following these steps:

1. Create a Facebook directory for the application you're testing against. 
2. Use the Facebook Client ID and Client Secret to make a request to Facebook to get an access token.

    ```
    http POST "https://graph.facebook.com/{FACEBOOK_CLIENT_ID}/accounts/test-users?access_token={FACEBOOK_CLIENT_ID}|{FACEBOOK_CLIENT_SECRET}"
    ```

3. Use the access token returned to make a `stormpath_social` request to our OAuth endpoint.

    ```
    http -f POST localhost:8080/oauth/token grant_type=stormpath_social providerId=facebook accessToken=XXX
    ```

You should receive a response with an access token.

For more details, see https://stormpath.atlassian.net/wiki/display/AM/Social+Access+Token+Grant+Type. 

Questions for @omgitstom: the example response on the wiki shows a `refresh_token` and a `stormpath_access_token_href` in the response. Are those required?

PR #970, opened 2016-09-09
Add support for stormpath_token grant type

### LINKED ISSUE #957: Support stormpath_token grant type

mraible:
Fixes #957.

NOTE: This PR does not have any integration tests. I tried adding them to the TCK, but after talking with @lhazlewood, I learned the TCK does not cover ID Site. I also briefly looked into adding an IT (with RestAssured) to the SDK, but it seemed like a lot of work.

Also, I did not add any authenticators due to lack of time.


api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
```diff
@@ -0,0 +1,11 @@
+package com.stormpath.sdk.oauth;
+
+/**
+ * This class represents a request for Stormpath to authenticate an Account and exchange its ID Site token for a valid
+ * OAuth 2.0 access token. Using stormpath_token grant type.
+ *
+ * @since 1.1.0
+ */
+public interface OAuthStormpathTokenGrantRequestAuthentication extends OAuthGrantRequestAuthentication {
+    String getToken();
```
mrioan: Javadoc missing in this method

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
```diff
@@ -0,0 +1,11 @@
+package com.stormpath.sdk.oauth;
+
+/**
+ * This class represents a request for Stormpath to authenticate an Account and exchange its ID Site token for a valid
+ * OAuth 2.0 access token. Using stormpath_token grant type.
+ *
+ * @since 1.1.0
+ */
+public interface OAuthStormpathTokenGrantRequestAuthentication extends OAuthGrantRequestAuthentication {
```
mrioan: The final kind of authentication type will be `stormpath_social` when this authentication is executed. Users are interested in such token rather than in IDSite. Therefore I would explain this the other way around. I would explain WHAT this authentication is about (stormpath_social) and then I would clarify HOW this is happening behind the scenes (IDSite)

api/src/main/java/com/stormpath/sdk/oauth/Authenticators.java
```diff
@@ -128,5 +128,13 @@ private Authenticators() {
      */
     public static final OAuthStormpathSocialRequestAuthenticatorFactory OAUTH_STORMPATH_SOCIAL_GRANT_REQUEST_AUTHENTICATOR =
             (OAuthStormpathSocialRequestAuthenticatorFactory) Classes.newInstance("com.stormpath.sdk.impl.oauth.DefaultOAuthStormpathSocialRequestAuthenticatorFactory");
+
+    /**
+     * Constructs {@link OAuthStormpathTokenGrantRequestAuthenticator}s.
+     *
+     * @since 1.1.0
+     */
+    public static final OAuthStormpathTokenRequestAuthenticatorFactory OAUTH_STORMPATH_TOKEN_GRANT_REQUEST_AUTHENTICATOR =
```
mrioan: The final kind of authentication type will be `stormpath_social` when this authentication is executed. Users might find this name misleading. If this is the proper name that we want to expose then at least I would clearly document what is this Authenticator actually for: `stormpath_social`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
```diff
@@ -67,6 +69,7 @@
     private static final String CLIENT_CREDENTIALS_GRANT_TYPE = "client_credentials";
     private static final String PASSWORD_GRANT_TYPE = "password";
     private static final String STORMPATH_SOCIAL_GRANT_TYPE = "stormpath_social";
+    private static final String STORMPATH_TOKEN_GRANT_TYPE = "stormpath_token";
```
mrioan: wait... this PR is for `stormpath_social` but I see here that the new grant type is `stormpath_token`... Can you please clarify?

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void setGrantType(String grantType) {
+        setProperty(GRANT_TYPE, grantType);
+    }
+
+    @Override
+    public void setToken(String token) {
+        setProperty(TOKEN, token);
```
mrioan: assert not null

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
```diff
@@ -0,0 +1,50 @@
...
+        super(dataStore);
+    }
+
+    public DefaultOAuthStormpathTokenGrantAuthenticationAttempt(InternalDataStore dataStore, Map<String, Object> properties) {
+        super(dataStore, properties);
+    }
+
+    @Override
+    public void setGrantType(String grantType) {
+        setProperty(GRANT_TYPE, grantType);
```
mrioan: the sole appropriate grantType must be an `stormpath_token` therefore please assert that the parameter is always `stormpath_token`. Additionally it should not be `public` (`protected` at the minimum), it should be just invoked internally when this class is created.

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathTokenGrantAuthenticationAttempt.java
```diff
@@ -0,0 +1,20 @@
+package com.stormpath.sdk.impl.oauth;
+
+import com.stormpath.sdk.resource.Resource;
+
+/**
+ * @since 1.1.0
+ */
+public interface OAuthStormpathTokenGrantAuthenticationAttempt extends Resource {
```
mrioan: Please add Javadoc to the class

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathTokenGrantAuthenticationAttempt.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * Method used to set the Authentication Grant Type that will be used for the token exchange request.
+     * @param grantType the Authentication Grant Type that will be used for the token exchange request.
+     */
+    void setGrantType(String grantType);
```
mrioan: Not clear to me what this method is for: if this interface is an StormpathToken Grant Type then the token cannot be anything other than an `stormpath_token`, therefore this method is error-prone, misleading and redundant. It should just be removed from the interface.

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
mrioan: Add `@see OAuthStormpathTokenGrantRequestAuthentication` to `OAuthRequestAuthentication`

mraible:
@mrioan I made a mistake in the title of this PR. It should be for `stormpath_token` grant type, not `stormpath_social`. Support for `stormpath_social` was added in https://github.com/stormpath/stormpath-sdk-java/pull/926.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
mraible: @mrioan This PR is for `stormpath_token`. I had a typo in the PR's name. Sorry! ;)

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
mraible: According to https://stormpath.atlassian.net/wiki/display/AM/ID+Site+OAuth+Access+Token+Grant+Type, the parameter name is `token`.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
mraible: I copied this pattern from other `DefaultOAuth*GrantAuthenticationAttempt` classes. They don't assert that properties are not null. For example:

[DefaultOAuthPasswordGrantAuthenticationAttempt](https://github.com/stormpath/stormpath-sdk-java/blob/master/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthPasswordGrantAuthenticationAttempt.java)
[DefaultOAuthClientCredentialsGrantAuthenticationAttempt](https://github.com/stormpath/stormpath-sdk-java/blob/master/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthClientCredentialsGrantAuthenticationAttempt.java)

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
mrioan: @mraible This comment has not yet been addressed

mraible: @mrioan See https://github.com/stormpath/stormpath-sdk-java/pull/970/files#diff-3c45b37aed8410f83f9a0d0ef4e0930cR26

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
mrioan: I still think that mentioning IDSite is actually confusing and not necessary. The users will be interested in knowing that this request represents a `stormpath_token` request, whether IDSite is used internally or not us just an implementation detail. You can mention it if you want as a clarification but it should be in second place, it is not the first message we want users to read. Additionally, the description is a little bit vague. Can you please clarify how this token was obtained/generated. It is a little bit confusing even for me

mraible: I can change the Javadoc, but isn't IDSite the only place you can get a `stormpath_token` currently? I implemented this feature and I'm still a bit confused by it, so I'm not surprised you are! I still haven't discovered how to UAT this because I don't know how to obtain a `stormpath_token` to use for testing.

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathTokenGrantAuthenticationAttempt.java
mrioan: @mraible I see, well I still think that this is not right. We cannot remove that method from the existent code but we can avoid having this issue replicate to new authenticators like this one. Therefore it should be removed. Setting the tokenType is redundant and error prone. The authentication class already knows the grant type that it represents. It should just be hardcoded

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathTokenGrantRequestAuthentication.java
mrioan: These comments have not been addressed

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
```diff
@@ -286,6 +289,29 @@ private AccessTokenResult stormpathSocialAuthenticationRequest(HttpServletReques
...
+     * @since 1.1.0
+     */
+    private AccessTokenResult stormpathTokenAuthenticationRequest(HttpServletRequest request, HttpServletResponse response) {
+        OAuthGrantRequestAuthenticationResult authenticationResult;
+
+        try {
+            Application app = getApplication(request);
+            String token = request.getParameter("token");
+            OAuthStormpathTokenGrantRequestAuthentication grantRequestAuthentication =
+                    new DefaultOAuthStormpathTokenGrantRequestAuthentication(token);
```
mrioan: There must be an authentication request available in `OAuthRequests`. The line `new DefaultOAuthStormpathTokenGrantRequestAuthentication(token)` must not be used. People must do:

```
OAuthStormpathTokenGrantRequestAuthentication grantRequestAuthentication = OAuthRequests.OAUTH_STORMPATH_TOKEN_GRANT_REQUEST.builder().setToken(token).build();
```

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathTokenGrantAuthenticationAttempt.java
mraible: OK, just to verify I'm doing the right thing - here's a patch that shows this change. Is it correct? https://gist.github.com/mraible/6658a749814fb92058be44533fd0782b

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
mrioan: I was referring to `grantType`: `grant_type=stormpath_token`. Anyways, this method should be removed from the interface as requested in upper comments. I am not sure if you can remove it from the implementation class as well but if you can just hard-code this value and remove this method it would be great.

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathTokenGrantAuthenticationAttempt.java
mrioan: ok, thanks

mrioan:
@mraible WRT `IDSITE_AUTHENTICATION_REQUEST` please add a description to its javadoc saying that it is deprecated because its functionality is redundant to what is provided via `OAUTH_STORMPATH_TOKEN_GRANT_REQUEST` and will be removed in `2.0.0`

mraible:
@mrioan I'm happy to make the requested changes. However, I'd rather have steps to UAT this implementation so I know it works. Until I have that, I'm just changing code and don't have any idea if it accomplishes the required goal.

mrioan:
@mraible lol, we are both flying blind :)

Is it me or this operation is already implemented as an `ID_SITE_AUTHENTICATOR`?

Take a look at `CookieAuthenticationResultSaver` you will see there that it constructs an authenticator that will hit IDSite but first it creates the `jwt` (that is the piece of code you need to test your implementation). The part that I am missing is to be 100% sure whether the `ID_SITE_AUTHENTICATOR` is creating a `grant_type=stormpath_token`. I do think that is the case as shown in `DefaultCreateIdSiteAuthenticationAttemptTest#testMethods`. If you find out that the `ID_SITE_AUTHENTICATOR` is actually doing the very same thing that you are trying to do then this issue can be considered implemented and this PR can be dismissed (unless you know for sure that the name of the authenticator should not be `ID_SITE_XXX` but `OAUTH_STORMPATH_TOKEN_GRANT_XXX`

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathTokenGrantAuthenticationAttempt.java
mraible: ^^ @mrioan I tried the above change and it fails because no `grant_type` is set. We should probably create an issue to refactor all authenticators to remove the grant_type. For now, I just want to get things working. :)

mrioan: @mraible Yes, totally it will fail, it is a required value. You have to set it internally yourself, I guess in the constructor or always returning a hardcoded value in the get (not sure how it needs to be read). It must not be open in the API. It is confusing and error prone. If you feel you need help I can give it a try. Thanks

mrioan:
@mraible I did some improvements to the code and it all seems to be working fine. Do you think you can take a look at it @mraible when you have a min please? Thanks

mraible:
@mrioan I tested this using the following steps:

Set environment variables:

```
export STORMPATH_APPLICATION_HREF=https://api.stormpath.com/v1/applications/3QIMlJKKN2whGCYzXXw1t8
export STORMPATH_API_KEY_FILE=/Users/mraible/.stormpath/apiKey-elastic-rebel.properties
```

Started `stormpath-sdk-java/examples/spring-security-spring-boot-webmvc` with `mvn spring-boot:run`. 

1. Go to http://shielded-journey-8142.herokuapp.com/
2. Login with Tom's user/password.
3. Grab the jwtResponse query parameter value and use it in the following POST request

```
http --form POST http://localhost:8080/oauth/token \
 content-type:application/x-www-form-urlencoded \
 grant_type=stormpath_token \
 token=TOKEN_GOES_HERE
```

Result is a 500 error / HTML page with the following message:

```
Unable to invoke Stormpath controller: Unable to instantiate instance with constructor 
[public com.stormpath.sdk.impl.oauth.DefaultAccessToken(com.stormpath.sdk.impl.ds.InternalDataStore,
java.util.Map)]
```

Server log has:

```
Caused by: io.jsonwebtoken.JwtException: JWT failed validation; it cannot be trusted.
        at com.stormpath.sdk.impl.oauth.DefaultAccessToken.ensureAccessToken(DefaultAccessToken.java:55)
        at com.stormpath.sdk.impl.oauth.DefaultAccessToken.<init>(DefaultAccessToken.java:34)
        ... 47 common frames omitted
```

mrioan:
@mraible if I generate an IDSite token [like this](https://github.com/stormpath/stormpath-sdk-java/blob/957-stormpath-token-grant-type/extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy#L2088) and [use it following your instructions](https://github.com/stormpath/stormpath-sdk-java/pull/970#issuecomment-252091524) then it works just fine for me. HOWEVER, I do not get a `refresh_token`, only an `access_token`. We need more clarification about this feature. [These are the open questions](https://github.com/stormpath/stormpath-sdk-java/issues/957#issuecomment-252267622). Do you have more? We need @omgitstom's help to answer them.

josebarrueta:
This was already handled by #1127

mraible:
Closing per @josebarrueta's comment: 

> This was already handled by #1127

Issue #351, opened 2015-11-13
Implement Organizations / multi-tenant for the Spring stack


dogeared:

dogeared:
We need a filter that uses heuristics to determine which (if any) Organization the user is associated with.

The Organization will likely be baked into JWTs associated with the session.

jarias:
@dogeared Is this something that is ready to be work on or it still requires some design or specification for this. Also we already have DefaultOrganizationNameKeyResolver

PR #973, opened 2016-09-13
Add support for organizationNameKey for oauth token exchange

### LINKED ISSUE #742: Add support for organiationNameKey for oauth token exchange
https://stormpath.atlassian.net/browse/AM-3287

jarias:
Fixes #742

### DIFF
--- a/extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/oauth/OAuthPasswordGrantRequestAuthenticatorIT.groovy
+++ b/extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/oauth/OAuthPasswordGrantRequestAuthenticatorIT.groovy
@@ -1,0 +1,66 @@
+package com.stormpath.sdk.impl.oauth
+
+import com.stormpath.sdk.account.Account
+import com.stormpath.sdk.application.Application
+import com.stormpath.sdk.application.Applications
+import com.stormpath.sdk.client.ClientIT
+import com.stormpath.sdk.directory.Directory
+import com.stormpath.sdk.impl.application.DefaultApplication
+import com.stormpath.sdk.organization.Organization
+import com.stormpath.sdk.organization.Organizations
+import org.testng.annotations.AfterMethod
+import org.testng.annotations.BeforeMethod
+import org.testng.annotations.Test
+
+import static org.testng.Assert.assertEquals
+
+/**
+ * @since 1.1.0
+ */
+class OAuthPasswordGrantRequestAuthenticatorIT extends ClientIT {
+    static final String TEST_PASSWORD = 'Changeme1'
+
+    Application testApplication;
+    Account testAccount;
+    Organization testOrganization;
+
+    @BeforeMethod
+    void setup() {
+        testApplication = client.instantiate(Application)
+        testApplication.name = uniquify("Java SDK IT oauth app")
+        client.currentTenant.createApplication(Applications.newCreateRequestFor(testApplication).build())
+
+        testOrganization = client.instantiate(Organization)
+        testOrganization.name = uniquify("Java SDK IT oauth org")
+        testOrganization.nameKey = uniquify("oauthIT")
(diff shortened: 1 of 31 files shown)


extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/oauth/OAuthPasswordGrantRequestAuthenticatorIT.groovy
```diff
@@ -0,0 +1,66 @@
+package com.stormpath.sdk.impl.oauth
```
lhazlewood: header

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/DefaultWrappedServletRequestFactory.java
```diff
@@ -34,24 +35,28 @@
...
                                                Saver<AuthenticationResult> authenticationResultSaver,
                                                Publisher<RequestEvent> eventPublisher,
+                                               Resolver<String> organizationNameKeyResolver,
                                                String userPrincipalStrategyName, String remoteUserStrategyName) {
         this.usernamePasswordRequestFactory = factory;
         this.authenticationResultSaver = authenticationResultSaver;
         this.eventPublisher = eventPublisher;
         this.userPrincipalStrategyName = userPrincipalStrategyName;
         this.remoteUserStrategyName = remoteUserStrategyName;
+        this.organizationNameKeyResolver = organizationNameKeyResolver;
```
lhazlewood: I may not understand the scope of this, but I don't think the request should know anything about org name keys - just the account store that is relevant for the request (obtained by an `AccountStoreResolver`).

In other words, I think the AccountStoreResolver is the primary component and it in turn delegates to an OrganizationNameKeyResolver to look up an org name key to use when finding the corresponding Organization (i.e. AccountStore).

Does this make sense?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/http/impl/StormpathHttpServletRequest.java
```diff
@@ -101,11 +102,13 @@
     private final Publisher<RequestEvent> eventPublisher;
     private final String userPrincipalStrategyName;
     private final String remoteUserStrategyName;
+    private final Resolver<String> organizationNameKeyResolver;
```
lhazlewood: Same - probably want to reference the relevant AccountStore, not the resolver, no?

lhazlewood: Also, I'd assume that the org (account store) would be resolved early in the request, presumably via a servlet filter, and once discovered, set on the request (i.e. `request.setAttribute("accountStore", organization);` )

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/DefaultWrappedServletRequestFactory.java
jarias: Makes sense. I'll check if the AccountStoreResolver really delagates to OrganizationNameKeyResolver

jarias: But does the API token endpoint checks if the account store is an org and sets the org claim in the resulting access token?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/http/impl/StormpathHttpServletRequest.java
jarias: Looks like we don't have a valid impl of an account store resolver nor a filter, I think @mrioan was working on this for ticket https://github.com/stormpath/stormpath-sdk-java/issues/351 may this needs to be finish before we can wrap this one out.

PR #980, opened 2016-09-16
Ensure that aws is in the path for coverage report

dogeared:


dogeared:
@jarias - I think there may still be an issue with the docs. Check out: https://travis-ci.org/stormpath/stormpath-sdk-java/builds/160286850

It's reporting:

```
-------> Build configuration: 
Version:             1.0.4
Is relase:           true
Build documentation: false
Running IT tests:    true
```

This was during the release. It set "is release" properly, but "build documentation" is showing as false.

Also, just so you know, Travis is a little finicky. If you generate too much output, it kills your build. By the same token, if you don't generate any output for a period of time, it kills your build. So, I refactored your original scripts to put whatever's going on in the background and to echo a status every 20 seconds.

jarias:
@dogeared I'll take a look at it

jarias:
-------> Build configuration: 
Version:             1.0.4
Is relase:           true
Build documentation: true
Running IT tests:    false

From the same build but on JDK 8

I think the configuration is correct.

PR #983, opened 2016-09-19
update TCK run(s) in the project to fail the build

### LINKED ISSUE #930: update TCK run(s) in the project to fail the build
Right now, when the TCK runs, failures are logged, but it does not fail the build. Now that we've achieved stability with the TCK, failures should fail the build.

Also, we should checkout the configured tag from the TCK so that changes or updates to master don't affect our ability to build.

jarias:
Fixes #930

Also fix some exit commands in build and install scripts


jarias:
This was already review by Les.

PR #991, opened 2016-09-21
Using our Sphinx Theme

rdegges:
Yo dudes.

As you may know, we have a centrally maintained Sphinx theme for use in all of our docs: https://github.com/stormpath/stormpath-sphinx-theme

This PR updates your Sphinx docs (for both the Java SDK, as well as your Spring boot integration) to use this central theme via git submodules.

This doesn't change any behavior or results, except that:

- This is much easier to maintain. When we roll out updates to the theme in the future, you can just update the submodule and your docs will reflect the latest changes with no code changes on your part.
- The theme now includes all necessary tracking code, which means you no longer have to worry about manually including tracking code stuff.
- There is less code in your repo now, since we're outsourcing lots of code to the submodules.

Let me know if you have any questions. I already tested this out extensively locally.


lhazlewood:
Awesome! 👍

PR #1021, opened 2016-10-06
The sdk throws foreign JWT exception

### LINKED ISSUE #1018: the sdk throws foreign JWT exception
The sdk contains `com.stormpath.sdk.error.jwt.InvalidJwtException` but in multiple places `io.jsonwebtoken.JwtException` is thrown.

I guess it would be cleaner for the sdk to only throw Stormpath (or general java) exceptions? The fact that you internally are using the io.jsonwebtoken.jwt library should not be visible to the users?

Also you could provide a bit more info to the user on what actually went wrong? I tried to reproduce a case where we see this JwtException from time to time and failed to do so because it could be a lot of things.

https://github.com/stormpath/stormpath-sdk-java/blob/9590dcaa21d7ba2ae46ca05108f87f9cab4e6b79/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java#L94
https://github.com/stormpath/stormpath-sdk-java/blob/9590dcaa21d7ba2ae46ca05108f87f9cab4e6b79/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java#L108
https://github.com/stormpath/stormpath-sdk-java/blob/3b5d68792cd60f413427505c7bd7772b95406d77/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultAccessToken.java#L55

jarias:
Fixes #1018

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java
@@ -124,8 +124,11 @@ public class DefaultOAuthBearerRequestAuthenticator extends AbstractOAuthRequest
                 String message = "The character encoding for the API secret is not supported.";
                 log.debug(message);
                 throw new InvalidJwtException(message, uee);
+            } catch (InvalidJwtException e) {
+                //See https://github.com/stormpath/stormpath-sdk-java/issues/1018
+                log.debug(e.getMessage());
+                throw e;
             } catch (Exception e) {
-                //Todo 2.0.0: this JwtException must be replaced by InvalidJwtException. See https://github.com/stormpath/stormpath-sdk-java/issues/1018
                 throw new JwtException("JWT failed validation; it cannot be trusted.", e);
             }
         }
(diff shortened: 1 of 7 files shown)


impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java
```diff
@@ -29,7 +30,10 @@
 import io.jsonwebtoken.Claims;
 import io.jsonwebtoken.JwtException;
```
mraible: This import should be removed.

mrioan:
@lhazlewood not sure about this... 

Is a change in a runtime exception a braking change? I am asking to know whether this should be `2.0.0` or can be `1.2.0`. Please see this as an example: https://github.com/stormpath/stormpath-sdk-java/pull/1021/files#diff-c1b258849dab92454ba040df2550eb84R71. You will see that I am keeping the same `JwtException` in case someone is catching it but they will be lacking the new ones (I am marking it for removal in 2.0.0). Thanks!

mrioan:
This functionality is not critical and will have impact on existent developer's code (i.e. existent code will fail to catch the new exceptions while they are catching them all just fine now). We have therefore decided to postpone this PR until we do 2.0.0

mrioan:
Hey @dogeared Please take [this comment](https://github.com/stormpath/stormpath-sdk-java/pull/1021#issuecomment-254926576) into account. This should not be released until 2.0.0

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthBearerRequestAuthenticator.java
dogeared: changed `JWT_INVALID_VALUE_ERROR` to `UNSUPPORTED_JWT_ERROR`

dogeared: @mraible - Would you please give the 👍  and I will merge this into the 2.x branch? Thanks.

PR #1023, opened 2016-10-06
1020 adding emailVerificationStatus to account

### LINKED ISSUE #1020: Add emailVerificationStatus to Account
There's a new property available on REST Account for the emailVerificationStatus.  This should be exposed in the SDK as well.  Valid values are `VERIFIED`, `UNVERIFIED` and `UNKNOWN`.  (`UKNOWN` is for accounts that were created before there was a separate value for email verification, so we can't be sure of the verification status.)

mzumbadoate:
fixes #1020


jarias:
LGTM but don't know if we should merge just yet since the next release is 1.1.1 and this targets 1.2.0, also check CI to why the build is failing.

dogeared:
This is failing CI currently because of the dependency on the update to the framework TCK

PR #1028, opened 2016-10-10
1025 fixing NPE in cookie account resolver

### LINKED ISSUE #1025: NullPointerException when trying to delete token in CookieAccountResolver
There is a bug in this line: https://github.com/stormpath/stormpath-sdk-java/blob/master/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java#L146

When the `access_token` or `refresh_token` are not present the `delete` operation causes a `NullPointerException`.

`Delete` operation must check cookie is not null.

See https://stormpath.zendesk.com/agent/tickets/9582

mzumbadoate:
fixes #1025

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
@@ -106,7 +106,7 @@ public class CookieAccountResolver implements Resolver<Account> {
     }
 
     protected void deleteCookie(HttpServletResponse response, Cookie cookie) {
-        if (!response.isCommitted()) {
+        if (!response.isCommitted() && cookie != null) {
             cookie.setValue("");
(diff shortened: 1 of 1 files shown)


jarias:
LGTM but we need to fix the build issue we discuss yesterday in Slack.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
```diff
@@ -106,7 +106,7 @@ protected Account getAccount(HttpServletRequest request, HttpServletResponse res
     }
 
     protected void deleteCookie(HttpServletResponse response, Cookie cookie) {
-        if (!response.isCommitted()) {
+        if (!response.isCommitted() && cookie != null) {
```
dogeared: @mzumbadoate - please just add a comment referencing the issue this change resolves. Then, it's ready to merge!

PR #1036, opened 2016-10-11
1033 adding new spec properties

### LINKED ISSUE #1033: Add support for Framework Spec multi-tenancy
Per: 

https://github.com/stormpath/stormpath-framework-spec/blob/master/example-config.yaml#L10

And

https://github.com/stormpath/stormpath-framework-spec/blob/master/example-config.yaml#L113

And

https://github.com/stormpath/stormpath-framework-spec/blob/master/example-config.yaml#L181

we need to implement support for these multi-tenancy related properties.

Also, NOTE: the `SpecConfigVersusWebPropertiesTest#verifyPropertiesInSpecAreInDefault` *must* be re-enabled. It was temporarily disabled to avoid test failures related to property support differences.

mzumbadoate:
fixes #1033

### DIFF
--- a/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
+++ b/extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
@@ -18,6 +18,9 @@
 stormpath.application.href =
 stormpath.application.name =
 stormpath.web.basePath =
+stormpath.web.baseDomain =
+stormpath.web.multiTenancy.enabled = false
+stormpath.web.multiTenancy.useSubdomain = true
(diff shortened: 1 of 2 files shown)


dogeared:
This needs to go beyond just defining the properties. It should also make use of them in all the integrations.

jarias:
There is also https://github.com/stormpath/stormpath-sdk-java/issues/351 and https://github.com/stormpath/stormpath-sdk-java/issues/742 that where started before there was a spec around multi tenancy, they are assign to @lhazlewood cause he wanted to review the implementation.

mzumbadoate:
@jarias do you think this should be labeled as blocked by #351 and #742?

PR #1048, opened 2016-10-18
Change /me endpoint so groups.items is returned as a collection instead of groups

### LINKED ISSUE #1044: /me endpoint should return a group object with an items array
In the Java SDK, /me (with groups expansion) returns groups data as follows:

```
    "groups" : [ {
      "href" : "https://api.stormpath.com/v1/groups/2HWD2PtXm3WwAsHtthnF5M",
      "createdAt" : "2016-06-30T19:53:19.232Z",
      "modifiedAt" : "2016-10-11T04:48:59.951Z",
      "name" : "ADMIN",
      "description" : null,
      "status" : "ENABLED"
    } ]

```

The Angular SDK has an `if-user-in-group` directive that expects "groups" to be an object and it to have the group items in an "items" collection. 

When I use `if-user-in-group="ADMIN"`, I see the following error in my console:

```
angular.js:13920 TypeError: Cannot read property 'filter' of undefined
    at User.inGroup (stormpath-sdk-angularjs.js:2895)
    at User.groupTest (stormpath-sdk-angularjs.js:2915)
    at evalElement (stormpath-sdk-angularjs.js:1016)
    at stormpath-sdk-angularjs.js:1029
    at Scope.$digest (angular.js:17524)
    at Scope.$apply (angular.js:17790)
    at done (angular.js:11831)
    at completeRequest (angular.js:12033)
    at XMLHttpRequest.requestLoaded (angular.js:11966)
```

If I look at line 2895 in `stormpath-sdk-angularjs.js`, I see the following function being called:

```js
  /**
  * This method may change in the future, do not use.
  * Please use the `ifUserInGroup` directive instead
  */
  User.prototype.inGroup = function inGroup(groupName) {
    return this.groups.items.filter(function(group){
      return group.name === groupName;
    }).length >0;
  };
```

If I debug, `this.groups` is defined, but `this.groups.items` is not.

mraible:
Fixes #1044.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultAccountModelFactory.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/DefaultAccountModelFactory.java
@@ -4,8 +4,8 @@ import com.stormpath.sdk.account.Account;
 import com.stormpath.sdk.impl.account.DefaultAccount;
 import com.stormpath.sdk.impl.resource.AbstractResource;
 import com.stormpath.sdk.impl.resource.DateProperty;
-import com.stormpath.sdk.impl.resource.Property;
 import com.stormpath.sdk.impl.resource.EnumProperty;
+import com.stormpath.sdk.impl.resource.Property;
 import com.stormpath.sdk.impl.resource.StringProperty;
 import com.stormpath.sdk.lang.Assert;
 import com.stormpath.sdk.lang.Strings;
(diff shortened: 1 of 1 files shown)


mraible:
To verify this works, add the following to `examples/spring-boot-webmvc/src/main/resources/application.properties`

```
stormpath.web.me.expand.groups = true
```

Start the application and verify the /me endpoint returns `groups.items`:

```json
{
   "account":{
      "groups":{
         "items":[]
      }
   }
}
```

Before this change, the following JSON was returned:

```json
{
   "account":{
      "groups": []
   }
}
```

mrioan:
@mraible The way I see it, this is a breaking change for developers interacting with us via JSON. If so, then It should not be released until `2.0.0`. 

I think this change will have veerry little impact (if any). However I do not want to get this approved just myself :), we should all agree on breaking [semver](https://github.com/stormpath/stormpath-sdk-java/wiki/Development-Process-Explained#semantic-versioning) in this PR.

mraible:
@mrioan Is there anyway to look at our current clients and see how many are using group expansion in their API calls to /me?

mrioan:
@mraible not that I am aware of. We could see a request with expansion coming to the backend in the logs but we will not know whether it was rendered as JSON or HTML by the Java SDK

mraible:
Closing to target other branch.

Issue #1041, opened 2016-10-13
Support relay state for SAML (and IDSIte?) as issue #849 originally requested


mrioan:
Issue #849 was wrongly implemented using the state property instead of a adding a new RelayState property (or relayState, not sure which is the actual claim that IDSite and SAML are expecting)

PR #1060, opened 2016-10-21
Undoing changes introduced by issue 849

### LINKED ISSUE #1040: Bug in state property
#849 wrongly used the state property instead of a adding a new RelayState property (or relayState, not sure which is the actual claim that IDSite and SAML are expecting)

The implementation of issue 849 has broken the functionality described here and here.

The Java SDK will redirect you to whatever was set in state. This means that if you set state = blah then after a successful authentication your application will try to open http://localhost:8080/blah

The changes introduced in #849 must be undone.

Additionally, the relay state must be implemented properly. However this will be done in issue https://github.com/stormpath/stormpath-sdk-java/issues/1041

mrioan:
Fixes #1040 

This fixes an issue where the `state` property was wrongly treated as `next` uri.

### DIFF
--- a/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
+++ b/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
@@ -83,7 +83,7 @@ class SpecConfigVersusWebPropertiesTest {
             specProperties.containsKey(k) ? null : k
         }
 
-        def expected_diff_size = 82
+        def expected_diff_size = 81
 
         if (diff.size != expected_diff_size) {
             println "It looks like a property was added or removed from the Framework Spec or web.stormpath.properties."
--- a/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
+++ b/extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
@@ -52,10 +52,10 @@ class SpecConfigVersusWebPropertiesTest {
     }
 
     /**
-     * NOTE: This test is temporarily disabled as new properties have been added to the framework spec for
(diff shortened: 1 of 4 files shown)


extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
```diff
@@ -74,7 +74,9 @@ class SpecConfigVersusWebPropertiesTest {
             println "Or you could adjust the assertEquals statement in this method to allow for this missing key as a temporary solution."
         }
 
-        assertEquals 0, diff.size(), "Missing keys in default config: ${diff}"
+        //todo: 15 new properties related to organizations were added to the spec, we do not yet suppor them.
+        //see https://github.com/stormpath/stormpath-sdk-java/issues/1052
+        assertEquals 15, diff.size(), "Missing keys in default config: ${diff}"
```
dogeared: @mrioan - didn't we (temporarily) disable the test that makes this an issue? I like this approach better, but I just want to make sure that the test is re-enabled.

dogeared:
@mrioan - should the original logic be used for `relayState` so that we cover the original intent of #849? Or, has that been handled elsewhere?

mrioan:
@dogeared The original relayState feature is going to be handled by https://github.com/stormpath/stormpath-sdk-java/issues/1041. 
I divided #849 into #1040 and #1041 in order to get the state bug solved fast as there is a customer impacted by it. See ticket 9583.

extensions/servlet/src/test/groovy/com/stormpath/sdk/servlet/config/SpecConfigVersusWebPropertiesTest.groovy
mrioan: @dogeared I can disable it, no problem at all. I just copied what I did in another branch and I did not realize this test was disabled in 1.1.2. I will undo this change now. Thanks!

dogeared:
LGTM @mrioan! Please merge after CI

PR #1062, opened 2016-10-24
Add support for ignoring regex paths in StormpathFilter

### LINKED ISSUE #1031: Stormpath filter gets hit for static assets (e.g. JavaScript files), causing performance issues
I found this issue while integrating Stormpath into a JHipster application. To reproduce, you can run https://github.com/mraible/jhipster-stormpath-example. I've also created https://github.com/mraible/jhipster-stormpath-example/issues/10 to track this as part of that project.

Without Stormpath enabled (in `SecurityConfiguration.java`), most JS files take under 100ms to load. The app finishes loading in 1.55s. When Stormpath is enabled, some JS files take over 2 seconds to load. The app finishes loading in 4.3s, and the server logs are filled with:

```
2016-10-07 14:20:55.961 DEBUG 71718 --- [nio-8080-exec-6]
 enticationSpringSecurityProcessingFilter : No token in request, will continue chain.
```

Adding `<logger name="com.stormpath.spring" level="WARN"/>` to quiet the logs does not improve load times.

Level of concern: high. This a bad developer experience because you have to wait for the files to load if you have caching turned off in Chrome Developer Tools.

mraible:
Fixes #1031.


dogeared:
Closing as may no longer be an issue. Will put to the backlog, just in case.

PR #1064, opened 2016-10-25
Fix for next parameter with context path

### LINKED ISSUE #1061: Next parameter contains context path
We have a bug reported by a user in StackOverflow:

```
I'm trying to add the Stormpath login workflow to my Spring Boot app, running under a context of /api which I've defined in my application.properties with server.context-path=/api.

When I hit a restricted path, I get the Stormpath login page with a "next" request param of http://localhost:8080/api/login?next=%2Fapi%2F. After a successful login, this causes my context to be repeated: http://localhost:8080/api/api/ . Is there a way for me to get around this?

I'm using stormpath-default-spring-boot-starter version 1.1.1
```

See: http://stackoverflow.com/questions/40196148/stormpath-spring-boot-integration-context-repeats

mrioan:
resolves #1061


mraible:
LGTM 👍

PR #1069, opened 2016-10-25
Updated all examples to be constructor injection rather than field injection.

### LINKED ISSUE #1068: Change examples and tutorials to constructor injection
per spring engineer: "field injection is evil" http://olivergierke.de/2013/11/why-field-injection-is-evil/

dogeared:
Fixes #1068


examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
```diff
@@ -29,9 +29,13 @@
 @Controller
 public class HelloController {
 
-    @Autowired
     private HelloService helloService;
```
mrioan: Can't we just keep the `@Autowired` field and add a `notNull` assertion  in a constructor with NO arguments?. The outcome will be the same with less code to read, and less impact to the interface (as a new contractor with parameters requires changes in the code that creates it).

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
```diff
@@ -29,9 +29,13 @@
...
 
+    @Autowired
+    public HelloController(HelloService helloService) {
+        this.helloService = helloService;
```
mrioan: This is missing a `notNull` assertion

examples/spring-security-spring-boot-webmvc/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
```diff
@@ -30,9 +30,13 @@
 @Controller
 public class HelloController {
 
-    @Autowired
     private HelloService helloService;
 
+    @Autowired
+    public HelloController(HelloService helloService) {
+        this.helloService = helloService;
```
mrioan: Missing a `notNull` assertion. There are more like this below. I will not flag them all so please take a look at the other ones yourself. Thanks !

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
dogeared: According to the article, having an explicit constructor improves testability as you can instantiate the object and pass in a mock as the injected dependency.

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
dogeared: It's not necessarily required to have the not null check. Do you think it's important that they all have the check?

examples/spring-boot-default/src/main/java/com/stormpath/spring/boot/examples/HelloController.java
mrioan: I kinda agree with you BUUUUT we are simply providing examples for People to use our product. From my point of view the examples need to be as readable as possible. A  working example that is simple to read is heaven to me. I do not think we should we decreasing readability in favor of following a good practice that we do not really need as those examples are not even Unit Tested (therefore the constructor is never really customized by tests). They are just run by ITs. To be honest I do not think we should be changing tests to follow this practices, it adds complexity and provides no real gain.

dogeared:
@mrioan - per conversation, only the tutorial section has been changed to use constructor injection

mrioan:
@dogeared approved. Back to you. Thanks!

PR #1070, opened 2016-10-26
issue 1059  - added AccessTokenControllerConfig to centralize the var…

### LINKED ISSUE #1059: OAuth grant type settings not being honored
Both `client_credentials grant` and `password grant` are still functional although they are configured as disabled.

Steps to reproduce:

Launch one of the test apps, with the following set (this is supposed to be disabling those grant types): 
```
stormpath.web.oauth2.password.enabled = false
stormpath.web.oauth2.client_credentials.enabled = false
```

However, both a `password` and `client_credentials` grant type request to `/oauth/token` will still be functional. This is a bug.

iancorcoran:
Fixes #1059

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/Config.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/Config.java
@@ -40,6 +40,7 @@ import com.stormpath.sdk.servlet.i18n.MessageSource;
 import com.stormpath.sdk.servlet.idsite.IdSiteOrganizationContext;
 import com.stormpath.sdk.servlet.mvc.RequestFieldValueResolver;
 import com.stormpath.sdk.servlet.mvc.WebHandler;
+import com.stormpath.sdk.servlet.util.GrantTypeStatusValidator;
 
 import javax.servlet.ServletException;
 import java.util.List;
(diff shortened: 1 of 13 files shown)


iancorcoran:
Fixes #1059

iancorcoran:
Thanks for the quick review @mrioan 

Added the servlet support for the grant type status validator, integration & example app will run fine now.

iancorcoran:
@mrioan This is working as designed.

Similar to how today we throw an OAuthException with an error code of UNSUPPORTED_GRANT_TYPE if we don't recognise the grant type (see AccessTokenController around line 345), we should throw the same exception if the grant type is valid but has been explicitly disabled. Per the framework spec, the response should be a 400 with a body of:

`{
  "error": "unsupported_grant_type"
}`

mrioan:
@iancorcoran I think I was not clear enough sorry. I do think that throwing an exception is just fine, I agree with you. However I do not like the idea about **not catching** it and **having it visible in the console**. 

Conceptually speaking invoking a disabled piece of functionality is not an error, it is totally valid and expected, therefore the application should not log an exception. 

Keep in mind that these endpoints will be open for anyone to invoke them from the outside world. A developer will be seeing exceptions in his log anytime an external party tries to authenticate with a disabled grant type. Agree? Back to you

iancorcoran:
@mrioan Fixed bug with OAuthException and now just logging error message rather than stack trace

PR #1086, opened 2016-10-31
social login flow #1022

### LINKED ISSUE #1022: New social login endpoints
See Stormpath's [Social Login V2](https://stormpath.atlassian.net/wiki/display/AM/Social+Login+V2#SocialLoginV2-Social_Flow) for detailed requirements.

jhericks:
Resolves #1022

The issue is the new [social login flow (v2)](https://github.com/stormpath/stormpath-sdk-java/issues/1022).  This is based on this spec in Confluence: [Social Login V2](https://stormpath.atlassian.net/wiki/display/AM/Social+Login+V2#SocialLoginV2-Social_Flow).

BTW, my intention was to target this for 1.2, so I was told to branch from master, but it turns out master is 1.1.3 right now.  Not sure the proper approach here so need advice on that.


dogeared:
@jhericks - master is correct for 1.2. Master has been updated to reflect `1.2.0-SNAPSHOT`.

I also updated the beginning of your comment to group the PR with the issue in waffle.io

jhericks:
I did branch from master.  I also merged from master recently, which is when I think those came in.

extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
```diff
@@ -158,6 +158,10 @@ stormpath.web.social.google.uri = /callbacks/google
 stormpath.web.social.google.scope = email profile
```
mrioan: Scopes should be comma-delimited (instead of space-delimited) just for the sake of keeping syntax consistency to other properties supporting lists

jhericks: This property is not used by me (or anyone else that I can find).  Also, in case it _is_ used, space-delimited is the OAuth2 standard, so if we had it comma-delimited here, we would have to tokenize it and then reassemble it to space-delimited if we actually wanted to use it.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/provider/ExternalAccountStoreModelFactory.java
```diff
@@ -85,7 +86,10 @@ public void visit(Directory directory) {
             Provider provider = directory.getProvider();
             ProviderModel providerModel = null;
 
-            if (provider instanceof OAuthProvider) {
```
jhericks: This is done now.

extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
mrioan: 1) If it is not used then please remove them
2) How are the scopes going to be configured by the developer
3) OAuth 2 standard: good catch, then totally fine with me if it is the standard way!

jhericks: In answer to 2), the developer configures it on the provider itself (in the admin console or through the API).

But in answer to 1), looking again, I don't think they are used, but that could be a bug? Look at the [framework spec for login](https://github.com/stormpath/stormpath-framework-spec/blob/d7991d80b0464a37556a71dc15a5bb51521d5554/login.md) it seems to imply that they should be included in (and used from) the login model.  If that's the case and we 'fix' it, we need to decide what to do if we have the property set (because an OAuth provider will _always_ have at least the default scopes set).  Or we could 'fix' it in 1.2 by just removing the property.

jhericks: @dogeared you are working on the framework spec for 1.2.0.  I think we should at _least_ comment out the property and perhaps let it override the one specified in the provider.  _Or_ we can just remove the property from the file and don't reference it in the spec.  Either is easy to do, but what is your recommendation?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/filter/AuthorizeCallbackFilterFactory.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public class AuthorizeCallbackFilterFactory extends ControllerFilterFactory<AuthorizeCallbackController> {
+
+    @Override
+    protected AuthorizeCallbackController newController() {
+        return new AuthorizeCallbackController();
+    }
+
+    @Override
+    protected void configure(AuthorizeCallbackController c, Config config) throws Exception {
```
jhericks: Easy change, but it comes from the `ControllerFilterFactory` abstract method signature.  Should we fix it there, too?

extensions/servlet/src/main/resources/com/stormpath/sdk/servlet/config/web.stormpath.properties
jhericks: Property is removed.

extensions/servlet/src/main/resources/META-INF/resources/WEB-INF/jsp/stormpath/login.jsp
```diff
@@ -75,7 +76,7 @@
                                         <c:otherwise>
                                             <div form-group="true" class="form-group group-${field.name}">
                                                 <label class="<c:out value="${!empty accountStores ? 'col-sm-12' : 'col-sm-4'}"/>">
-                                                    ${field.label}</label>
+                                                        ${field.label}</label>
```
dogeared: Why the extra indent?

extensions/servlet/src/main/resources/META-INF/resources/assets/js/stormpath.js
```diff
@@ -42,44 +46,58 @@ function baseUrl() {
...
-function linkedinLogin(clientId) {
+function linkedinLogin(clientId, scope) {
+    scope = scope || 'r_emailaddress r_basicprofile';
     window.location.replace(
         buildUrl('https://www.linkedin.com/uas/oauth2/authorization',
             {
                 client_id: clientId,
                 response_type: 'code',
-                scope: 'r_emailaddress r_basicprofile',
+                scope: scope,
```
dogeared: this is so nice! Thank you!

impl/src/main/java/com/stormpath/sdk/impl/provider/DefaultGoogleProvider.java
```diff
@@ -16,23 +16,27 @@
...
  * @since 1.0.beta
  */
-public class DefaultGoogleProvider extends AbstractProvider implements GoogleProvider {
+public class DefaultGoogleProvider extends AbstractOAuthProvider<GoogleProvider> implements GoogleProvider {
 
     // SIMPLE PROPERTIES
-    static final StringProperty CLIENT_ID = new StringProperty("clientId");
-    static final StringProperty CLIENT_SECRET = new StringProperty("clientSecret");
-    static final StringProperty REDIRECT_URI = new StringProperty("redirectUri");
+    private static final StringProperty REDIRECT_URI = new StringProperty("redirectUri");
```
dogeared: can we name this something like `INTERNAL_REDIRECT_URL` or put some comments around it so people (like me) don't conflate it with the `redirect_uri` that the final leg redirects to?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/provider/DefaultAccountStoreModel.java
```diff
@@ -24,10 +27,24 @@
 
     private final Directory directory;
     private final ProviderModel providerModel;
+    private final String authorizeUri;
 
-    public DefaultAccountStoreModel(Directory directory, ProviderModel provider) {
+    public DefaultAccountStoreModel(Directory directory, ProviderModel provider, String authorizeBaseUri) {
         this.directory = directory;
         this.providerModel = provider;
+        if (providerModel instanceof OAuthProviderModel && authorizeBaseUri != null) {
```
jhericks: That's right.

extensions/servlet/src/main/resources/META-INF/resources/WEB-INF/jsp/stormpath/login.jsp
jhericks: No good reason, I'm sure.

impl/src/main/java/com/stormpath/sdk/impl/provider/DefaultGoogleProvider.java
jhericks: Sure.  For 'clarity' though, this property is basically irrelevant to social v2.  We no longer want you to set this at the time of _configuration_ of the provider, although it's there for backwards compatibility.  I'll put in a comment to that effect.

dogeared:
LGTM 👍

PR #1094, opened 2016-11-02
Fixes Spring Security whereby it was forcing refresh on authenticatio…

### LINKED ISSUE #1092: Spring Security is forcing auth refresh on every call
Currently, the code that evaluates whether or not to force a refresh on authentication is broken. 

It is resulting in a refresh of the authentication on every call in an application.

The bug is in: `SpringSecurityResolvedAccountFilter:42`

This is impacting current customers and should be prioritized accordingly.

dogeared:
Fixes #1092


mraible:
LGTM 👍

PR #1101, opened 2016-11-09
1081 delete jwt cookie

### LINKED ISSUE #1081: JWT cookie for account is not present any longer: dangling properties and classes
In `web.stormpath.properties` we have these properties which I am pretty sure are not used any longer (since we switched from Account Cookie to AccessToken cookie https://github.com/stormpath/stormpath-sdk-java/pull/633):

If I am right: 
1. these properties **must** be removed
2. the related unused classes **should** be removed or at least deprecated (unless we want to keep them for some reason)
3. Documentation needs to be reviewed and updated accordingly

These are the properties:

```
stormpath.web.account.jwt.resolver =
stormpath.web.account.jwt.factory = 
stormpath.web.account.jwt.signingKey.resolver = 
```

mrioan:
resolves #1081

I did all the analysis and I can conclude that there is not really anything to be done here. Although the Account cookie is not used any longer (replaced by `access token` and `refresh token`) the related classes are still being used for these new cookies. 

I also checked that documentation is still valid.

I only removed an unused property from `web.stormpath.properties`


dogeared:
LGTM - please merge after CI passes

PR #1102, opened 2016-11-09
1065 fixed context path issue in spring-sec-boot-webmvc

### LINKED ISSUE #1065: Hunt down the context path issue in all the existing examples
see: https://github.com/stormpath/stormpath-sdk-java/pull/1064

During the fix for https://github.com/stormpath/stormpath-sdk-java/issues/1061 some hard-coded redirections were found in the `spring-boot-default` example.

A link like this `<a href="/restricted" class="btn btn-primary">Restricted</a>` will redirect the browser to `/restricted` without taking into account the context path. So if the app is running in `http://localhost:8080/myapp/` then that statement will go to `http://localhost:8080/restricted`

The proper Thymeleaf way is: `<a th:href="@{/restricted}" class="btn btn-primary">Restricted</a>` -> this will keep the context path relevant

We would need to check that this error is not happening in the other JSP and Thymeleafs examples.

mrioan:
resolves #1065

### DIFF
--- a/examples/spring-security-spring-boot-webmvc/src/main/resources/templates/error.html
+++ b/examples/spring-security-spring-boot-webmvc/src/main/resources/templates/error.html
@@ -28,7 +28,7 @@
                     <h1 th:text="${errors.status}"></h1>
                 </div>
                 <p class="lead text-muted" th:text="${errors.message}">Unauthorized</p>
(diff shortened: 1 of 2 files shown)


mraible:
LGTM 👍

PR #1115, opened 2016-11-15
1099 Add support to /oauth/revoke endpoint.

### LINKED ISSUE #1099: Support /oauth/revoke endpoint
This issue is to support the new application /oauth/revoke endpoint that implements the revocation functionality as can been in the [Oauth Spec](https://tools.ietf.org/html/rfc7009
)

josebarrueta:
fixes #1099


extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
```diff
@@ -2149,6 +2156,192 @@ class ApplicationIT extends ClientIT {
...
+            accessTokens.add(accessToken)
+
+            refreshToken = refreshToken == null ? refreshTokenResult.refreshToken : refreshToken
+        }
+
+        AccessToken toRevoke = accessTokens.get(3)
+
+        OAuthTokenRevocators.OAUTH_TOKEN_REVOCATOR.forApplication(app).revoke(OAuthRequests.OAUTH_TOKEN_REVOCATION_REQUEST.builder().setToken(toRevoke.getJwt()).build())
+
+        accessTokens.each {
```
dogeared: If you are revoking a single access token from the list, why should they all be revoked?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
```diff
@@ -170,7 +169,7 @@ public void init() {
         Assert.notNull(resultFactory, "accessTokenResultFactory cannot be null.");
         Assert.notNull(accountSaver, "accountSaver cannot be null.");
         Assert.notNull(eventPublisher, "eventPublisher cannot be null.");
-        Assert.notNull(grantTypeStatusValidator, "grantTypeStatusValidator cannot be null.");
+        Assert.notNull(grantTypeValidator, "grantTypeStatusValidator cannot be null.");
```
dogeared: message should be: `grantTypeValidator cannot be null.`

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
```diff
@@ -356,7 +355,7 @@ protected ViewModel doPost(HttpServletRequest request, HttpServletResponse respo
 
         } catch (OAuthException e) {
 
-            log.debug("OAuth Access Token request failed. Root cause: " + e.getMessage());
+            log.debug("OAuth Access Token request failed. Root cause: {}", e.getMessage());
```
dogeared: nice catch!

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
```diff
@@ -389,9 +418,46 @@ private void uncache(String cacheKey, Class<? extends Resource> resourceType) {
...
+            String nonSignedToken = token.substring(0, signatureIndex + 1);
+            Jwt<Header, Claims> jwt = Jwts.parser().parseClaimsJwt(nonSignedToken);
+
+            Object stt = jwt.getHeader().get("stt");
+
+            String rti = null;
+
+            Claims body = jwt.getBody();
+
+            if ("refresh".equals(stt) || "refresh_token".equals(tokenTypeHint)) {
```
dogeared: Why do we need hint and `stt`? Are there situations where we have a token without `stt`? If not, then why do we need tokenTypeHint at all?

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
```diff
@@ -389,9 +418,46 @@ private void uncache(String cacheKey, Class<? extends Resource> resourceType) {
...
     }
 
-    private boolean isApiKeyCollectionQuery(ResourceDataRequest request) {
-        return ApiKeyList.class.isAssignableFrom(request.getResourceClass()) &&
-                request.getUri().hasQuery() && request.getUri().getQuery().containsKey(ID.getName());
+    private void uncacheRevokedToken(Map<String, Object> data) {
+
+        String token = data.get(OAuthTokenRevocationAttempt.TOKEN.getName()).toString();
+
+        int signatureIndex = token.lastIndexOf('.');
```
dogeared: This seems hacky to me. If it's an unsigned token, in our use cases, then there's something wrong, no?

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
```diff
@@ -452,6 +453,8 @@ public ResourceDataResult filter(final ResourceDataRequest req) {
                     // Fix for https://github.com/stormpath/stormpath-sdk-java/issues/218
                     if ( response.getHttpStatus() == 202 ) { //202 means that the request has been accepted for processing, but the processing has not been completed. Therefore we do not have a response body.
                         responseBody = java.util.Collections.emptyMap();
+                    } else if(response.getHttpStatus() == 200 && OAuthTokenRevoked.class.isAssignableFrom(returnType)) {
```
dogeared: won't the response be 404 for a revoked token?

extensions/httpclient/src/test/groovy/com/stormpath/sdk/impl/application/ApplicationIT.groovy
josebarrueta: propagation.

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
josebarrueta: Worst side effect is removing something from cache, so I don't think it's worth it to do more validations.

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
josebarrueta: The developer might remove the stt header from the token, then re-package it and send it.

impl/src/main/java/com/stormpath/sdk/impl/ds/DefaultDataStore.java
josebarrueta: Invalid tokens according to the spec should return 200 as well.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/AccessTokenController.java
josebarrueta: cool

impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
dogeared: We use the `stt` header param to identify token type. removing it invalidates the token, I think. If users want to repackage, I think they should include the header as that's part of what contributes to it being a valid token.

josebarrueta: I actually was able to assert this in the last three lines of: ApplicationIT.testInvalidTokens_NoTokenType(..)

api/src/main/java/com/stormpath/sdk/oauth/TokenTypeHint.java
```diff
@@ -0,0 +1,45 @@
...
+
+    TokenTypeHint(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public static TokenTypeHint fromValue(String value) {
```
josebarrueta: aren't the two approaches equivalent? any advantages from this one?

PR #1122, opened 2016-11-21
Added LogoutController config setting for invalidateHttpSession and produces settings.

### LINKED ISSUE #1121: logout -> invalidateHttpSession = false ignored?
In conjunction with the Stormpath java servlet plugin (tested v1.1.4 and v1.1.5), it seems that the following property:

stormpath.web.logout.invalidateHttpSession = false

... is ignored if set to "false", as it seems that the session is destroyed once a post to the logout URI is made, regardless of the aforementioned configuration.

dogeared:
Fixes #1121 

Steps to UAT:

1. Add the following to `stormpath.properties` in the Servlet Example:
    
    `stormpath.web.logout.invalidateHttpSession=false`

2. Start the Servlet Example and hit the `/login` endpoint. Make note of the `JSESSIONID` in the inspector:

    image

3. Login (`JSESSIONID` should be the same and you should see `access_token` and `refresh_token` cookies)

    image

4. Logout, and make note of the `JSESSIONID` in the inspector (it should not have changed from step 2.)

    image

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/filter/LogoutFilterFactory.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/filter/LogoutFilterFactory.java
@@ -31,5 +31,10 @@ public class LogoutFilterFactory extends ControllerFilterFactory<LogoutControlle
     @Override
     protected void configure(LogoutController controller, Config config) throws Exception {
         controller.setNextUri(config.getLogoutConfig().getNextUri());
+        /*
+            resolves https://github.com/stormpath/stormpath-sdk-java/issues/1121
+         */
+        controller.setInvalidateHttpSession(config.isLogoutInvalidateHttpSession());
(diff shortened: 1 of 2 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/config/filter/LogoutFilterFactory.java
```diff
@@ -31,5 +31,10 @@ protected LogoutController newController() {
     @Override
     protected void configure(LogoutController controller, Config config) throws Exception {
         controller.setNextUri(config.getLogoutConfig().getNextUri());
+        /*
+            resolves https://github.com/stormpath/stormpath-sdk-java/issues/1121
+         */
+        controller.setInvalidateHttpSession(config.isLogoutInvalidateHttpSession());
+        controller.setProduces(config.getProducedMediaTypes());
```
dogeared: Similar response to previous comment

PR #1170, opened 2016-12-14
Issue/1138

### LINKED ISSUE #1138: When reseting your password - entering one that fails the password policy fails badly
If a user triggers a password reset email to be sent to them, clicks the link and then enters a password that violates the currently configured password policy they do not see a message telling them this, instead they're told "The password reset link you tried to use is no longer valid." as shown here:

image

So on entering a policy violating password they see a message relating to a different situation, i.e. a message that one should only ever see immediately after clicking a password reset link that has already gone stale.

Here are the steps to reproduce (assuming that you already have `~/.stormpath` set up and have an existing account). First get an example up and running as shown here. You need to edit the `pom.xml` (the `vi` line below) to remove the `<parent>` block and change the artifact `<version>` to `1.2.1`.

```text
$ curl -OL https://github.com/stormpath/stormpath-sdk-java/archive/master.zip
$ unzip master.zip
$ cp -r stormpath-sdk-java-master/tutorials/spring-boot/03-spring-security-refined password-reset-example
$ cd password-reset-example
$ vi pom.xml
$ mvn package
$ java -jar target/*.jar
```

Now:

* Go to [http://localhost:8080/](http://localhost:8080/), go to login and then to the "Forgot password?" page and enter your email address.
* Click the link in the resulting email and enter a new password, e.g. 123457, that violates the current password policy. Instead of being told your password isn't strong enough you see the message shown up above.

The current password policy can be seen in the Stormpath admin dashboard, under Directories / Policies. By default it requires a password to have at least 8 characters, containing at least one uppercase letter, one lowercase letter and a digit.

**Thoughts:** I've logged a lot of (hopefully useful) issues against the Stormpath SDK. On the whole I've been positive about everything. However this bug upsets me - it's not super serious but it does suggest that at least some fairly basic scenarios have not been covered in testing 😞

dogeared:
Fixes #1138

This is a proposed interim solution until https://stormpath.atlassian.net/browse/AM-3744 is resolved.

@george-hawkins-aa - as you don't have access to our internal Jira system, here's the explanation:

The Stormpath backend does not currently send back different codes for different types of password policy failures. It does send back different messages.

This solution uses the message passed back from Stormpath to display as an error on change password view. The message is specific, such as the below:

image

HOWEVER, this breaks internationalization. Once the backend is returning specific codes for password policy failures, we will be able to implement this in a way that respects internationalization.

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ChangePasswordController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ChangePasswordController.java
@@ -28,6 +28,8 @@ import com.stormpath.sdk.servlet.form.Form;
 import com.stormpath.sdk.servlet.http.MediaType;
 import com.stormpath.sdk.servlet.http.Saver;
 import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -42,6 +44,8 @@ import java.util.Map;
  * @since 1.0.RC4
  */
 public class ChangePasswordController extends FormController {
+
+    private static final Logger log = LoggerFactory.getLogger(ChangePasswordController.class);
 
     private String forgotPasswordUri;
     private String loginUri;
@@ -244,14 +248,18 @@ public class ChangePasswordController extends FormController {
                 next = this.nextUri;
             }
         } catch (ResourceException e) {
-            // resolves https://github.com/stormpath/stormpath-sdk-java/issues/1138
             // 404 is invalid, expired or used sptoken
             if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
                 next = this.errorUri;
             } else {
+                // resolves https://github.com/stormpath/stormpath-sdk-java/issues/1138
+                // TODO This breaks i18n. Fix when Stormpath backend returns specific password policy failure codes.
                 ErrorModel errorModel = errorModelFactory.toError(request, e);
                 next = getUri() + "?sptoken=" + sptoken + "&error=" + URLEncoder.encode(errorModel.getMessage(), "UTF-8");
             }
+        } catch (Exception e) {
+            log.error("Caught exception: {}. Redirecting to: {}", e.getMessage(), errorUri, e);
+            next = errorUri;
         }
 
         return new DefaultViewModel(next).setRedirect(true);
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ChangePasswordController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ChangePasswordController.java
@@ -27,9 +27,11 @@ import com.stormpath.sdk.servlet.form.Field;
 import com.stormpath.sdk.servlet.form.Form;
 import com.stormpath.sdk.servlet.http.MediaType;
 import com.stormpath.sdk.servlet.http.Saver;
+import org.apache.http.HttpStatus;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -212,7 +214,7 @@ public class ChangePasswordController extends FormController {
         String sptoken = form.getFieldValue("sptoken");
 
         if (isJsonPreferred(request, response)) {
-            Map<String, Object> model = new HashMap<String, Object>();
+            Map<String, Object> model = new HashMap<>();
             try {
                 Account account = application.resetPassword(sptoken, password);
                 if (autoLogin) {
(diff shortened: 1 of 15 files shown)


extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/ChangePasswordController.java
```diff
@@ -237,12 +239,19 @@ protected ViewModel onValidSubmit(HttpServletRequest request, HttpServletRespons
             if (autoLogin) {
                 final AuthenticationResult result = new TransientAuthenticationResult(account);
                 this.authenticationResultSaver.set(request, response, result);
-                next = loginNextUri;
+                next = this.loginNextUri;
             } else {
                 next = this.nextUri;
             }
-        } catch (Exception e) {
```
mrioan: I would keep catching Exception as well below the ResourceExcepton. If this catch was properly written here then this means an an exception other than a ResourceExcpetion could be thrown. If that is the case then the exception will not be caught and may cause errors in the invoking code.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/FormController.java
```diff
@@ -130,17 +132,22 @@ protected ViewModel doGet(HttpServletRequest request, HttpServletResponse respon
...
+            ErrorModel error = null;
+            HttpSession session = request.getSession(false);
+            // session null check fixes https://github.com/stormpath/stormpath-sdk-java/issues/908
+            if (session != null && session.getAttribute(SPRING_SECURITY_AUTHENTICATION_FAILED_KEY) != null) {
+                //The login page is being re-rendered after an unsuccessful authentication attempt from Spring Security
+                //Fix for https://github.com/stormpath/stormpath-sdk-java/issues/648
+                //See StormpathAuthenticationFailureHandler
+                error = (ErrorModel) session.getAttribute(SPRING_SECURITY_AUTHENTICATION_FAILED_KEY);
+            } else {
+                //Fix for https://github.com/stormpath/stormpath-sdk-java/issues/1138
```
mrioan: Please add a //TODO here saying "this line breaks i18n we need to improvise this during implementation of issue XXX"

dogeared: fixed. Back to you...

mrioan: Just a comment, I created issue https://github.com/stormpath/stormpath-sdk-java/issues/1171 in order to remember improving this

mrioan:
LGTM @dogeared, as you mentioned it would be awesome to have ITs for this. I think you are already working on that so please add them here so I can review them. Thanks

dogeared:
@mrioan - Added the ITs. Please take a look when you can. FYI, it didn't work to compare error values as sometimes they are the same.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/ChangePasswordIT.groovy
```diff
@@ -0,0 +1,102 @@
...
+
+    @BeforeMethod
+    void setup() {
+        mvc = MockMvcBuilders.webAppContextSetup(context)
+            .apply(StormpathMockMvcConfigurers.stormpath())
+            .build()
+    }
+
+    @Test
+    void testChange() {
```
mrioan: Can you please add a link to the github issue so we know why this test is important? Additionally could you change the name to something more meaningful like "testChangePasswordErrorMessagesAreDescribingTheRealError". 

Now that I think about it George's original concern was that the error message was not really accurately describing the real problem that caused the password not to be changed. This new tests are doing all the logic but at the end they fail to check that the error message is actually realistically describing the real problem. Therefore in the loop you should check that each message is different to all the previos ones you received already. If one of them is causing an identical message then you can move that one to a different loop. Makes sense?

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/MinimalStormpathSpringSecurityWebMvcConfigurationIT.groovy
```diff
@@ -243,50 +233,6 @@ class MinimalStormpathSpringSecurityWebMvcConfigurationIT extends AbstractTestNG
         Assert.isTrue(SecurityContextHolder.getContext().getAuthentication().equals(authentication))
     }
 
-    ///Supporting properties and methods
```
mrioan: Nice, a lot of redundant code to an abstract class. Good.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/model/GuerillaEmail.groovy
```diff
@@ -0,0 +1,48 @@
+package com.stormpath.spring.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+/**
+ * @since 1.2.3
+ */
+public class GuerillaEmail {
```
mrioan: Could you please add a very description for this class?

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/config/ChangePasswordIT.groovy
dogeared: yup. on it.

extensions/spring/stormpath-spring-security-webmvc/src/test/groovy/com/stormpath/spring/model/GuerillaEmail.groovy
mrioan: This comment was not address don't worry I am adding it myself now.

mrioan:
LGTM. You can merge as soon as CI is green. Thanks

george-hawkins-aa:
Thanks for all the work on this issue. i18n is actually very important to us as we're based in Switzerland which is i18n heaven (or hell depending on how you look at it) as most consumer websites have to target German and French at a minimum and usually Italian and English too.

To be honest we're a bit bemused at how seriously you take i18n when it comes to the SDK when there seems to be zero support for it when it comes to the password reset emails etc. that one sets up via api.stormpath.com.

This is covered by http://ideas.stormpath.com/ideas/IAM-I-78 - while the emails sent out by the Stormpath backend remain monolingual there's only so much value to all the i18n support in the SDK 😞

dogeared:
Hey @george-hawkins-aa - As always, thanks for your insights, feedback and contributions to the Java SDK. I think the first hurdle here will be to get specific error codes from the backend. That will make it easy to continue the i18n tradition in the JSDK.

As for the larger issue, I will definitely bring it up to the engineering team. I'd imagine that we'd have to have some sort of data structure to hold templates for all the emails we support in different languages.

PR #1175, opened 2016-12-21
Installing a pinned release of Sphinx.

rdegges:
This will help us avoid issues with docs search across SDKs in the future.

**NOTE**: Please don't merge this yet, please review.


ci/install.sh
```diff
@@ -13,7 +13,7 @@ if [ -n "$BUILD_DOCS" ]; then
   fi
 
   info "Installing Sphinx..."
-  pip -q install --user sphinx &> $WORKDIR/target/pip.log
+  pip -q install -r requirements.txt &> $WORKDIR/target/pip.log
```
mrioan: Since this requirements file is only for docs, can you please move it from root and place it inside the `docs` dir?

mrioan: I asume `--user sphinx` is not needed for any reason, please confirm. I just want to be sure why it was removed.

rdegges: Correct: `--user sphinx` is now deprecated with the new Travis stuff =)

rdegges:
@mrioan just updated the file location. Let me know if this works for ya!

Issue #1073, opened 2016-10-26
Zuul Spring Config for account included/excluded fields does not work


lhazlewood:
Configuring `stormpath.zuul.account.header.includedProperties` or `stormpath.zuul.account.header.excludedProperties` does not work.  Need to use `@ConfigurationProperties` to properly acquire/parse a list.

See http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html#boot-features-external-config-validation

PR #1210, opened 2017-01-11
1073 asymmetric keys

lhazlewood:
Adds asymmetric private key support to #1073


api/src/main/java/com/stormpath/sdk/lang/DefaultRuntimeEnvironment.java
```diff
@@ -0,0 +1,72 @@
...
+                    return;
+                }
+            }
+
+            //bc provider not enabled - add it:
+            Security.addProvider((Provider)Classes.newInstance(clazz));
+            bcLoaded.set(true);
+
+        } catch (UnknownClassException e) {
+            //not available
```
dogeared: maybe a warn or debug log message? Nature abhors an empty catch block! ;)

lhazlewood: This is a perfectly fine/expected condition, but I suppose a debug statement couldn't hurt, except for waiting for CI for another hour before merge. :/

PR #1213, opened 2017-01-11
1212 Fix typo in header name.

### LINKED ISSUE #1212: Stormpath-Client-Request-Id header has a typo
current header is: `X-Stormapth-Client-Request-Id` instead of `X-Stormpath-Client-Request-Id`.

josebarrueta:
Fixes #1212

### DIFF
--- a/extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ClientIT.groovy
+++ b/extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ClientIT.groovy
@@ -39,8 +39,7 @@ import org.testng.annotations.BeforeClass
 import org.testng.annotations.BeforeTest
 
 import static com.stormpath.sdk.application.Applications.newCreateRequestFor
-import static org.testng.Assert.assertEquals
-import static org.testng.Assert.assertTrue
+import static org.testng.Assert.*
 
 abstract class ClientIT {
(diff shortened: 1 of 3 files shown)


extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ClientIT.groovy
```diff
@@ -167,28 +166,22 @@ abstract class ClientIT {
...
             assertEquals(re.status, 404)
             assertEquals(re.getCode(), 404)
         }
-
-        assertTrue(e instanceof ResourceException)
     }
 
-    protected void updatedSaveableError(Saveable input, int expectedErrorCode){
-        Throwable e = null;
+    static void updatedSaveableError(Saveable input, int expectedErrorCode) {
```
dogeared: why static?

dogeared:
@josebarrueta - only one comment about `static` on method in test.

extensions/httpclient/src/test/groovy/com/stormpath/sdk/client/ClientIT.groovy
josebarrueta: It was an Intellij recommendation, I think the reason was because there is no instance state require in the method.

PR #1222, opened 2017-01-12
Updated bom to get rid of parent and force spring sec 4.2.1

### LINKED ISSUE #1221: Update BOM to get rid of parent

dogeared:
Resolves #1221


bom/pom.xml
```diff
@@ -146,4 +139,36 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.sonatype.plugins</groupId>
+                <artifactId>nexus-staging-maven-plugin</artifactId>
+                <version>1.6.7</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <serverId>sonatype-nexus-staging</serverId>
+                    <nexusUrl>https://oss.sonatype.org/</nexusUrl>
+                    <skipStagingRepositoryClose>true</skipStagingRepositoryClose>
```
bdemers: I'm not sure how `skipStagingRepositoryClose` when we have both of these in the same reactor. But _ideally_ we end up with a single staging repo that is closed.  Both this pom and the root pom leave it open.

Not the end of the world, but something we may want to fix, if not now, then in the future.

dogeared: our process (at least up until now) is to close the build on the sonatype website

PR #1234, opened 2017-01-17
Issue 1233 respect challenge creation field

### LINKED ISSUE #1233: Google Authenticator factor creation does not respect "createChallenge" setting on request
DefaultGoogleAuthenticatorCreateFactorRequest has a "createChallenge" field which is apparently not being used.  It should be used in the same way as the identical field in DefaultSmsCreateFactorRequest.  Namely, if its value is true, then a challenge should be created during DefaultAccount.createFactor().

palimpsestor:
Fixes #1233


dogeared:
Closing as things got out of whack after 1.3.1 release and merge to master. #1235 cherry picks @palimpsestor's original commit.

PR #1240, opened 2017-01-17
DefaultTokenResponse.toJson did not include id_token in the JSON #1239

### LINKED ISSUE #1239: DefaultTokenResponse.toJson does not include id_token in the JSON
Now that we can have idToken in the TokenResponse, we need to include id_token in the JSON produced by the toJson() method when idToken is not null.

jhericks:
Fixes #1239

### DIFF
--- a/extensions/oauth/pom.xml
+++ b/extensions/oauth/pom.xml
@@ -58,6 +58,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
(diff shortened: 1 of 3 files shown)


extensions/oauth/pom.xml
```diff
@@ -58,6 +58,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+        </dependency>
```
dogeared: Is this just for testing? If so, would  you please add test scope.

jhericks: Yes, it is.  Thanks for the change.

PR #1244, opened 2017-01-17
Refactor GoogleAuthenticatorChallenge to allow for setting code along with challenge.

### LINKED ISSUE #1243: Refactor GoogleAuthenticatorChallenge to enable setting code at same time as challenge creation

dogeared:
Resolves #1243

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
+++ b/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Stormpath, Inc.
+ * Copyright 2017 Stormpath, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,4 +17,11 @@ package com.stormpath.sdk.challenge.google;
 
 import com.stormpath.sdk.challenge.CreateChallengeRequest;
 
+/**
+ * A Google Authenticator specific create challenge request
+ *
+ * Google Authenticator challenges can validate a code at the same time that the challenge is created.
+ *
+ * @since 1.4.0
+ */
 public interface GoogleAuthenticatorCreateChallengeRequest extends CreateChallengeRequest {}
--- a/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
+++ b/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
@@ -1,0 +1,5 @@
+package com.stormpath.sdk.challenge.google;
+
+import com.stormpath.sdk.challenge.CreateChallengeRequest;
+
+public interface GoogleAuthenticatorCreateChallengeRequest extends CreateChallengeRequest {}
--- a/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
+++ b/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2016 Stormpath, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.stormpath.sdk.challenge.google;
 
 import com.stormpath.sdk.challenge.CreateChallengeRequest;
--- a/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
+++ b/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Stormpath, Inc.
+ * Copyright 2017 Stormpath, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,19 @@ package com.stormpath.sdk.challenge.google;
 
 import com.stormpath.sdk.challenge.CreateChallengeRequestBuilder;
 
+/**
+ * A builder to construct {@link GoogleAuthenticatorCreateChallengeRequest}s Google Authenticator specific create challenge requests.
+ *
+ * Google Authenticator can both create the challenge and set the code in one API call.
+ * See {@link com.stormpath.sdk.factor.google.GoogleAuthenticatorFactor}
+ *
+ * @since 1.4.0
+ */
 public interface GoogleAuthenticatorCreateChallengeRequestBuilder  extends CreateChallengeRequestBuilder<GoogleAuthenticatorChallenge> {
 
+    /**
+     * @param code to be used to validate the challenge for the {@link com.stormpath.sdk.factor.google.GoogleAuthenticatorFactor}
+     * @return GoogleAuthenticatorCreateChallengeRequestBuilder for method chaining with the builder pattern
+     */
     GoogleAuthenticatorCreateChallengeRequestBuilder withCode(String code);
 }
--- a/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
+++ b/api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2016 Stormpath, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.stormpath.sdk.challenge.google;
 
 import com.stormpath.sdk.challenge.CreateChallengeRequestBuilder;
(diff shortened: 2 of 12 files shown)


dogeared:
This enables creating a challenge including the code at the same time for GoogleAuthenticator:

```
        GoogleAuthenticatorChallenge challenge = client.instantiate(GoogleAuthenticatorChallenge.class);
        challenge = (GoogleAuthenticatorChallenge) factor.createChallenge(
            Challenges.GOOGLE_AUTHENTICATOR.newCreateRequestFor(challenge).withCode(code).build()
        );
```

NOTE: I will add the above to the javadocs before deploying. Just wanted to get it built on CI first.

impl/src/main/java/com/stormpath/sdk/impl/challenge/AbstractChallenge.java
```diff
@@ -92,16 +92,11 @@ public Challenge setFactor(T factor) {
...
             return true;
         }
         return false;
     }
-
-    protected Challenge setCode(String code) {
-        setProperty(CODE, code);
-        return this;
-    }
 }
```
mrafiei: Why did we remove the `setCode()` method and made the property public instead?

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorChallenge.java
```diff
@@ -53,4 +53,9 @@ public GoogleAuthenticatorChallengeStatus getStatus() {
         }
         return GoogleAuthenticatorChallengeStatus.valueOf(value.toUpperCase());
     }
+
+    @Override
+    public void setCode(String code) {
+        setProperty(CODE, code);
+    }
```
mrafiei: This method could have stayed in the Abstract Class, no? `SmsChallenge` also needs to be able to set the code.
Looking at the following snippet in IAM for `SmsChallenge`
```JsonPath challengeResponse = postRestCall(challenge.href + "?expand=factor", [code: code])```

File: impl/src/main/java/com/stormpath/sdk/impl/challenge/AbstractChallenge.java (hunk 1 above)
Comment (palimpsestor) [inline impl/src/main/java/com/stormpath/sdk/impl/challenge/AbstractChallenge.java]: Before this PR, that method was only used in one place in the validate method, and in two places in tests which should not have had access to that protected method anyway.  So there was no need for it.
The property, on the other hand, is used both in the validate method and now also when creating a Google Authenticator challenge and providing the code at the same time (which is not allowed in the SMS case).

File: impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorChallenge.java (hunk 2 above)
Comment (palimpsestor) [inline impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorChallenge.java]: See above.  There is no reason for a code to be set on an SMS challenge outside of the validate method.

File: api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java [hunk 3]
```diff
@@ -0,0 +1,5 @@
+package com.stormpath.sdk.challenge.google;
```
mrioan: Missing license header

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
```diff
@@ -0,0 +1,5 @@
...
+
+import com.stormpath.sdk.challenge.CreateChallengeRequest;
+
+public interface GoogleAuthenticatorCreateChallengeRequest extends CreateChallengeRequest {}
```
mrioan: Missing Javadoc. In particular both a short description and the version are important since it is an interface in the api module

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
```diff
@@ -0,0 +1,8 @@
+package com.stormpath.sdk.challenge.google;
```
mrioan: Missing license

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
```diff
@@ -0,0 +1,8 @@
...
+
+import com.stormpath.sdk.challenge.CreateChallengeRequestBuilder;
+
+public interface GoogleAuthenticatorCreateChallengeRequestBuilder  extends CreateChallengeRequestBuilder<GoogleAuthenticatorChallenge> {
```
mrioan: Missing Javadoc in the class and in the method. Both a short description and the version are important since it is an interface in the api module

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorCreateChallengeRequest.java
```diff
@@ -0,0 +1,13 @@
+package com.stormpath.sdk.impl.challenge.google;
```
mrioan: Missing license

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorCreateChallengeRequestBuilder.java
```diff
@@ -0,0 +1,24 @@
+package com.stormpath.sdk.impl.challenge.google;
```
mrioan: Missing license

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorChallenge.java
mrafiei: But doesn't the snippet above post to the challenge href with a code in the body? That means it is supported on the API.

dogeared: @mrafiei - Previously, the interface didn't support calling `setCode` for either SMS or GA as it was not public. Based on the way that GA works, it makes sense to set the code at the same time you create the challenge. It doesn't make the same sense for SMS, which is why we moved it into the more specific GA side of things.

What I am saying is that with GA, you can algorithmically know the code, which is how the TOTP app gives it to you, so it makes sense to be able to create the challenge and submit the code all in one shot.

With SMS, you can't know the code (from an end-user perspective) until you've received the text. I will dig into the spec in Jira/Confluence some more, but am I missing something here?

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequest.java
dogeared: addressed.

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
dogeared: addressed

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorCreateChallengeRequest.java
dogeared: addressed

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorCreateChallengeRequestBuilder.java
dogeared: adressed

impl/src/main/java/com/stormpath/sdk/impl/challenge/google/DefaultGoogleAuthenticatorChallenge.java
palimpsestor: @mrafiei POSTing to a challenge href with a code is indeed supported by the REST API, and that is what is mirrored by the `validate` method on `AbstractChallenge`.

mrafiei: @dogeared Its correct what you stated above, my only thing is since it is supported on the REST API we should also support it for Sms challenges on SDK side to be consistent.

mrioan:
@dogeared now that I see that a developer would do:
```
GoogleAuthenticatorChallenge challenge = client.instantiate(GoogleAuthenticatorChallenge.class);
        challenge = (GoogleAuthenticatorChallenge) factor.createChallenge(
            Challenges.GOOGLE_AUTHENTICATOR
                .newCreateRequestFor(challenge)
                .withCode(code)
                .build()
        );
``` 
I does not look good to me. You are creating a challenge and then creating another one passing that challenge. I think the proper sentence should be like this:

```
GoogleAuthenticatorChallenge challenge = (GoogleAuthenticatorChallenge) factor.createChallenge(
            Challenges.GOOGLE_AUTHENTICATOR
                .withCode(code)
                .build()
        );
```

See that the operation `createChallenge` already knows that it will create a challenge and it is receiving a google code as a parameter.

dogeared:
@mrioan - I agree. However, this pattern is all throughout MFA code and will require some more refactoring to have the change be more systemic. I will endeavor to update the code you pointed out as the model for other changes and I'll put in issues for those other changes in upcoming releases.

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A builder to construct {@link GoogleAuthenticatorCreateChallengeRequest}s Google Authenticator specific create challenge requests.
+ *
+ * Google Authenticator can both create the challenge and set the code in one API call.
+ * See {@link com.stormpath.sdk.factor.google.GoogleAuthenticatorFactor}
+ *
+ * @since 1.4.0
+ */
+public interface GoogleAuthenticatorCreateChallengeRequestBuilder  extends CreateChallengeRequestBuilder<GoogleAuthenticatorChallenge> {
```
mrioan: is this class still needed?

api/src/main/java/com/stormpath/sdk/factor/google/GoogleAuthenticatorFactor.java
```diff
@@ -80,4 +88,12 @@
...
      */
     String getBase64QrImage();
+
+    /**
+     *
+     * @param code The code to validate the challenge that is created. With Google Authenticator, you can create a
+     *             challenge and validate it all in one call.
+     * @return the {@link GoogleAuthenticatorChallenge}
+     */
+    GoogleAuthenticatorChallenge createChallenge(String code);
```
mrioan: @since

api/src/main/java/com/stormpath/sdk/challenge/google/GoogleAuthenticatorCreateChallengeRequestBuilder.java
dogeared: @mrioan - I think so. There are two supported ways to validate a challenge on the backend - by hitting the `.../factors...` collection endpoint and by hitting a specific challenge endpoint, so I think we need to support that in the SDK as well.

dogeared: but, maybe not - looking

PR #1245, opened 2017-01-18
1242 idsite pre and post handlers

### LINKED ISSUE #1242: Pre and Post Handlers are not invoked when IDSite is enabled
Pre/Post Login/Register Handlers are not invoked when IDSite is enabled.

mrioan:
Resolves #1242

If 1.4.0 gets release before this code is approved then this PR should be updated to target 1.4.x

### DIFF
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/CallbackController.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/CallbackController.java
@@ -19,7 +19,6 @@ import com.stormpath.sdk.application.Application;
 import com.stormpath.sdk.authc.AuthenticationResult;
 import com.stormpath.sdk.idsite.LogoutResult;
 import com.stormpath.sdk.lang.Assert;
-import com.stormpath.sdk.lang.Strings;
 import com.stormpath.sdk.saml.SamlResultListener;
 import com.stormpath.sdk.servlet.application.ApplicationResolver;
 import com.stormpath.sdk.servlet.authc.impl.DefaultSuccessfulAuthenticationRequestEvent;
@@ -43,6 +42,8 @@ public abstract class CallbackController extends AbstractController {
     protected Saver<AuthenticationResult> authenticationResultSaver;
     protected Publisher<RequestEvent> eventPublisher;
     protected List<SamlResultListener> samlResultListeners = new ArrayList<SamlResultListener>();
+    protected WebHandler postLoginHandler;
+    protected WebHandler postRegisterHandler;
 
     public void setLoginNextUri(String loginNextUri) {
         this.loginNextUri = loginNextUri;
@@ -105,6 +106,12 @@ public abstract class CallbackController extends AbstractController {
         saveResult(request, response, authcResult);
 
         publish(new DefaultSuccessfulAuthenticationRequestEvent(request, response, null, authcResult));
+
+        if (postLoginHandler != null) {
+            if (!postLoginHandler.handle(request, response, result.getAccount())) {
+                return null;
+            }
+        }
 
         return new DefaultViewModel(loginNextUri).setRedirect(true);
     }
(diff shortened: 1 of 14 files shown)


mrioan:
This branch is already being used by a customer who confirms that things are working fine for him:

See support ticket 10950

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/CallbackController.java
```diff
@@ -106,6 +107,12 @@ protected ViewModel onAuthentication(HttpServletRequest request,
 
         publish(new DefaultSuccessfulAuthenticationRequestEvent(request, response, null, authcResult));
 
+        if (postLoginHandler != null) {
```
dogeared: Is there any reason not to collapse this to:

`if (postLoginHandler != null && !postLoginHandler.handle(request, response, result.getAccount()) {`

?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteController.java
```diff
@@ -65,13 +68,22 @@ public void init() {
         Assert.notNull(callbackUri, "callbackUri must be configured.");
         Assert.notNull(idSiteOrganizationResolver, "idSiteOrganizationResolver must be configured.");
         Assert.notNull(alreadyLoggedInUri, "alreadyLoggedInUri must be configured.");
+        Assert.isTrue(!(preRegisterHandler != null && preLoginHandler != null), "This IDSite controller must contain only one of preRegisterHandler and preLoginHandler");
```
dogeared: Why can you not have both a pre-login handler and a pre-register handler? You can when not using id site.

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
```diff
@@ -126,7 +127,12 @@ private ViewModel onRegistration(final HttpServletRequest request, final HttpSer
             AuthenticationResult authcResult = new TransientAuthenticationResult(account);
             saveResult(request, response, authcResult);
         }
-        // else - do we need to do anything else?
+
+        if (postRegisterHandler != null) {
```
dogeared: may be more work than is worth it, but I wonder if we should have default handlers that return false so we don't need the null checks? NOT high prio...

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/CallbackController.java
mrioan: Yeah, right... Changed!

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteController.java
mrioan: @dogeared the truth is that this `IDSiteController` is instantiated twice (one for `Login` and another for `Registration`; actually I think there `ForgotPassword` and others create more instances as well). So for the sake of speed (a customer need a quick a fix) I did not do what I should have done: creating specific new classes for each case. Do you want me to code it the proper way?

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteResultController.java
mrioan: Makes sense but, as you said, this change will impact several places and there will not be a major benefit now...

extensions/servlet/src/main/java/com/stormpath/sdk/servlet/mvc/IdSiteController.java
mrioan: I am thinking that I should really create those classes: with the current status a `RegisterController` could have a `LoginHandler` attached....

dogeared: oh, I see. I understand we are in a crunch for a client. If you want to release as-is, that's fine. Let's just create an issue for the next release to refactor and clean it up a bit.

dogeared:
After CI, please merge

PR #1266, opened 2017-01-25
1264 - Add support for state token for stormpath_factor_challenge grant_type requests.

### LINKED ISSUE #1264: Support state token for stormpath_factor_challenge grant_type
When handling /oauth/token requests that have a grant_type of stormpath_factor_challenge, we currently expect a challenge href and a code.  We need to also support a state token (with the challenge href embedded) and the code.

palimpsestor:
Fixes #1264

### DIFF
--- a/api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
@@ -23,6 +23,8 @@ package com.stormpath.sdk.oauth;
  */
 public interface OAuthStormpathFactorChallengeGrantRequestAuthentication extends OAuthGrantRequestAuthentication {
 
+    String getState();
+
     String getChallenge();
 
     String getCode();
--- a/api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
+++ b/api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
@@ -23,6 +23,9 @@ package com.stormpath.sdk.oauth;
  */
 public interface OAuthStormpathFactorChallengeGrantRequestAuthentication extends OAuthGrantRequestAuthentication {
 
+    /**
+     * @since 1.5.0
+     */
     String getState();
 
     String getChallenge();
--- a/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication.java
@@ -16,6 +16,7 @@
 package com.stormpath.sdk.impl.oauth;
 
 import com.stormpath.sdk.lang.Assert;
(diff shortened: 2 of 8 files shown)


impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication.java
```diff
@@ -25,18 +26,27 @@
 public class DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication implements OAuthStormpathFactorChallengeGrantRequestAuthentication {
     private final static String grant_type = "stormpath_factor_challenge";
 
+    private String state;
     private String challenge;
     private String code;
 
-    public DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication(String challenge, String code) {
-        Assert.hasText(challenge, "challenge cannot be null or empty.");
+    public DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication(String state, String challenge, String code) {
```
dogeared: What if there's both state and challenge? Is that a valid scenario? If not, do we need to guard against it?

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathFactorChallengeGrantAuthenticationAttempt.java
```diff
@@ -37,6 +37,13 @@
     void setChallenge(String challenge);
 
     /**
+     * Method used to set the state token referencing a challenge to be verified.  Typically this token will have come
+     * in response to a previous OAuth token request.
+     * @param state the state token referencing the challenge to be verified.
```
dogeared: @since 1.5.0

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
```diff
@@ -23,6 +23,8 @@
  */
 public interface OAuthStormpathFactorChallengeGrantRequestAuthentication extends OAuthGrantRequestAuthentication {
 
+    String getState();
```
dogeared: @since 1.5.0

impl/src/main/java/com/stormpath/sdk/impl/oauth/DefaultOAuthStormpathFactorChallengeGrantRequestAuthentication.java
palimpsestor: If they are both provided, they will both be passed on to IAM.  Code on that side basically says "if state is provided, use that, otherwise assume that challenge is set".  So, tl;dr, it's okay (but weird) if both are set.

impl/src/main/java/com/stormpath/sdk/impl/oauth/OAuthStormpathFactorChallengeGrantAuthenticationAttempt.java
palimpsestor: Added.

api/src/main/java/com/stormpath/sdk/oauth/OAuthStormpathFactorChallengeGrantRequestAuthentication.java
palimpsestor: Added.

PR #1268, opened 2017-01-25
Issue 1265

### LINKED ISSUE #1265: Default HttpClient store cookies sent by the server and appends them in every request
Since the HttpClient interacts with a REST server, cookies shouldn't be allowed.

Additionally the apache HttpClient is build in a deprecated way, so probably this needs to be fixed as well.

josebarrueta:
Fixes #1265


mrioan:
In order to UAT I will check no cookie is sent in the requests to the backend

extensions/httpclient/src/main/java/com/stormpath/sdk/impl/http/httpclient/HttpClientRequestExecutor.java
```diff
@@ -42,15 +43,18 @@
 import org.apache.http.HttpVersion;
```
mrioan: unused

PR #1272, opened 2017-01-27
Allow for configurable maxAge. Docs still need to be updated.

### LINKED ISSUE #1129: add max-age property to list of stormpath.web.accessTokenCookie and stormpath.web.refreshTokenCookie properties
By default, we have the `max-age` for `access_token` and `refresh_token`cookies tied to the `exp` claim of the JWT, which is tied to the OAuth Policy in Stormpath.

Developers should be able to override `max-age` explicitly through configuration.

Use cases include setting `max-age` to `-1` which is used to ensure that when the browser is closed the session ends.

`stormpath.web.accessTokenCookie.max-age` and
`stormpath.web.refreshTokenCookie.max-age`

should be added as available properties with an empty default value (which indicates the default behavior described above).

dogeared:
Resolves #1129

### DIFF
--- a/docs/source/appendix/spring-boot-web-properties.rst
+++ b/docs/source/appendix/spring-boot-web-properties.rst
@@ -14,7 +14,7 @@ stormpath.web.authc.savers.cookie.enabled
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 **Default Value:** ``true``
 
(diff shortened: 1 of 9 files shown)


dogeared:
Good catches! I updated. Back to you @mrioan

mrioan:
@dogeared I see this comments is still not addressed: https://github.com/stormpath/stormpath-sdk-java/pull/1272#discussion_r98457320. I think it is important since I am not even myself aware of the unit there, therefore users will be confused as well.

PR #1280, opened 2017-02-01
1017 - Fix RuntimeException thrown by non-resource maps

### LINKED ISSUE #1017: Everytime a new map (non-resource property) is added to the RestApi we need to hardcode the WriteCacheFilter to not break it.
the WriteCacheFilter currently validates all properties that are deserialized to maps, and enforces them to have an href property

WriteCacheFilter:234

```
  Assert.notNull(nested.get(AbstractResource.HREF_PROP_NAME),
                               "Resource references must have an '" + AbstractResource.HREF_PROP_NAME + "' attribute.");
```

Currently in the RestAPI we are adding more Json objects (maps/list, etc) that are not resources, if we release the RestAPI we'll be breaking all existent clients.

Instead of hardcoding each new property as we add them in the core api, we should implement a generic solution that doesn't break clients every time.

Of course in the REST API we will filter versions of the SDK to not break them, but going forward we shouldn't need it.

josebarrueta:
Fix #1017

### DIFF
--- a/impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
+++ b/impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
@@ -18,12 +18,9 @@ package com.stormpath.sdk.impl.ds.cache;
 import com.stormpath.sdk.account.Account;
 import com.stormpath.sdk.account.EmailVerificationToken;
 import com.stormpath.sdk.account.PasswordResetToken;
-import com.stormpath.sdk.api.ApiKey;
-import com.stormpath.sdk.application.webconfig.ApplicationWebConfig;
 import com.stormpath.sdk.cache.Cache;
 import com.stormpath.sdk.directory.CustomData;
 import com.stormpath.sdk.impl.account.DefaultAccount;
-import com.stormpath.sdk.impl.api.ApiKeyParameter;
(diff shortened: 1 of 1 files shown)


impl/src/main/java/com/stormpath/sdk/impl/ds/cache/WriteCacheFilter.java
```diff
@@ -245,27 +230,10 @@ private void cache(Class<? extends Resource> clazz, Map<String, ?> data, QuerySt
...
-            // do need to have an href property
-            if (isDefaultModelMap || isTokenDataMap || isWebConfigurationMap) {
-                value = new LinkedHashMap<String, Object>((Map) value);
-            }
+            Map<String, ?> nested;
 
-            if (value instanceof Map && !isDefaultModelMap && !isTokenDataMap && !isWebConfigurationMap && !isApiEncryptionMetadata) {
-                //the value is a resource reference
-                Map<String, ?> nested = (Map<String, ?>) value;
+            if (value instanceof Map && (nested = (Map<String, ?>) value).containsKey(AbstractResource.HREF_PROP_NAME)) {
```
dogeared: oo - fancy. I like it.

dogeared:
LGTM - any reason not to target the 1.5.x release? I update the target in the PR.

PR #1286, opened 2017-02-02
Added static default index page with logout form

### LINKED ISSUE #1283: Add default home page with /logout POST form
Right now, if you have an empty project, such as one generated by start.spring.io, there is no way to logout since `/logout` requires a POST.

dogeared:
Fixes #1283 

This is the "simplest thing that will work today".

It solves the problem of having a (static) default view when there's an empty project, such as one created by `start.spring.io`

UAT:

**Basic**:

1) Create a completely empty project that references: `stormpath-default-spring-boot-starter`.
2) Browse to: `http://localhost:8080`
3) You should be redirected to `/login`
4) Login and you should be redirected back to the default static home page that includes a logout button
5) Logout and you should be redirected back to the `/login` page

**Controller Override**:

1) Create a completely empty project that references: `stormpath-default-spring-boot-starter`.
2) Add a Controller that maps `/` as in:

    ```
    @Controller
    public class HomeController {

        @RequestMapping("/")
        public String home() {
            return "home";
        }
    }
    ```

3) Create a simple template in `src/main/resources/templates/home.html` like:

    ```
    <h1>Hi!</h1>
    ```

4) Browse to: `http://localhost:8080`
5) You should be redirected to `/login`
6) Login and you should be redirected back to the home template you created in step 3

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -7,7 +7,7 @@
     <name>Stormpath Java SDK :: BOM</name>
     <description>Stormpath Bill of Materials</description>
     <packaging>pom</packaging>
-    <version>1.5.0</version>
+    <version>1.5.2-SNAPSHOT</version>
 
     <url>https://github.com/stormpath/stormpath-sdk-java</url>
     <organization>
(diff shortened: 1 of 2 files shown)


dogeared:
@mrioan - that's true, but it's really only relevant for spring boot projects. That is, there really isn't another type of project (spring or servlet) that would even work without any source code at all.

mrioan:
@dogeared I understand your point but I am concerned about homogeneity here.... 

One developer just trying Stormpath for the first time will not need to worry about `logout` in Spring Boot, but they will need to if they are using Spring or Servlet. I think integrations should not differ behavior-wise if there is not a technical reason for them to differ.

mrioan:
@dogeared I have UATed and Spring Boot is working fine. My only suggesting would be to have the text and the button in the `logout` page centered.

dogeared:
@mrioan - Thanks. I will update.

PR #1303, opened 2017-02-10
Reuse BOM for interproject dependencies.

bdemers:
cc: @dogeared You may have already tried this, if so just close


dogeared:
We actually had it this way originally. Spring team prefers explicit
versions in the BOM.

Maven updates the individual versions automatically, so we didn't mind
complying.

PR #1313, opened 2017-03-08
Log invalid JWT token exception at debug level

### LINKED ISSUE #1076: Please log "invalid JWT in refresh_token cookie" as just an error (without an exception)
Currently we run tests where we do things such as create and delete users and test that the given user can no longer access anything.

When we do this we see a reasonable Stormpath error message in our logs:

> Encountered invalid JWT in refresh_token cookie. We will now delete both the access and refresh cookies for safety.

However these messages are always accompanied by an exception stack trace. However this is a _handled_ situation - the system knows what's going on and has reacted to it.

Generally I think logging stack traces should be reserved for situations where the system isn't quite sure what's happened and is logging where things went wrong so that someone can debug the point of failure (possibly at an unknown point far away from where the exception has been caught).

Currently we treat exceptions in our logs as indications that something is probably broken. This is not the case in this situation. Our QA investigates any exceptions found in the logs - so currently this exception has to be put in a special category of expected exceptions.

Perhaps log the message at log level ERROR and the exception at log level DEBUG if you're worried that the exception may not always indicate a JWT issue (you are after all catching `Exception e` rather than a JWT specific exception).

Stephan202:
Rationale: this is a client error for which no server-side action is required.
Additionally, elsewhere in the class a similar error is also logged at debug
level.

This fixes #1076


Stephan202:
(The Travis CI failure seems completely unrelated to my change...)

Stephan202:
Updated the PR to target the `okta` branch. Is the Stormpath team interested in merging this change?

Stephan202:
Development of this code base seems to have fully ceased. Will leave the patch here in a comment and drop the associated fork.
```
Date:   Wed Mar 8 18:39:39 2017 +0100

    Log invalid JWT token exception at debug level
    
    Rationale: this is a client error for which no server-side action is required.
    Additionally, elsewhere in the class a similar error is also logged at debug
    level.
    
    This fixes #1076

diff --git a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
index 12bbf69..12b865d 100644
--- a/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
+++ b/extensions/servlet/src/main/java/com/stormpath/sdk/servlet/filter/account/CookieAccountResolver.java
@@ -141,7 +141,7 @@ public class CookieAccountResolver implements Resolver<Account> {
 
         } catch (Exception e) {
             String msg = "Encountered invalid JWT in refresh_token cookie. We will now delete both the access and refresh cookies for safety.";
-            log.error(msg, e);
+            log.debug(msg, e);
             deleteCookie(response, cookie);
             deleteCookie(response, accessTokenCookieResolver.get(request, response));
         }
```

PR #1315, opened 2017-03-10
Okta Password Policy

dogeared:
UAT: `RegisterIT` of the TCK should pass fully

This PR builds in the ability to fetch the password policy from Okta.

It also has some foundational updates that should benefit other efforts, including:

* handling JSON List responses (by wrapping them in a Map)
* handling Okta error responses


bdemers:
Resolved by #1319

PR #1317, opened 2017-03-17
Master without client api

bdemers:
The diff of the pull request is going to look jacked.

Compare it with 1.5.x and it should look fine.
https://github.com/stormpath/stormpath-sdk-java/compare/1.5.x...master-without-client-api

This basically moves master back to 1.5.x (without changing the version poms)
cc: @dogeared


bdemers:
green means go!

bdemers:
@dogeared  bump

tck.sh
```diff
@@ -44,6 +44,8 @@ case "$OPTION" in
         ;;
     run)
         SCRIPT_DIR=$(cd "$(dirname "$0")"; pwd)
+        CI_DIR_CONTENTS=`ls $SCRIPT_DIR/ci`
```
dogeared: Is this necessary? I don't think it's even outputting the CI_DIR_CONTENTS anyway

dogeared:
@bdemers - one minor comment. LGTM 👍

tck.sh
bdemers: I would say no, but I'll let you ask the author: https://github.com/stormpath/stormpath-sdk-java/commit/1afd2a83802ae93e98f3233d48271f08dc73b3ae 🔥

bdemers: I kid, I kid, I just checked the build log, and I don't see that log line, I'll remove those and merge

PR #1322, opened 2017-04-10
Upgrade to latest version of Spring Cloud Netflix

mraible:


bdemers:
Merging, the failing CI is likely due to @mraible's CI config (likely setup for master). If this is not the case, this is trivial to back out.

PR #1326, opened 2017-04-28
Passing TCK!

bdemers:
Bunch of changes in this, but they can be broken down by commit


bdemers:
travis config still needs some work, merging as the spring and servlet tests all pass

PR #1341, opened 2017-06-14
Removes duplicate configured account resolver filter

bdemers:
When more then one account resolver filter is in the chain, multiple authn requests are made to the server
(with Okta this may cause a rate limiting error)

Fixes: 1340


bdemers:
NOTE: there is some dead code that needs to be cleaned up first.  Looking for feedback first

dogeared:
let's get travis working, re-run the jobs and then FISI! :shipit:

dogeared (approved):

PR #1347, opened 2017-07-14
Use UID when sending verification email

bdemers:


dogeared (approved):
LGTM 🤘 Release the Kraken!

PR #1352, opened 2017-08-17
Added revoke refresh token on logout so refresh token cannot be reused to obtain more access tokens.

dogeared:


bdemers (approved):
🚢

mraible:
@bdemers If this fix is in a release, can you please mark its milestone?

Issue #1355, opened 2017-08-23
Default web.stormpath.properties missing Okta values


mraible:
Should have:

```
stormpath.client.authenticationScheme=SSWS
okta.enabled=true
stormpath.web.account.jwt.resolver = com.stormpath.sdk.servlet.filter.account.config.OktaJwtAccountResolverFactory
stormpath.web.account.jwt.signingKey.resolver = com.stormpath.sdk.servlet.filter.account.config.OktaJwtSigningKeyResolverFactory
stormpath.jwt.signingKey.resolver = com.stormpath.sdk.servlet.filter.account.config.OktaSigningKeyResolverFactory
stormpath.web.refreshToken.resultFactory = com.stormpath.sdk.servlet.filter.oauth.config.OktaRefreshTokenResultFactoryFactory
```

This affects people trying to use the servlet plugin.

PR #1357, opened 2017-08-25
Add okta properties to the stock web.stormpath.properties

bdemers:
Fixes: #1355 

The tests are taking a long time to run, so we may need to add the previous defaults to the test `stormpath.properties`, but I want to see how CI runs first


mraible (approved):
LGTM!

